PR #1, opened 2017-01-18
Initial JJB job

mattt416:
This commit creates the initial JJB bootstrap job. This is based off
the change in [1], but moves the scl outside of the if statement and
removes the installation of virtualenv (since it should already be
installed).

[1] https://github.com/rcbops-qe/jenkins-jobs/pull/2

Connects https://github.com/rcbops/u-suk-dev/issues/960

### DIFF
--- a/rpc-jobs/jjb-setup.yml
+++ b/rpc-jobs/jjb-setup.yml
@@ -53,12 +53,13 @@
 
             [jenkins]
             url=${JENKINS_URL}
-            user=${JENKINS_USER}
-            password=${JENKINS_API_PASSWORD}
             EOF
 
             if [ "$IGNORE_CACHE" = "true" ]; then
                 JJB_ARGS="--ignore-cache"
             fi
 
-            jenkins-jobs --conf jenkins_jobs.ini $JJB_ARGS update $JOBS
+            jenkins-jobs --conf jenkins_jobs.ini \
+                         --user $JENKINS_USER \
+                         --password $JENKINS_API_PASSWORD \
+                         $JJB_ARGS update $JOBS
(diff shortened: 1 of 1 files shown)


melissa-kam:
:+1: Looks good to me! Though I did make the original PR so I'm not sure if someone else should also review it.

rpc-jobs/jjb-setup.yml
```diff
@@ -0,0 +1,43 @@
...
+            fi
+            source .venv/bin/activate
+
+            pip install jenkins-job-builder
+
+            if [ "$IGNORE_CACHE" = "true" ]; then
+                JJB_ARGS="--ignore-cache"
+            fi
+
+            jenkins-jobs --conf ~/jenkins_jobs.ini $JJB_ARGS update $JOBS
```
hughsaunders: Where does jenkins_jobs.ini come from? Could we write it out using creds from Jenkins Credential storage? Then this will work on other slaves. Or pull the config file from the repo and replace creds placeholders with creds from Jenkins.

Also this could be a template+project with an operation axis, so that we can create an update job for post merge and syntax check for PRs.

mattt416: I like the idea of storing the sensitive bits in the credentials store.  I'll have a poke at that.

rpc-jobs/jjb-setup.yml
```diff
@@ -0,0 +1,64 @@
...
+            scl enable python27 bash
+
+            if [[ ! -d ".venv" ]]; then
+                virtualenv -p /opt/rh/python27/root/usr/bin/python .venv
+            fi
+            source .venv/bin/activate
+
+            pip install jenkins-job-builder
+
+            cat > jenkins_jobs.ini << EOF
```
melissa-kam: If we put the credential file in the workspace, though, it'll be viewable in plain text from the UI since you can look at the current state of a job's workspace.

hughsaunders: Gah, I didn't think of that. This method LGTM but as @melissa-kam says, lets write the file outside the workspace.

PR #2, opened 2017-01-23
Adding Jenkins slave node setup

melissa-kam:
Instead of updating the current Jenkins swarm script to support Xenial as well as Trusty, I switched to a python script that uses the Jenkins API to accommodate for the CIT Jenkins restrictions. I also left out the server deployment since that's being converted to groovy scripts.

Connects https://github.com/rcbops/u-suk-dev/issues/940


playbooks/setup-jenkins-slave.yml
```diff
@@ -0,0 +1,69 @@
...
+- hosts: cloud_instances
+  tasks:
+    - name: Install apt packages
+      apt:
+        pkg: "{{ item }}"
+        state: installed
+        update_cache: yes
+      with_items:
+        - git
+        - default-jre
```
mattt416: Is this needed?

mattt416:
Ha, guess we didn't think about this.  Anyway, one inline nit but otherwise this LGTM!

playbooks/setup-jenkins-slave.yml
melissa-kam: If you mean default-jre specifically, it's needed to run the Jenkins slave jar since it's Java based.

mattt416: 👍

PR #4, opened 2017-01-25
RPC-AIO Pipeline Job First Steps

hughsaunders:
This commit adds the RPC-AIO Pipeline job and the first couple of stages
 * Allocate Resources
 * Cleanup Resources

 The rest of the stages will be added in subsequent commits.

 Connects rcbops/u-suk-dev#1038


hughsaunders:
https://rpc.jenkins.cit.rackspace.net/job/RPC_AIO/

pipeline-steps/common.groovy
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/* Run a stage if the stage name is contained in an env var
+ * Args:
+ *   - stage_name: String name of this stage
+ *   - stage: Closure to execute
+ * Environment:
+ *    - STAGES: String list of stages that should be run
+ */
+def conditionalStage(Map args){
```
melissa-kam: So there might be a drawback with doing it this way because the Jenkins stage view can't handle dynamic stages. Basically, if the structure of the stages change, it'll reset the stage view and we lose the nice stage view of the older runs. For example, if you previously had `Allocate | Prepare | Deploy | Test | Cleanup` currently in the stage view and you do a new build with just `Prepare | Deploy | Test | Cleanup`, the stage view will no longer show the old build and only have the latest. It's okay if it's just missing the last stages like `Allocate | Prepare | Deploy | Test`, though.

It's not a huge deal since it's just a visual aid, but it's maybe something to consider.



https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Stage+View+Plugin

> Dynamic stages: in general, if you want to visualize dynamically changing stages, make it conditional to execute the stage contents, not conditional to include the stage** Stage view can handle a limited subset of cases where you're appending additional stages, but the cell-based view is not matched to changing stage structures

melissa-kam: The alternative isn't great either, though, which would be to always have all stages showing even though nothing happened in that stage...

rpc-jobs/RPC-AIO.yml
```diff
@@ -0,0 +1,71 @@
...
+    name: RPC_AIO
+    project-type: workflow
+    concurrent: true
+    parameters:
+      - string:
+          name: RPC_GATING_REPO
+          default: "https://github.com/rcbops/rpc-gating"
+      - string:
+          name: RPC_GATING_BRANCH
+          default: "bug/1038"
```
melissa-kam: This looks like it needs to be changed back to master.

hughsaunders: Thanks, done.

pipeline-steps/common.groovy
hughsaunders: I've switched to the empty stage method. It's ok for now and we can add a stage annotation to show that the stage was skipped once https://issues.jenkins-ci.org/browse/JENKINS-26522 merges.

rpc-jobs/RPC-AIO.yml
```diff
@@ -0,0 +1,71 @@
...
+                keyname: "jenkins",
+                workspace: env.WORKSPACE
+              )
+            } //stage
+          ) //conditionalStages
+        } // node
+      } finally {
+        node(){
+          common.conditionalStage(
+            stage_name: 'Cleanup',
```
melissa-kam: This'll run into that same issue of resetting the stage view when we have more stages unfortunately. If a stage fails before the whole pipeline is done and the cleanup stage becomes the next stage instead, it'll reset. I've previously worked around this by not running the cleanup tasks in a stage, though that has the downside of not being able to see at a glance whether cleanup ran or not.

hughsaunders: I've added conditionalStep(step_name, step) which is the same as conditionalStage but doesn't wrap the execution in a stage. It still takes a name parameter as that is checked against env.STAGES to determine if the step should be run.

melissa-kam:
Looks good to me! :+1:  :tada:

PR #6, opened 2017-01-26
RPC-AIO Pipeline Connect Slave

hughsaunders:
Add reusable step for setting up an existing node as a jenkins slave.

Depends on #4

Connects rcbops/u-suk-dev#1066

### DIFF
--- a/pipeline-steps/pub_cloud_slave_highlevel.groovy
+++ b/pipeline-steps/pub_cloud_slave_highlevel.groovy
@@ -1,0 +1,56 @@
+def getPubCloudSlave(Map args){
+  build_name = ""
+  common.conditionalStage(
+    stage_name: 'Allocate Resources',
+    stage: {
+      def allocate = load 'rpc-gating/pipeline-steps/allocate_pubcloud.groovy'
+
+      if (env.INSTANCE_NAME == "AUTO"){
+        job_name_acronym=""
+        job_words=env.JOB_NAME.split("[-_ ]")
+        for (i=0; i<job_words.size(); i++){
+          job_name_acronym += job_words[i][0]
+        }
+        build_name = "${job_name_acronym}-${env.BUILD_NUMBER}"
+      }
+      else {
+        build_name = env.JOB_NAME
+      }
+
+      def resources = allocate (
+        name: build_name,
+        count: 1,
+        region: env.REGION,
+        flavor: env.FLAVOR,
+        image: env.IMAGE,
+        keyname: "jenkins",
+      )
+    } //stage
+  ) //conditionalStages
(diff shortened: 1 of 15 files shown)


playbooks/remove-jenkins-slave.yml
```diff
@@ -0,0 +1,13 @@
+- hosts: job_nodes
```
mattt416: Would this not need to run on the CIT jenkins slave itself?

hughsaunders: It does, its delegated.

mattt416: :eagle:

rpc-jobs/RPC-AIO.yml
```diff
@@ -19,12 +19,21 @@
...
+          name: INSTANCE_NAME
+          default: AUTO
+          description: |
+            Set instance name or "AUTO" to have it generated based on job
+            name and build number. Beware the 63 char container name limit
+            as container names includes the host name.
+      - string:
           name: STAGES
-          default: "Allocate Resources, Cleanup"
+          default: "Allocate Resources, Connect Slave, Pause, Cleanup"
```
melissa-kam: If we pause by default, will that turn into a manual burden of having to input continue for all the builds?

hughsaunders: Yes! that was only there for testing, I'll remove it.

pipeline-steps/pub_cloud_slave_highlevel.groovy
```diff
@@ -0,0 +1,56 @@
...
+      if (env.INSTANCE_NAME == "AUTO"){
+        job_name_acronym=""
+        job_words=env.JOB_NAME.split("[-_ ]")
+        for (i=0; i<job_words.size(); i++){
+          job_name_acronym += job_words[i][0]
+        }
+        build_name = "${job_name_acronym}-${env.BUILD_NUMBER}"
+      }
+      else {
+        build_name = env.JOB_NAME
```
melissa-kam: This should be `env.INSTANCE_NAME`.

rpc-jobs/RPC-AIO.yml
hughsaunders: Done.

pipeline-steps/pub_cloud_slave_highlevel.groovy
hughsaunders: Good catch, fixed.

hughsaunders:
This is now ready for review again, after a non-functional period yesterday. That was caused by me forward porting the single use slave template commit into this branch.

PR #7, opened 2017-01-26
Remove node label from JJB job

hughsaunders:
The CIT Jenkins slave doesn't have the general label by default, so
whenever it is reconfigured, that label is removed.

ITs better to not specify a label, then jenkins will automatically
allocate a slave based on the configuration of slaves.

connects rcbops/u-suk-dev#1070

### DIFF
--- a/rpc-jobs/jjb-setup.yml
+++ b/rpc-jobs/jjb-setup.yml
@@ -3,7 +3,6 @@
     description: Creates and updates jobs with Jenkins Job Builder.
     logrotate:
         daysToKeep: 20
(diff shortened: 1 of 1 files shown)


mattt416:
We have some other misc slaves in at the moment, that's not going to cause an issue is it?

hughsaunders:
@mattt416 good point, but will be ok, because misc slaves have "Only buil jobs with label expressions matching this node" set, whereas the CIT slave has "use this node as much as possible" 
screen shot 2017-01-26 at 12 13 57

PR #10, opened 2017-02-02
Add AIO Prepare Stage

hughsaunders:
Also refactors public cloud allocate and cleanup, so that cleanup still
works when there is a failure part way through the allocate stage.

connects rcbops/u-suk-dev#1069


pipeline-steps/aio_prepare.groovy
```diff
@@ -0,0 +1,39 @@
...
+  common.conditionalStage(
+    stage_name: "Prepare Deployment",
+    stage: {
+      dir("/opt/rpc-openstack"){
+        git branch: env.RPC_BRANCH, url: env.RPC_REPO
+        sh "git submodule update --init"
+        withEnv([
+          "DEPLOY_AIO=yes",
+          "DEPLOY_OA=no",
+          "DEPLOY_RPC=no"
```
mattt416: Do you want OSA to be deployed at this stage?  I think `DEPLOY_OA` defaults to `yes`.

rpc-jobs/single_use_slave_template.yml
```diff
@@ -6,23 +6,40 @@
...
               sh """
                 hostname; date
                 echo "your script here"
               """
             }
-          }
+          } // pub cloud node
+        } catch (Exception e){
+          currentBuild.result = 'FAILURE'
+          print e
```
mattt416: Would it be better to do e.printStackTrace() so we get the full stack trace?  I had an exception in my run and couldn't find the error because all it prints is:

```
...
[Pipeline] }
[Pipeline] // withEnv
[Pipeline] }
[Pipeline] // dir
[Pipeline] withCredentials
[Pipeline] // withCredentials
[Pipeline] }
[Pipeline] // stage
[Pipeline] }
[Pipeline] // node
[Pipeline] echo
org.jenkinsci.plugins.credentialsbinding.impl.CredentialNotFoundException: dev_pubcloud_user_key
[Pipeline] load
[Pipeline] { (rpc-gating/pipeline-steps/ssh_slave.groovy)
[Pipeline] }
[Pipeline] // load
[Pipeline] echo
Skipped: Pause
[Pipeline] echo
Step Start: Cleanup
[Pipeline] withCredentials
[Pipeline] {
[Pipeline] dir
Running in /var/lib/jenkins/workspace/RPC_AIO/rpc-gating/scripts
[Pipeline] {
[Pipeline] sh
...
```

hughsaunders: I would like a full stack trace as well, but I haven't found a method that works yet :(

hughsaunders: Re-throw seems to be the most reliable way to get a stack trace.

PR #11, opened 2017-02-02
Multi-Node AIO: Slave allocation and setup

melissa-kam:
Successful Run:
https://rpc.jenkins.cit.rackspace.net/job/OnMetal_Multi_Node_AIO/25/

Connects https://github.com/rcbops/u-suk-dev/issues/1091

### DIFF
--- a/rpc-jobs/multi-node-aio.yml
+++ b/rpc-jobs/multi-node-aio.yml
@@ -1,13 +1,26 @@
 - job:
-    name: 'OnMetal_Multi_Node_AIO_Prepare_Deployment'
+    name: 'OnMetal_Multi_Node_AIO'
     project-type: workflow
     build-discarder:
         days-to-keep: 30
     concurrent: true
     parameters:
       - string:
-          name: NODE
-          description: "Name or label of Jenkins node to run on"
+          name: REGION
+          default: "IAD"
+      - string:
+          name: FLAVOR
+          default: "onmetal-io1"
+      - string:
+          name: IMAGE
+          default: "OnMetal - Ubuntu 14.04 LTS (Trusty Tahr)"
+      - string:
+          name: INSTANCE_NAME
+          default: AUTO
+          description: |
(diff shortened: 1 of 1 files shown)


rpc-jobs/multi-node-aio.yml
```diff
@@ -36,77 +49,95 @@
...
+                    script: 'deploy-osa.sh',
+                    environment_vars: [
+                    "OSA_BRANCH=${env.OPENSTACK_ANSIBLE_BRANCH}",
+                    "RUN_OSA=false"]
+                  )
+                } //stage
+              ) //conditionalStage
+            } // dir
+          }// public cloud node
+        } catch (Exception e){
```
hughsaunders: I've been playing with this today.. Exception doesn't catch all throwables (some of them derive from Error instead). But as @mattt416 pointed out, you can just catch(e). 

On that note, if you print(e) you don't get a full stack trace, so it may be better to set the build result then re-throw e.

rpc-jobs/multi-node-aio.yml
```diff
@@ -1,13 +1,26 @@
...
+    name: 'OnMetal_Multi_Node_AIO'
     project-type: workflow
     build-discarder:
         days-to-keep: 30
     concurrent: true
     parameters:
       - string:
-          name: NODE
-          description: "Name or label of Jenkins node to run on"
+          name: REGION
```
hughsaunders: These are in single_use_slave_params in rpc-gating/rpc-jobs/params.yml

Maybe that could have a better name, cloud_slave_params?

melissa-kam: I was thinking about that too, but the OnMetal jobs will have different defaults for image and flavor and potentially region in the future if the non-OnMetal move from IAD to a different region.

hughsaunders: Ah yes, I hadn't considered the defaults.

PR #12, opened 2017-02-02
Bug/1069 deploy

hughsaunders:
Adds a step which runs all the playbooks specified by a json parameter.

Depends on #10

connects rcbops/u-suk-dev#1069


pipeline-steps/deploy.groovy
```diff
@@ -0,0 +1,19 @@
+def deploy(){
+  common.conditionalStep(
+    step_name: "Deploy",
+    step: {
```
melissa-kam: Could we also have the option of deploying using the deploy.sh script? From a QE perspective, we'll want to also test that deploy.sh is also working properly in addition to the individual playbooks. We could also make it a separate pipeline-step if that would be more straightforward (like a deploy-playbooks and a deploy-script kind of business).

hughsaunders: Heh, I should have checked earlier, that method is much easier, can use your common.run_script. However this method will be good for customisation, re-running etc.

melissa-kam: Oh yeah, I wrote that, didn't I! That makes sense. :)

hughsaunders: I added a quick function for this, but it occurred to me that running deploy.sh is going to repeat most of the prep stage. There are conditionals in the deploy script that allow prep steps only but no actual deploys, but not the other way round.

The majority of the deploy script is tested in the aio prep stage however: https://github.com/rcbops/rpc-gating/blob/bug/1069/pipeline-steps/aio_prepare.groovy#L14

hughsaunders:
Successful run including MaaS: https://rpc.jenkins.cit.rackspace.net/job/RPC_AIO/221/
This is now ready for review.

PR #13, opened 2017-02-03
Add job to install Jenkins plugins

mattt416:
Please let me know if I'm missing any plugins.  I'm also not sure if this is a good approach to take so input appreciated.  :)

Connects https://github.com/rcbops/u-suk-dev/issues/1082

### DIFF
--- a/playbooks/install_jenkins_plugins.yml
+++ b/playbooks/install_jenkins_plugins.yml
@@ -1,0 +1,106 @@
+---
+- hosts: localhost
+  vars:
+    jenkins_credentials:
+      url_username: "{{ ansible_env.JENKINS_USER }}"
+      url_password: "{{ ansible_env.JENKINS_API_PASSWORD }}"
+      url: "{{ ansible_env.JENKINS_URL }}"
+  handlers:
+    - name: Safely restart jenkins
+      uri:
+        url: "{{ jenkins_credentials['url'] }}/safeRestart"
+        user: "{{ jenkins_credentials['url_username'] }}"
+        password: "{{ jenkins_credentials['url_password'] }}"
+        method: POST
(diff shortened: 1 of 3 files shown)


rpc-jobs/install-jenkins-plugins.yml
```diff
@@ -0,0 +1,37 @@
...
+            wipe-workspace: false
+    wrappers:
+      - credentials-binding:
+          - username-password-separated:
+             credential-id: service_account_jenkins_api_creds
+             username: JENKINS_USER
+             password: JENKINS_API_PASSWORD
+    builders:
+        - shell: |
+            scl enable python27 bash
```
hughsaunders: pipeline-steps/common.groovy has install_ansible() for installing ansible and venv_playbook() for running a playbook from ansible in a venv. Those functions set the right wrappers and env vars for colours etc.

mattt416:
@hughsaunders OK all updated.  It looks like the run is a bit different now as Jenkins will restart while the job is running, before it would finish and then restart.  It seems alright though as the job just finishes once Jenkins is back, but it's a bit alarming to see at first.

playbooks/install_jenkins_plugins.yml
```diff
@@ -0,0 +1,106 @@
+---
+- hosts: localhost
+  vars:
+    jenkins_credentials:
+      url_username: "{{ ansible_env.JENKINS_USERNAME }}"
+      url_password: "{{ ansible_env.JENKINS_API_KEY }}"
+      url: "{{ ansible_env.JENKINS_URL }}"
```
hughsaunders: Couldn't see where this is set, but its a Jenkins global env var.

PR #17, opened 2017-02-09
CIT Stages: Add tempest testing stage

hughsaunders:
* Runs tempest playbook

 * Runs openstack-tempest-gate to execute tempest tests.

 * Adds try/catch within the single use slave block so that artifacts
   are retrieved before that node is destroyed

 * Adds shell script for collecting and compressing relevant logs and
   config. Compression is inconvenient but we don't have much space on
   the Jenkins master.

 * Adds tempest test result collection for display in the jenkins ui.

Connects rcbops/u-suk-dev#1130
Co-Authored-By: Matt Thompson <mattt@defunct.ca>

### DIFF
--- a/pipeline-steps/common.groovy
+++ b/pipeline-steps/common.groovy
@@ -208,4 +208,40 @@ def gen_instance_name(){
   return instance_name
 }
 
+def archive_artifacts(){
+  try{
+    sh """#!/bin/bash
+    d="artifacts_\${BUILD_TAG}"
+    mkdir -p \$d
+
+    # logs and config from the single use slave
+    mkdir -p \$d/\$HOSTNAME/log
+    cp -rp /openstack/log/\$HOSTNAME-* \$d/\$HOSTNAME/log ||:
+    cp -rp /etc/ \$d/\$HOSTNAME/etc
+    cp -rp /var/log/ \$d/\$HOSTNAME/var_log
+
+    # logs and config from the containers
+    while read c
+    do
+      mkdir -p \$d/\$c/log
+      cp -rp /openstack/log/\$c/* \$d/\$c/log ||:
+      cp -rp /var/lib/lxc/\$c/rootfs/etc \$d/\$c ||:
+    done < <(lxc-ls)
+
+    # compress to reduce storage space requirements
+    tar cvjf "\$d".tar.bz2 \$d > artifact_index.txt
+    """
+  } catch (e){
+    print(e)
+    throw(e)
+  } finally{
+    // still worth trying to archiveArtifacts even if some part of
+    // artifact collection failed.
+    archiveArtifacts(
+      allowEmptyArchive: true,
+      artifacts: "artifacts_${env.BUILD_TAG}.tar.bz2,artifact_index.txt"
+    )
+  }
+}
+
 return this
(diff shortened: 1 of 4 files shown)


rpc-jobs/RPC-AIO.yml
```diff
@@ -2,20 +2,26 @@
     name: RPC_AIO
     project-type: workflow
     concurrent: true
+    properties:
+      - build-discarder:
+          num-to-keep: 50
```
mattt416: Do we really need to keep that many builds?

pipeline-steps/common.groovy
```diff
@@ -208,4 +208,44 @@ def gen_instance_name(){
...
+      cp -rp /openstack/log/\$c/* \$d/\$c/log ||:
+      cp -rp /var/lib/lxc/\$c/rootfs/etc \$d/\$c ||:
+    done < <(lxc-ls)
+
+    # compress to reduce storage space requirements
+    tar cjf "\$d".tar.bz2 \$d
+
+    # create an index so people can see whats available without downloading
+    # and unarchiving the tar.
+    find \$d > artifact_index.txt
```
mattt416: Could you just `tar cjvf` above, and redirect output to artifact_index.txt?

pipeline-steps/common.groovy
```diff
@@ -208,4 +208,44 @@ def gen_instance_name(){
...
+    """
+  } catch (e){
+    print(e)
+    throw(e)
+  } finally{
+    // still worth trying to archiveArtifacts even if some part of
+    // artifact collection failed.
+    archiveArtifacts(
+      allowEmptyArchive: true,
+      artifacts: 'artifacts_*.tar.bz2,artifact_index.txt'
```
melissa-kam: Do we want to archive all the artifacts here? This would include logs from old test runs too, right?

pipeline-steps/tempest.groovy
```diff
@@ -0,0 +1,41 @@
...
+    stage: {
+      tempest_install()
+      try{
+        tempest_run()
+      } catch (e){
+        print(e)
+        throw(e)
+      } finally{
+        sh "cp /openstack/log/*utility*/**/*tempest*.xml . ||:"
+        junit allowEmptyResults: true, testResults: '**/*tempest*.xml'
```
melissa-kam: Same here. We might want to clean up any old results here before we publish the new ones.

pipeline-steps/common.groovy
mattt416: This runs on the one-time slaves right?

melissa-kam: So generally yes, but it can also run on a slave that skipped the cleanup step or is paused before cleanup where you would want to rerun the tests. Additionally, we could reuse this for lab environments, which would also likely have multiple test runs.

hughsaunders: During testing I have had it find artifacts for multiple runs, which is annoying, so I'll clean that up.

rpc-jobs/RPC-AIO.yml
hughsaunders: done

pipeline-steps/common.groovy
hughsaunders: done

PR #19, opened 2017-02-09
Install urllib3

mattt416:
CIT Jenkins has a python27-python-libs package w/ ssl that does not
support create_default_context and SSLContext, which are needed for
SNI handling.  ansible.module_utils.urls will fall back to using
urllib3 for SNI handling, however it uses a method that is no longer
available in urllib3.  As such, we have to install 1.18.1 specifically
so that ansible.module_utils.urls will detect the SNI support
correctly.

Side note, there is an ansible PR [1] to update
ansible.module_utils.urls accordingly.

[1] https://github.com/ansible/ansible/pull/19962

Connects https://github.com/rcbops/u-suk-dev/issues/1178


mattt416:
Note that even with this merged the job will fail because the cert on https://updates.jenkins-ci.org/ actually expired today.  This has resulted in a fair bit of head scratching.

PR #20, opened 2017-02-10
Dynamically build pyrax config

mattt416:
Currently we store the pyrax config as a credential but we should store
individual cloud credentials instead.  This commit builds a pyrax
config and writes it to temporary space using existing credentials.

Additionally, we remove some unnecessary indentation in create and
cleanup in pubcloud.groovy.

### DIFF
--- a/pipeline-steps/common.groovy
+++ b/pipeline-steps/common.groovy
@@ -244,4 +244,19 @@ def archive_artifacts(){
   }
 }
 
+def writePyraxCfg(Map args){
+  cfg = """[rackspace_cloud]
+username = ${args.username}
+api_key = ${args.api_key}
+"""
+
+  tmp_dir = pwd(tmp:true)
+  pyrax_cfg = "${tmp_dir}/.pyrax.cfg"
+  sh """
+    echo "${cfg}" > ${pyrax_cfg}
+  """
+
+  return pyrax_cfg
+}
+
 return this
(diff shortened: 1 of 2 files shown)


mattt416:
This isn't ready to merge yet -- some credentials will need to be created before we do.  We'll need to:

1. create dev_pubcloud_username credential
2. create dev_pubcloud_api_key credential
3. create dev_pubcloud_tenant credential

We can then:

1. update anything using dev_pubcloud_tennant to use dev_pubcloud_tenant
2. delete dev_pubcloud_tennant credential
3. update anything using dev_pubcloud_user_key to use dev_pubcloud_username and dev_pubcloud_api_key instead
4. delete dev_pubcloud_user_key credential
5. delete RPCJENKINS_RAXRC credential

pipeline-steps/common.groovy
```diff
@@ -208,4 +208,19 @@ def gen_instance_name(){
   return instance_name
 }
 
+def writePyraxCfg(Map args){
+  cfg = """[rackspace_cloud]
+username = ${args.username}
+api_key = ${args.api_key}
+"""
+
+  tmp_dir = pwd(tmp:true)
```
mattt416: @melissa-kam you pointed out before the issue with writing secret files to the workspace (since the workspace can be visible through the UI) -- any concerns writing this to the temporary space?

melissa-kam: None from me!

PR #21, opened 2017-02-13
Credential cleanup

mattt416:
This commit oves from using the dev_pubcloud_user_key user/password
credential to using dev_pubcloud_username and dev_pubcloud_api_key
credential strings instead.

We also use the newly created credential dev_pubcloud_tenant_id
instead of dev_pubcloud_tennant.

### DIFF
--- a/pipeline-steps/aio_prepare.groovy
+++ b/pipeline-steps/aio_prepare.groovy
@@ -19,13 +19,16 @@ def prepare(){
         } //color
       } //dir
       withCredentials([
-        usernamePassword(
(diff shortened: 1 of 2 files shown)


mattt416:
Test here: https://rpc.jenkins.cit.rackspace.net/job/RPC_AIO/281/

mattt416:
The above test is failing because the job has holland bits in its config, while the holland PR hasn't actually merged yet.

mattt416:
The failure in the above job is a known issue, and should be fixed in https://github.com/rcbops/rpc-gating/pull/22.

PR #22, opened 2017-02-13
Multi-Node AIO: Tempest Tests

melissa-kam:
I couldn't find a way for Jenkins to switch context from the host node to a VM on the host node, so I added some parameters in order to have the same commands work for both AIO and multi-node AIO. 

I did some separate work not in this PR to run openstack_tempest_gate.sh using ansible, similar to how this script does it: https://github.com/openstack/openstack-ansible/blob/stable/newton/scripts/run-tempest.sh If that's a more preferable way, I can close this PR and open one with those changes.

Connects https://github.com/rcbops/u-suk-dev/issues/1093

PR #23, opened 2017-02-14
Ignore failure to remove tempest results.

hughsaunders:
If there isn't an existing results file, it doesn't matter that it
couldn't be removed.

Connects rcbops/u-suk-dev#1130


hughsaunders:
@melissa-kam already fixed this in #22

PR #24, opened 2017-02-14
Add multi-node AIO MaaS deployment/verification

jpmontez:
Add `deploy_maas` function which configures MaaS, creates entities, and runs the MaaS setup and verification playbooks.

Add MaaS entitiy cleanup playbook to the main cleanup function.

Connects https://github.com/rcbops/u-suk-dev/issues/1125

### DIFF
--- a/pipeline-steps/aio_prepare.groovy
+++ b/pipeline-steps/aio_prepare.groovy
@@ -19,33 +19,10 @@ def prepare(){
           } //env
         } //color
       } //dir
-      withCredentials([
-        string(
-          credentialsId: "dev_pubcloud_username",
-          variable: "PUBCLOUD_USERNAME"
-        ),
-        string(
-          credentialsId: "dev_pubcloud_api_key",
-          variable: "PUBCLOUD_API_KEY"
-        ),
-        string(
-          credentialsId: "dev_pubcloud_tenant_id",
-          variable: "PUBCLOUD_TENANT_ID"
-        )
-      ]){
-        dir("/opt/rpc-gating"){
-          git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
-        } //dir
-        dir("/opt/rpc-gating/playbooks"){
-          common.install_ansible()
-          common.venvPlaybook(
-            playbooks: ["aio_config.yml"],
-            venv: ".venv",
-            args: [ "-i inventory" ]
-          )
-        } //dir
-      } //withCredentials
+      common.prepareConfigs(
+        deployment_type: "aio"
+      )
     } //stage param
-  ) //conditional stage
+  )
 }
 return this;
--- a/pipeline-steps/common.groovy
+++ b/pipeline-steps/common.groovy
@@ -269,4 +269,36 @@ api_key = ${args.api_key}
   return pyrax_cfg
 }
 
+def prepareConfigs(Map args){
+  withCredentials([
+    string(
+      credentialsId: "dev_pubcloud_username",
+      variable: "PUBCLOUD_USERNAME"
+    ),
+    string(
+      credentialsId: "dev_pubcloud_api_key",
+      variable: "PUBCLOUD_API_KEY"
+    ),
+    string(
+      credentialsId: "dev_pubcloud_tenant_id",
+      variable: "PUBCLOUD_TENANT_ID"
+    )
+    ]){
+      dir("/opt/rpc-gating"){
+        git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
+      } //dir
+      dir("/opt/rpc-gating/playbooks"){
+        common.install_ansible()
+        common.venvPlaybook(
+          playbooks: ["aio_config.yml"],
+          venv: ".venv",
+          args: [
+            "-i inventory",
+            "--extra-vars \"@vars/${args.deployment_type}.yml\""
+          ]
+        )
+      } //dir
+  } //withCredentials
+}
+
 return this
--- a/pipeline-steps/maas.groovy
+++ b/pipeline-steps/maas.groovy
@@ -1,0 +1,67 @@
+def prepare(Map args) {
+  common.conditionalStage(
+    stage_name: 'Prepare MaaS',
+    stage: {
+      withCredentials([
+        string(
+          credentialsId: "dev_pubcloud_username",
+          variable: "PUBCLOUD_USERNAME"
+        ),
+        string(
+          credentialsId: "dev_pubcloud_api_key",
+          variable: "PUBCLOUD_API_KEY"
+        ),
+        string(
+          credentialsId: "dev_pubcloud_tenant_id",
+          variable: "PUBCLOUD_TENANT_ID"
+        )
+      ]){
+        dir("rpc-gating/playbooks"){
+          pyrax_cfg = common.writePyraxCfg(
+            username: env.PUBCLOUD_USERNAME,
+            api_key: env.PUBCLOUD_API_KEY
+          )
+          withEnv(["RAX_CREDS_FILE=${pyrax_cfg}"]){
+            common.venvPlaybook(
+              playbooks: ['aio-maas-entities.yml'],
+              venv: ".venv",
+              args: [
+                "-e server_name=\"${args.instance_name}\""
+              ],
+              vars: args
+            )
+          } // withEnv
+        } // dir
+      } // withCredentials
+    } // stage
+  ) // conditionalStage
+}
+
+def deploy() {
+  common.conditionalStage(
+    stage_name: 'Setup MaaS',
+    stage: {
+      common.prepareConfigs(
+        deployment_type: "onmetal"
+      )
+      common.openstack_ansible(
+        path: '/opt/rpc-openstack/rpcd/playbooks',
+        playbook: 'setup-maas.yml'
+      ) //openstack_ansible
+    } //stage
+  ) //conditionalStage
+}
+
+def verify() {
+  common.conditionalStage(
+    stage_name: 'Verify MaaS',
+    stage: {
+      common.openstack_ansible(
+        path: '/opt/rpc-openstack/rpcd/playbooks',
+        playbook: 'verify-maas.yml'
+      ) //openstack_ansible
+    } //stage
+  ) //conditionalStage
+}
+
+return this;
--- a/pipeline-steps/pubcloud.groovy
+++ b/pipeline-steps/pubcloud.groovy
@@ -51,7 +51,7 @@ def create(Map args){
 
 /* Remove public cloud instances
  */
-def cleanup(){
+def cleanup(Map args){
   withEnv(['ANSIBLE_FORCE_COLOR=true']){
     withCredentials([
       string(
@@ -75,12 +75,17 @@ def cleanup(){
         )
         withEnv(["RAX_CREDS_FILE=${pyrax_cfg}"]){
           common.venvPlaybook(
-            playbooks: ['cleanup_pubcloud.yml'],
+            playbooks: ['aio-maas-cleanup.yml',
+                        'cleanup_pubcloud.yml'],
             venv: ".venv",
             args: [
               "--private-key=\"${env.JENKINS_SSH_PRIVKEY}\"",
             ],
-            vars: ["instance_name": instance_name]
+            vars: [
+              "instance_name": instance_name,
+              "server_name": args.server_name,
+              "region": args.region
+            ]
           )
         } // withEnv
       } // directory
@@ -122,7 +127,11 @@ def delPubCloudSlave(Map args){
     step_name: 'Cleanup',
     step: {
       ssh_slave.destroy()
-      cleanup()
+      cleanup (
+        instance_name: instance_name,
+        server_name: instance_name,
+        region: env.REGION,
+      )
     } //stage
   ) //conditionalStage
 }
--- a/rpc-jobs/multi-node-aio.yml
+++ b/rpc-jobs/multi-node-aio.yml
@@ -2,7 +2,7 @@
     name: 'OnMetal_Multi_Node_AIO'
     project-type: workflow
     build-discarder:
-        days-to-keep: 30
+      days-to-keep: 30
     concurrent: true
     parameters:
       - rpc_gating_params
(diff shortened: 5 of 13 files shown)


jpmontez:
CIT Jenkins results: https://rpc.jenkins.cit.rackspace.net/job/OnMetal_Multi_Node_AIO/81/console

One thing to note about the results is that the Nova MaaS checks fail. According to Melissa, this has happened in the past with MaaS testing, but didn't impede results. I'm looking into what underlying issue there is with the product, besides the playbooks being run successfully.

For that reason, I think the PR is ready for review, since the criteria's being met. I'm planning on raising another card/issue on the Nova MaaS bug itself.

pipeline-steps/deploy.groovy
```diff
@@ -35,4 +35,62 @@ def deploy_sh(Map args) {
...
+maas_fqdn_extension: .${env.NODE_NAME}
+memory_used_percentage_warning_threshold: 99.0
+memory_used_percentage_critical_threshold: 99.5
+net_max_speed: 1000
+lb_name: \${HOSTNAME}
+
+keystone_service_adminurl: http://172.29.236.200:5000/v3
+keystone_admin_user_name: admin
+keystone_admin_tenant_name: admin
+keystone_auth_admin_password: \${ADMIN_PWD}" >> /etc/openstack_deploy/user_rpco_variables_overrides.yml'
```
mattt416: Do we need to rewrite all these keystone_* vars?

jpmontez: Eh, probably not. The only one I'm wary about leaving out is the `keystone_service_adminurl` variable.

This is a direct copy/paste from @melissa-kam's MaaS job. What do you think, Melissa?

melissa-kam: I had a lot of problems before where these variables weren't getting picked up for some reason, and this was the only way I could get it to work. I'm not sure if it would still be an issue if we don't override them since that I made that job a long time ago, though, so maybe it's worth trying again without them.

pipeline-steps/deploy.groovy
```diff
@@ -35,4 +35,62 @@ def deploy_sh(Map args) {
...
+          usernameVariable: "PUBCLOUD_USERNAME",
+          passwordVariable: "PUBCLOUD_APIKEY"
+        ),
+        string(
+          credentialsId: "dev_pubcloud_tennant",
+          variable: "PUBCLOUD_TENANT_ID"
+        )
+      ]){
+        sh """
+          export ADMIN_PWD=\$(sudo grep keystone_auth_admin_password /etc/openstack_deploy/user_osa_secrets.yml  | awk '{print \$2}')
```
melissa-kam: Can you surround this export with a `set +x/set -x` so the admin password isn't visible in the UI?

jpmontez: Done.

jpmontez:
@mattt416 I've updated the credentials.

pipeline-steps/deploy.groovy
```diff
@@ -35,4 +35,62 @@ def deploy_sh(Map args) {
...
+        )
+      ]){
+        sh """
+          set +x
+          export ADMIN_PWD=\$(sudo grep keystone_auth_admin_password /etc/openstack_deploy/user_osa_secrets.yml  | awk '{print \$2}')
+          set -x
+          export HOSTNAME=\$(hostname)
+
+          sudo -E sh -c 'echo "
+rackspace_cloud_username: ${env.PUBCLOUD_USERNAME}
```
hughsaunders: If we split gating_overrides into a few vars_files includes, could you then make use of the same mechanism? https://github.com/rcbops/rpc-gating/blob/master/playbooks/aio_config.yml That playbook could be generalised to config_overrides or similar.

pipeline-steps/deploy.groovy
```diff
@@ -35,4 +35,62 @@ def deploy_sh(Map args) {
...
+maas_fqdn_extension: .${env.NODE_NAME}
+memory_used_percentage_warning_threshold: 99.0
+memory_used_percentage_critical_threshold: 99.5
+net_max_speed: 1000
+lb_name: \${HOSTNAME}" >> /etc/openstack_deploy/user_rpco_variables_overrides.yml'
+        """
+       } //withCredentials
+    } //stage
+  ) //conditionalStage
+  common.conditionalStage(
```
hughsaunders: Could the deploy stage be used in place of this? The RPC-AIO job is already doing that; https://github.com/rcbops/rpc-gating/blob/master/pipeline-steps/deploy.groovy  and https://github.com/rcbops/rpc-gating/blob/master/rpc-jobs/RPC-AIO.yml#L31-L40

melissa-kam: One thing about that is that we're going to want to have setup and verify as separate stages so that we can run just the verify stage if needed. We could call the deploy stage twice for each playbook, but I'm not sure if there's much value added when using it for running a single playbook.

hughsaunders: Fair enough. I remember thinking I should really move verify-maas out of deploy as it is a test not a deploy step. However the deploy task generates a stage per playbook, so I wouldn't gain much from moving it. 

Calling the deploy stage twice would be complicated as you'd need two playbooks variables and two stage names. Easier to not use it. 

@jpmontez maybe setup maas and verify maas could be functions, then I could reuse them in rpc-aio.

pipeline-steps/deploy.groovy
jpmontez: Yeah, that would be possible. I'll look into using that now.

pipeline-steps/deploy.groovy
jpmontez: Sure thing. I'll get that going.

playbooks/group_vars/aio.yml
```diff
@@ -0,0 +1,37 @@
...
+  # user_rpco_user_variables_defaults.yml file, this will result in
+  # migrate-yaml.py adding 'raw_multi_journal: True' in the overrides.
+  # To avoid this behavior in gate, it is overridden here.
+  # The same is true for journal_size, and maas_notification_plan.
+  raw_multi_journal: "False"
+  journal_size: 1024
+  osd_directory: "True"
+
+maas_creds:
+  rackspace_cloud_auth_url: "https://identity.api.rackspacecloud.com/v2.0/"
```
hughsaunders: Why is this not in group_vars/all with the other maas_creds keys? Same for group_vars/onmetal.yml.

pipeline-steps/aio_prepare.groovy
```diff
@@ -18,27 +18,9 @@ def prepare(){
...
-        )
-      ]){
-        dir("/opt/rpc-gating"){
-          git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
-        } //dir
-        dir("/opt/rpc-gating/playbooks"){
-          ansiblePlaybook playbook: "aio_config.yml"
-        } //dir
-      } //withCredentials
+      common.prepareConfigs(
```
hughsaunders: This should be outside the the prepare deployment conditional stage block, as nested stages don't work well with the current UI. Hopefully this situation will be better handled in blueocean. 

Also the RPC-AIO job will need `Prepare Configs` in it's default value for the STAGES param.

playbooks/group_vars/aio.yml
jpmontez: I think I have a misunderstanding on how the combining works. I want both `aio.yml` and `onmetal.yml` to inherit the MaaS credentials, seeing as they're the same. So, the only thing that should be injected further into that section is each deployment type's `rackspace_cloud_auth_url` variable.

pipeline-steps/aio_prepare.groovy
jpmontez: Noted. I've moved this out of the deployment conditional block. It's not re-pushed just yet; pending a clarification on the other comment thread.

I've also updated the default `STAGES` parameter with the save caveat.

playbooks/group_vars/aio.yml
melissa-kam: The `rackspace_cloud_auth_url` variable is the same for all deployments, though, which I think is what Hugh was getting at.

rpc-jobs/multi-node-aio.yml
```diff
@@ -45,9 +45,11 @@
       - string:
           name: STAGES
           default: |
-            Allocate Resources, Connect Slave, Prepare Multi-Node AIO, Prepare RPC Configs,
-             Deploy RPC w/ Script, Install Tempest, Tempest Tests, Prepare Horizon Selenium,
-             Horizon Tests, Prepare Kibana Selenium, Kibana Tests, Cleanup
+            Allocate Resources, Connect Slave, Prepare Multi-Node AIO,
+             Prepare RPC Configs, Deploy RPC w/ Script, Prepare MaaS Configs,
```
melissa-kam: There's still `Prepare MaaS Configs` instead of `Prepare Configs`, though that might be moot if we don't make Prepare Configs its own stage.

pipeline-steps/maas.groovy
```diff
@@ -0,0 +1,67 @@
...
+            )
+          } // withEnv
+        } // dir
+      } // withCredentials
+    } // stage
+  ) // conditionalStage
+}
+
+def deploy() {
+  common.overwriteConfig(
```
melissa-kam: I don't see this method in common.groovy. Is it supposed to be `prepareConfigs()`? Also, it may make more sense to have this method call moved into either the prepare method of this groovy script (and rename Create MaaS Entities to Prepare MaaS) or into the `Prepare RPC Configs` stage of multi_node_aio_prepare.groovy.

rpc-jobs/multi-node-aio.yml
```diff
@@ -56,6 +58,10 @@
               Prepare Multi-Node AIO
               Prepare RPC Configs
               Deploy RPC w/ Script
+              Prepare MaaS Configs
```
melissa-kam: Same here with the  `Prepare MaaS Configs`.

playbooks/group_vars/aio.yml
jpmontez: Ah, I see what you mean. I was stuck inside my head. Ha ha

pipeline-steps/common.groovy
```diff
@@ -259,4 +259,37 @@ api_key = ${args.api_key}
   return pyrax_cfg
 }
 
+def prepareConfigs(Map args){
+  conditionalStage(
```
melissa-kam: I think this would be better off not in its own stage, mainly because other stages call this method, so we'll start having things like this Prepare Configs stage nested inside of the Prepare Deployment stage. I also don't think we'll have a scenario where we would want to run Prepare Deployment but not run Prepare Configs, so it shouldn't be a problem to not have it as a conditional stage.

hughsaunders: conditionalStep can be used so that the block is still conditional, but doesn't create a nested stage.

pipeline-steps/pubcloud.groovy
```diff
@@ -121,7 +122,11 @@ def delPubCloudSlave(Map args){
     step_name: 'Cleanup',
     step: {
       ssh_slave.destroy()
-      cleanup()
+      cleanup (
```
hughsaunders: The cleanup function needs to be cleanup(Map args){ in order to accept these args. Also other jobs (eg RPC-AIO) that call cleanup will need to be modified to pass the right args.

jpmontez:
Successful run here: https://rpc.jenkins.cit.rackspace.net/job/OnMetal_Multi_Node_AIO/111/console

The verification didn't succeed, but all the steps executed correctly from an automation standpoint. Further cards can be added to address the MaaS issues themselves.

rpc-jobs/RPC-AIO.yml
```diff
@@ -74,13 +74,14 @@
       - tempest_params
       - string:
           name: STAGES
-          default: "Allocate Resources, Connect Slave, Prepare Deployment, Deploy RPC w/ Script, Install Tempest, Tempest Tests, Cleanup"
+          default: "Allocate Resources, Connect Slave, Prepare Deployment, Prepare Configs, Deploy RPC w/ Script, Install Tempest, Tempest Tests, Cleanup"
```
melissa-kam: There isn't a Prepare Configs stage anymore since it got moved back into Prepare Deployment.

jpmontez:
Good catch! I've removed that bit.

pipeline-steps/pubcloud.groovy
```diff
@@ -75,12 +75,17 @@ def cleanup(){
         )
         withEnv(["RAX_CREDS_FILE=${pyrax_cfg}"]){
           common.venvPlaybook(
-            playbooks: ['cleanup_pubcloud.yml'],
+            playbooks: ['aio-maas-cleanup.yml',
+                        'cleanup_pubcloud.yml'],
```
mattt416: Bit of a nit ... but the aio-maas-entities.yml playbook is run inside one of the maas stages. Does it matter that aio-maas-cleanup.yml is always run even if a maas stage isn't selected?

pipeline-steps/maas.groovy
```diff
@@ -0,0 +1,67 @@
...
+        )
+      ]){
+        dir("rpc-gating/playbooks"){
+          pyrax_cfg = common.writePyraxCfg(
+            username: env.PUBCLOUD_USERNAME,
+            api_key: env.PUBCLOUD_API_KEY
+          )
+          withEnv(["RAX_CREDS_FILE=${pyrax_cfg}"]){
+            common.venvPlaybook(
+              playbooks: ['aio-maas-entities.yml'],
```
mattt416: The aio-maas-entities.yml and aio-maas-cleanup.yml really only need to be run on a multi-node, as an AIO won't need these entities created (and will already have its own entity created at instance creation time).  I don't think you need to fix this now but this is more a note so we remember to add some logic to deal w/ this later.

pipeline-steps/pubcloud.groovy
melissa-kam: It just checks that the entities are absent, so there shouldn't be any issues if they never existed in the first place.

jpmontez: ^ This is correct

PR #26, opened 2017-02-14
Use build.sh instead of individual scripts

melissa-kam:
Because upstream changes tend to only work when using build.sh, I switched to that instead of separate stages for each part of the prepare.

Connects https://github.com/rcbops/u-suk-dev/issues/1209


odyssey4me:
Why not change upstream to provide stages as you want it?

melissa-kam:
Yeah, my hope was that I could get it back to working with individual stages like before and also update the stable/newton branch, but this was my stopgap until I can get that work done.

PR #29, opened 2017-02-17
Add Mitaka Horizon selenium integration tests

melissa-kam:
The Horizon integration test suite is no longer supported in Newton, and there's only a stable branch going back to Mitaka, so these particular tests will run only for Mitaka builds.

Connects https://github.com/rcbops/u-suk-dev/issues/1124

### DIFF
--- a/pipeline-steps/horizon.groovy
+++ b/pipeline-steps/horizon.groovy
@@ -1,0 +1,83 @@
+def horizon_integration(){
+  common.conditionalStage(
+    stage_name: "Prepare Horizon Selenium",
+    stage: {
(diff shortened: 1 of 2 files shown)


melissa-kam:
Got the tests to run, though there are a lot of failures, some of which were known issues. I ran against the same node with the old QE Jenkins and got similar results, though, so I don't think there should be any problems with the setup and execution of the tests at any rate.

https://rpc.jenkins.cit.rackspace.net/job/mkam-OnMetal_Multi_Node_AIO/49/
http://jenkins.openstack-ansible.com/job/Horizon-Integration-Tests/43/

PR #32, opened 2017-02-21
Remove variable file for multi node AIO setup

melissa-kam:
The multi-node setup has a variables.sh file now that takes precedent over environment variables, so I'm removing that file completely. I've got a review open to fix setup-host.sh, so once that merges, I'll test this change out.

https://review.openstack.org/#/c/436700/
https://review.openstack.org/#/c/437040/

Connects https://github.com/rcbops/u-suk-dev/issues/1274


melissa-kam:
Successful run: https://rpc.jenkins.cit.rackspace.net/job/mkam-OnMetal_Multi_Node_AIO/69/

I'm deploying RPC on it right now too in a separate job run, but the setup is passing now.

PR #35, opened 2017-02-22
[WIP] 1080 Global Library and RPC-Gating lint job

hughsaunders:
This PR paves the way for triggering AIOs via Jenkinsfile.

* Adds rpc-gating as a 'Global Library' which means that scripts are automatically loaded and available to all pipelines. This is very useful for Jenkinsfiles as they will be placed in other repos (eg rpc-openstack) and wouldn't have direct access to rpc-gating/pipeline-steps. Using a global library prevents boilerplate code that clones rpc-gating and loads the various steps. This change also removes all the current load statements as they are no longer required.
* Adds a runonpublcoud function that allows a user to easily run a script or some steps on a public cloud slave. The RPC-AIO and Single_Use_Slave_Example jobs have been updated to reflect this.
* Adds a Jenkinsfile to rpc-gating, which will be used to test PRs to this repo. Currently it runs lint.sh

In addition to the change in this commit, I have created a github organisation folder in Jenkins for rcbops, this will scan all repos looking for Jenkinsfile, and will build branches and PRs found to contain one. **The job that has run against this PR is a good demonstration of this mechanism working.**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+FROM ubuntu
+RUN apt-get update && apt-get install -y groovy2 python-pip build-essential python-dev libssl-dev
+RUN apt-get install -y libffi-dev
+RUN apt-get install -y sudo
+RUN pip install jenkins-job-builder ansible
+RUN useradd jenkins --shell /bin/bash --create-home --uid 500
+RUN echo "jenkins ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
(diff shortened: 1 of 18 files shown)


hughsaunders:
This is quite a wide-ranging change so I'm running an AIO to ensure they aren't broken. https://rpc.jenkins.cit.rackspace.net/job/RPC_AIO/289/console Don't merge if that doesn't pass.

PR #36, opened 2017-02-23
Add Jenkinsfile for rpc-gating lint job

hughsaunders:
Required a change to the CIT jenkins slave:
echo 'other_args="-G jenkins"' >> /etc/sysconfig/docker

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,16 @@
+properties([])
+node(){
+  deleteDir()
+  stage("Prepare"){
+    checkout scm
+    lint_container = docker.build 'lint'
+  }
+  lint_container.inside {
+    stage("checkout"){
+      checkout scm
+    }
+    stage("lint"){
+      sh "sudo ./lint.sh 2>&1"
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,16 @@
+properties([])
+node(){
+  deleteDir()
+  stage("Prepare"){
+    checkout scm
+    lint_container = docker.build 'lint'
+  }
+  lint_container.inside {
```
mattt416: `ERROR: The docker version is less than v1.8. Running a 'docker.inside' from inside a container will not work.`

Noticed that in the log, not sure if we're already inside the container and that `.inside` isn't necessary?

hughsaunders: That message is odd, I think its a misclassified warning that docker-in-docker won't work because the job actually does work. https://github.com/jenkinsci/docker-workflow-plugin/blob/68c1fd628f0918974b460bdc55598834a281093b/src/main/java/org/jenkinsci/plugins/docker/workflow/WithContainerStep.java#L141

PR #39, opened 2017-02-23
Add project and job template

mattt416:

1. Adds the basic project and converts the job to a job-template so
   that we have jobs for each series, etc.
2. Removes the PLAYBOOKS array and runs deploy_sh stage instead
3. Adds a few more parameters, such as DEPLOY_CEPH, etc.

### DIFF
--- a/rpc-jobs/RPC-AIO.yml
+++ b/rpc-jobs/RPC-AIO.yml
@@ -1,5 +1,61 @@
-- job:
-    name: RPC_AIO
+- project:
+    name: 'RPC-AIO-Jobs'
+    # Note: branch is the branch for periodics to build
+    #       branches is the branch pattern to match for PR Jobs.
+    series:
+      - liberty:
+          branch: liberty-12.2
+          branches: "liberty-.*"
+      - mitaka:
+          branch: mitaka-13.1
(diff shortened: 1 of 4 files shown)


rpc-jobs/RPC-AIO.yml
```diff
@@ -1,82 +1,135 @@
-- job:
-    name: RPC_AIO
+- defaults:
```
hughsaunders: May be worth putting global defaults in its own file as it will effect all jobs.

rpc-jobs/RPC-AIO.yml
```diff
@@ -1,82 +1,135 @@
...
+          branches: "newton-.*"
+      - master:
+          branch: master
+          branches: "master"
+    context:
+      - swift
+      - ceph:
+          DEPLOY_CEPH: "yes"
+          # Nothing uses USER_VARS atm
+          USER_VARS: |
```
hughsaunders: I think USER_VARS needs a param to ensure it gets passed though to the environment.

mattt416:
@hughsaunders I just moved the defaults under the `job-template` to avoid those causing any issues w/ other jobs.  Is that alright?

PR #41, opened 2017-02-23
Fix deleteDir error in lint job

hughsaunders:
Jenkins bind mounts the workspace into the docker instance, so if the
root user within the container creates uid 0 files/dirs, then the
jenkins user outside the instance can't clean them up. This causes
problems with the deleteDir() step.

Instead of running lint jobs as root, ensure the jenkins user fully owns
the workspace.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,7 +10,7 @@ node(){
       checkout scm
     }
     stage("lint"){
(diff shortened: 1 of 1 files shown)


hughsaunders:
Have build this PR 4x and not had any deleteDir() problems
screen shot 2017-02-23 at 20 18 10

PR #43, opened 2017-02-24
Add Upgrade stage

mattt416:

1. Adds an UPGRADE_FROM_REF parameter to params.yml
2. Adds an upgrade stage to deploy.groovy and adds TERM=linux to deploy
   environment vars which needs to be set by one of the upgrade scripts
3. Updates common.groovy to strip out ansible args that don't work
   across all ansible versions
4. Updates RPC-AIO.yml to set a default STAGE var and overrides it on
   upgrades with the necessary stage
5. Updates RPC-AIO.yml to exclude upgrades on certain series since we
   don't test upgrades across all
5. Updates RPC-AIO.yml to manually apply an iptables rule, which is
   necessary since we run deploy.sh with DEPLOY_OA=no and
   DEPLOY_AIO=yes and then DEPLOY_OA=yes and DEPLOY_AIO=no (these
   combinations means that it will never get applied via deploy.sh)

Connects rcbops/u-suk-dev#1270

### DIFF
--- a/pipeline-steps/aio_prepare.groovy
+++ b/pipeline-steps/aio_prepare.groovy
@@ -3,7 +3,11 @@ def prepare(){
     stage_name: "Prepare Deployment",
     stage: {
       dir("/opt/rpc-openstack"){
-        git branch: env.RPC_BRANCH, url: env.RPC_REPO
+        if (env.STAGES.contains("Upgrade")) {
+          git branch: env.UPGRADE_FROM_REF, url: env.RPC_REPO
+        } else {
+          git branch: env.RPC_BRANCH, url: env.RPC_REPO
+        }
         sh "git submodule update --init"
         ansiColor('xterm'){
           withEnv([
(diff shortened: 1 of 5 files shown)


mattt416:
Finally got a pass running this upgrade, but had to skip the tempest tests.  :(  Re-running a test now w/ tempest tests in place.

mattt416:
Finally got tempest tests passing (on initial deploy), looks like it wasn't dropping the iptables rule inside the neutron agents container.

hughsaunders:
@mattt416 As a headsup, I'm planning to remove the JJB matrix, in favour of calling RPC-AIO with different params for each situation:
- Removal: https://github.com/rcbops/rpc-gating/compare/1313_Jenkinsfile 
- Replacement logic in Jenkinsfile: https://github.com/rcbops/rpc-openstack/blob/c53b07d85efabd16d83b903de084b08afe4e2a4e/Jenkinsfile#L147-L230

I'll add upgrade logic based on your changes to the JJB matrix.

hughsaunders:
Added upgrade logic here: https://github.com/rcbops/rpc-openstack/compare/1313_Jenkinsfile#diff-58231b16fdee45a03a4ee3cf94a9f2c3R228

PR #46, opened 2017-02-28
Run tempest with 3 workers

melissa-kam:
Multi-node AIO: https://rpc.jenkins.cit.rackspace.net/view/Multi-Node%20AIO/job/mkam-OnMetal_Multi_Node_AIO/117/consoleFull

AIO: https://rpc.jenkins.cit.rackspace.net/view/Multi-Node%20AIO/job/mkam-OnMetal_Multi_Node_AIO/118/consoleFull

Connects https://github.com/rcbops/u-suk-dev/issues/1268

### DIFF
--- a/pipeline-steps/tempest.groovy
+++ b/pipeline-steps/tempest.groovy
@@ -11,7 +11,6 @@ def tempest_run(Map args) {
     ${args.wrapper} lxc-attach \
       --keep-env \
       -n \$utility_container \
-      -v RUN_TEMPEST_OPTS=${env.RUN_TEMPEST_OPTS} \
       -- /opt/openstack_tempest_gate.sh \
       ${env.TEMPEST_TEST_SETS}
   """
@@ -23,7 +22,8 @@ def tempest_run(Map args) {
  */
 def tempest(Map args){
   if (args != null && args.containsKey("vm")) {
-    wrapper = "sudo ssh -T -oStrictHostKeyChecking=no ${args.vm} "
+    wrapper = "sudo ssh -T -oStrictHostKeyChecking=no ${args.vm} \
+                RUN_TEMPEST_OPTS=\\\"${env.RUN_TEMPEST_OPTS}\\\" TESTR_OPTS=\\\"${env.TESTR_OPTS}\\\" "
     copy_cmd = "scp -o StrictHostKeyChecking=no -p  -r infra1:"
   } else{
     wrapper = ""
--- a/pipeline-steps/tempest.groovy
+++ b/pipeline-steps/tempest.groovy
@@ -6,14 +6,16 @@ def tempest_install(){
 }
 
 def tempest_run(Map args) {
-  sh """#!/bin/bash
-    utility_container="\$(${args.wrapper} lxc-ls |grep -m1 utility)"
+  def output = sh (script: """#!/bin/bash
+  utility_container="\$(${args.wrapper} lxc-ls |grep -m1 utility)"
     ${args.wrapper} lxc-attach \
       --keep-env \
       -n \$utility_container \
       -- /opt/openstack_tempest_gate.sh \
       ${env.TEMPEST_TEST_SETS}
-  """
+  """, returnStdout: true)
+  print output
+  return output
 }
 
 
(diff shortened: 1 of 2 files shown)


mattt416:
@melissa-kam neither of those jobs had a tempest pass -- likely for other reasons -- but do we want to see tempest passing 100% before we merge this one?

melissa-kam:
@mattt416 The multi-node failure is a test I'm working on getting skipped with https://github.com/rcbops/u-suk-dev/issues/1299 and the second one is actually testing an AIO server I built a while ago, but I just updated the same job to temporarily run AIO tests. Not sure why that one has so many failures, but I can make a new AIO server and see if I get the same issues.

mattt416:
@melissa-kam I'll spin up an AIO test against this PR also, so we can be sure it's good to go.

mattt416:
Getting a varying number of failures running the AIO tests w/ this concurrency level :(  This seemed to work well for your multinode job -- do we want to leave these two vars empty by default and then we override in each respective job?  So for RPC-AIO.yml, we'll just set `RUN_TEMPEST_OPTS` to `--serial`.  Thoughts?

melissa-kam:
I'm currently looking into checking the output for the testr error and retrying or failing the build if it's there. Running the `scenario defcore api` test suite twice is still probably going to be faster than running them in serial.

melissa-kam:
The downside of the retries is that it prints stdout only after all the tests have run, so we can't really monitor the tests as they run. Additionally, this will only retry on a false positive. If the test fail AND there's a testr issue, the shell step throws an error so it never gets to the retry. Since 3 workers has been fine for the multi-node AIO, though, my hope is that the retry feature won't be needed very often.

AIO in serial: https://rpc.jenkins.cit.rackspace.net/job/mkam-AIO/9/
Concurrent with a retry: https://rpc.jenkins.cit.rackspace.net/job/mkam-OnMetal_Multi_Node_AIO/165/consoleFull

hughsaunders:
Did you have a look at stestr? From [the docs](http://stestr.readthedocs.io/en/latest/MANUAL.html#listing-tests) it looks like it has the functionalities we need. I'm not sure how much work it is to integrate though. Stestr will allocate tests to CPUs according to how long each test takes, but only for tests that already have results in the repository (otherwise they are allocated round robin). So test execution may be quicker if we pre-populate the testr/stestr repo with a passing set of tests.

hughsaunders:
It seems upstream haven't switched to stestr yet, at least not in tempest run, so maybe its a bit early. Also that "Race in testr accounting"  line has had a few repo moves recently, it was originally in tempest, then moved to tempest-lib, and is now in os-testr. I asked in #openstack-qa if stestr will replace testr in tempest run, but will probably have to wait for this afternoon for an answer.

hughsaunders:
Apparently the race is fixed in latest testr:
```
14:16:30 mtreinis+| gmann, hughsaunders: that race should be fixed in both testr and stestr
14:17:02 mtreinis+| we left the code in subunit-trace because you can hit it with older versions of testr
14:17:09 hughsaun+| mtreinish: in latest released version of testr? or need to use head?
14:17:30 mtreinis+| it's been in the past couple of releases, so if you use the latest release it should be fine
14:17:43 mtreinis+| gmann: I don't really have a timetable for switching everything over at this point, we're still shaking out bugs in the ui and stuff
14:18:04 hughsaun+| mtreinish: thanks
14:18:19 mtreinis+| hughsaunders: this was the fix iirc: https://github.com/mtreinish/testrepository/commit/8d6ff626377e8ad6d248bc02ab85cf9dc3488471
```

melissa-kam:
I wanted to add a print statement to say that the tests are currently running, but I forgot that part of the tempest execution isn't captured, so I can't add a print anything unless I add it to openstack_tempest_gate.sh. This is the last thing that is shown before the tests run, though:
```
+ test_list_summary='defcore (295 tests)'
+ echo 'Using test list defcore (295 tests)'
+ ./run_tempest.sh --no-virtual-env -- --load-list test_list --concurrency 10
```

I created https://github.com/rcbops/u-suk-dev/issues/1350 for the follow up work.

PR #49, opened 2017-03-02
Add maas stages to AIO job

mattt416:
Note that we also add the tasks that need to run post upgrade inside
an if statement to ensure they don't run again outside of an upgrade
job.


mattt416:
Re-tested this after rebasing, job passed without issue.

PR #55, opened 2017-03-03
Password syncing

mattt416:
This commit adds a job which is intended to run on a clean jenkins
deployment to sync passwords from the safe to jenkins' credential
store.  The list of passwords to sync is hard-coded in
pull_passwords.py and will need to be pulled out at a later date.

At time of writing add_jenkins_cred.groovy supports text or file
credential types.

As this is a bootstrap job, we have tried to limit the amount of
credentials that already need to be in place before this runs.  The
only one that will need to be manually created are
service_account_jenkins_ssh_key and rpc-jenkins-svc-github-key.
The former is the ssh key assigned to the service account jenkins user
(in jenkins itself) which gets used when running jenkins-cli.jar, and
the latter is the GitHub private key to install pwsafe.

### DIFF
--- a/rpc-jobs/sync-credential-store.yml
+++ b/rpc-jobs/sync-credential-store.yml
@@ -1,0 +1,68 @@
+- job:
+    name: 'Sync-Credential-Store'
+    project-type: workflow
+    concurrent: false
+    parameters:
+      - rpc_gating_params
+      - string:
+          name: "PWSAFE_PROJECT_ID"
+          description: |
+            PasswordSafe Project ID
+      - password:
+          name: "SSO_USERNAME"
+          description: |
+            SSO username to connect to PasswordSafe with
+      - password:
+          name: "SSO_PASSWORD"
+          description: |
+            SSO password to connect to PasswordSafe with
+      - password:
+          name: "PIP_PWSAFE_LOCATION"
+          description: |
+            pip location to install pwsafe library from,
+            eg. git+ssh://git@repo_url/user/pwsafe.git
+
+    dsl: |
+      node(){ // CIT slave node
+        withCredentials([
+          file(
+            credentialsId: 'service_account_jenkins_ssh_key',
+            variable: 'JENKINS_SSH_KEY'
+          )
+        ]){
+          dir("rpc-gating"){
+            git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
+            dir("scripts"){
+              TMP_DIR = pwd(tmp:true)
+              withEnv([
+                "TMP_DIR=${TMP_DIR}",
+                "PWSAFE_PROJECT_ID=${PWSAFE_PROJECT_ID}",
+                "JENKINS_URL=${JENKINS_URL}",
+                "JENKINS_SSH_KEY=${JENKINS_SSH_KEY}",
+                "SSO_USERNAME=${SSO_USERNAME}",
+                "SSO_PASSWORD=${SSO_PASSWORD}"])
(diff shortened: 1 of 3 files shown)


hughsaunders:
As a headsup, theres no gate job on this build because its a PR from a fork. I disabled building fork PRs to prevent credential leaks via malicious PRs.

scripts/pull_passwords.py
```diff
@@ -0,0 +1,37 @@
+import os
+import subprocess
+import sys
+import pwsafe
+
+USERNAMES = [
+  'dev_pubcloud_api_key',
```
hughsaunders: Could we also sync:
- rpc-jenkins-svc-pat
- rpcghwebhooks

mattt416: Sure, I'll get these added.  For simplicity (and because not everything is a username/password), I've been storing username and passwords separately as secret text credentials.  I'll need to create two new credentials for rpcghwebhooks and update any JJB references to use those.  Any objections?

evrardjp:
I think this is a great idea for many things.

Too bad password safe can't handle files instead of just text secrets. If that was the case, we could used it for storing keys.
I guess we can still technically do it, but then we would have to do a special thing to map to file credentials in jenkins. 

So question, should we add this feature (credential files) or keep it that way? I'm fine with the current scope.

mattt416:
@evrardjp I didn't realise that Jenkins couldn't store multi-line strings (SSH private keys, etc.) in the secret text credential type, so I'll be updating this PR to allow multi-line strings from PasswordSafe to be stored as secret files.

rpc-jobs/sync-credential-store.yml
```diff
@@ -0,0 +1,69 @@
...
+      node(){ // CIT slave node
+        withCredentials([
+          file(
+            credentialsId: 'service_account_jenkins_ssh_key',
+            variable: 'JENKINS_SSH_KEY'
+          )
+        ]){
+          dir("rpc-gating"){
+            git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
+            common = load 'pipeline-steps/common.groovy'
```
mkam: Are you using this anywhere?

mattt416: Nope, removing!

rpc-jobs/sync-credential-store.yml
```diff
@@ -0,0 +1,68 @@
...
+                "TMP_DIR=${TMP_DIR}",
+                "PWSAFE_PROJECT_ID=${PWSAFE_PROJECT_ID}",
+                "JENKINS_URL=${JENKINS_URL}",
+                "JENKINS_SSH_KEY=${JENKINS_SSH_KEY}",
+                "SSO_USERNAME=${SSO_USERNAME}",
+                "SSO_PASSWORD=${SSO_PASSWORD}"])
+              {
+                sshagent (credentials: ['rpc-jenkins-svc-github-key']) {
+                  sh """#!/bin/bash
+                    scl enable python27 bash
```
hughsaunders: Does this work in a non-interactive context? is it equivalent to ` source /opt/rh/python27/enable`?

mattt416: I actually copied this from the Jenkins-Job-Builder job, but to answer your question I think that is the documented way of enabling the software collection.  It does work, I can change to sourcing instead -- either works for me.

hughsaunders: As it works, lets leave it.

PR #57, opened 2017-03-06
1313 jenkinsfile

hughsaunders:
RPC-Gating side of Jenkinsfile job trigger. The Jenkinsfile itself is in the RPC repo.

- Add PR refspec, so PRs can be checked out directly
- Fix DEPLOY_* quoting so they can be overridden at build time.
- Remove the JJB AIO matrix. No need for an explosion of jobs in JJB as Jenkins githb org folder plugin will create a job for each branch and pr that contain a Jenkinsfile. These Jobs call a single AIO build with the appropriate parameters for each scenario.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 *.retry
 *.swp
 *.pyc
+.lintvenv
+playbooks/roles/geerlingguy.nginx/
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,7 +10,11 @@ node(){
       checkout scm
     }
     stage("lint"){
-      sh "./lint.sh 2>&1"
-    }
-  }
-}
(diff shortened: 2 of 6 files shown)


melissa-kam:
Since there won't be the JJB matrix jobs, will there be an easy way to view the status of a particular set of AIO runs (like, if I want to see an overview of how the periodic master builds have been doing)? Or is that something that we can only get once we add the dashboard to track builds?

hughsaunders:
@melissa-kam I tried to solve this with two plugins: nested view and view job filter. That almost worked as it allows to filter builds based on params, but its too unreliable to be useful. 

I'll try again using the JJB matrix.

hughsaunders:
Closed as #58 merged

PR #58, opened 2017-03-07
1313 jenkinsfile jjb matrix

hughsaunders:
Alternative to #57
Corresponding RPCO branch: https://github.com/rcbops/rpc-openstack/pull/1991/files

### DIFF
--- a/lint.sh
+++ b/lint.sh
@@ -1,6 +1,24 @@
 #!/bin/bash
 
 rc=0
+venv=.lintvenv
+# exclude venv and git dir from linting
+fargs=(. -not -path \*${venv}\* -not -path \*.git\*)
+
+trap cleanup EXIT
+cleanup(){
+  type -t deactivate >/dev/null && deactivate
+}
+install(){
+  which virtualenv >/dev/null \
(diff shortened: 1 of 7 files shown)


lint.sh
```diff
@@ -1,6 +1,24 @@
 #!/bin/bash
 
 rc=0
+venv=.lintvenv
+# exclude venv and git dir from linting
+fargs=(. -not -path \*${venv}\* -not -path \*.git\*)
+
+trap cleanup EXIT
+cleanup(){
+  deactivate
```
melissa-kam: In the build that ran, this line isn't executing properly because of `RPC_GATING_LINT_USE_VENV=no`.

https://rpc.jenkins.cit.rackspace.net/job/rcbops/job/rpc-gating/job/PR-58/1/console
`./lint.sh: line 10: deactivate: command not found`

What's the reasoning behind not running the rpc-gating PR builds in a virtualenv?

hughsaunders: The idea was to use venv for local developer testing (eg via githooks.sh), but use docker for gate jobs for speed. 

Fixed deactivate by testing for the function before calling it.

PR #60, opened 2017-03-08
Add runonpubcloud function

hughsaunders:
High level function for running a stage on public cloud.
Also fixes a bug where instance_name was not passed to delpubcloudslave
correctly.

### DIFF
--- a/rpc-jobs/RPC-AIO.yml
+++ b/rpc-jobs/RPC-AIO.yml
@@ -105,48 +105,40 @@
     dsl: |
       // CIT Slave node
       node() {{
-        try {{
-          dir("rpc-gating") {{
-              git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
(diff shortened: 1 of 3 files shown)


rpc-jobs/RPC-AIO.yml
```diff
@@ -105,48 +105,40 @@
...
-              tempest.tempest()
-            }} catch (e) {{
-              print(e)
-              throw e
-            }} finally {{
-              common.archive_artifacts()
-            }}
-          }} // pub cloud node
-        }} catch (e) {{
-          currentBuild.result = 'FAILURE'
```
mattt416: Do we no longer need to set `currentBuild.result`?

mattt416: From chat:

```
@mattt no, because there is no top level exception handler - the exception will bubble up to jenkins itself and fail the build.
```

PR #61, opened 2017-03-08
Create periodic jobs for multi-node AIO

mkam:
This is only adding the periodic jobs for mitaka, newton, and master. I haven't deployed liberty on multi-node AIO yet since it'll take some different prepare steps so I didn't include that at all here.

Connects https://github.com/rcbops/u-suk-dev/issues/936

### DIFF
--- a/rpc-jobs/multi-node-aio.yml
+++ b/rpc-jobs/multi-node-aio.yml
@@ -1,9 +1,34 @@
-- job:
-    name: 'OnMetal_Multi_Node_AIO'
+- project:
+    name: 'Multi-Node-AIO-Jobs'
+    series:
+      - mitaka:
+          branch: mitaka-13.1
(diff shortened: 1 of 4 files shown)


rpc-jobs/multi-node-aio.yml
```diff
@@ -1,48 +1,63 @@
...
+      - 'OnMetal_Multi_Node_AIO_{series}-{context}'
+
+- job-template:
+    name: 'OnMetal_Multi_Node_AIO_{series}-{context}'
     project-type: workflow
     build-discarder:
       days-to-keep: 30
     concurrent: true
+    triggers:
+      - timed: "H 3 * * 1-5"
```
mattt416: Do we want to run these all at 3 AM?  I know the minute will be spread out but we'll still have how many instances in this matrix all running at the end of 3 AM.  Just concerned there may be capacity issues if someone is trying to spin up on-commit tests also?

PR #63, opened 2017-03-09
[WIP] Remove trigger config from AIO JJB jobs

hughsaunders:
These jobs will be triggered via github org folder and Jenkinsfile
so do not need their own triggers.

Depends on rcbops/rpc-openstack#1991

### DIFF
--- a/rpc-jobs/RPC-AIO.yml
+++ b/rpc-jobs/RPC-AIO.yml
@@ -5,16 +5,12 @@
     series:
       - liberty:
           branch: liberty-12.2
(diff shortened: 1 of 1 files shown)


hughsaunders:
@mattt416 yep

odyssey4me:
The more that we can move job configuration into the repo itself, the better it will be. Ideally the job config in rpc-gating should be largely static, with special cases being handled for particular branches in the repo. This gives a *lot* more flexibility and keeps us from having to do all sorts of hedging in this non-branched repo.

PR #66, opened 2017-03-13
Remove swift from ceph job

mattt416:
Currently, all our builds include swift even when ceph is being used.  This commit explicitly removes swift from being deployed for the ceph job so that we have at least one job where swift is not being used.  This is necessary to partially cover the use case of swift not being present in an environment.


hughsaunders:
This can't have been tested as the ceph jobs were never going to pass tests :(

hughsaunders:
Apologies for my previous (rude & deleted) comment. The issue was mine for overriding the default user_vars param which removed the tempest_service_available_swift var :(

PR #67, opened 2017-03-14
Run post-deployment steps in parallel

mkam:
The MaaS and test stages don't depend on each other, so I'm running them in parallel so that they don't block each other. We should probably revisit the Blue Ocean plugin at some point so that we can more easily view parallel steps in the UI.

https://rpc.jenkins.cit.rackspace.net/job/OnMetal_Multi_Node_AIO/171/

Connects https://github.com/rcbops/u-suk-dev/issues/1382

### DIFF
--- a/pipeline-steps/maas.groovy
+++ b/pipeline-steps/maas.groovy
@@ -21,6 +21,7 @@ def prepare(Map args) {
             username: env.PUBCLOUD_USERNAME,
             api_key: env.PUBCLOUD_API_KEY
           )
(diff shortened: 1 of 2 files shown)


mkam:
@hughsaunders Updated it to checkout the repo again before running. I think it had been working for me before since it reused workspaces that had the repo in it from previous runs. I tried to get stash to work, but it kept giving me issues with the unstash and the virtualenv, and I couldn't get excluding the venv to work either. :/

PR #69, opened 2017-03-14
Skip linting scripts/add_jenkins_cred.groovy

mattt416:
The lint job is currently failing because it cannot find classes that
are imported in scripts/add_jenkins_cred.groovy.  This commit adds a
temporary exclusion until we figure out how to best handle this
situation.


mattt416:
@hughsaunders this script is actually failing lint -- it never run the job to begin with as I creating a PR from a fork of rpc-gating.  :(

Is it better to try to stub out these classes, or just skip the file all-together?

PR #72, opened 2017-03-16
Calculate and set ANSIBLE_FORKS

mattt416:
This commit adds a function to common.groovy to calculate a reasonable #
for ANSIBLE_FORKS, and then modifies any functions that call deploy.sh
or openstack-ansible to pass in the env var ANSIBLE_FORKS.

The reason for doing this is that on rpc-openstack's liberty and mitaka
branches, deploy.sh does not correctly set forks when running
rpc-openstack playbooks.  This is resulting in a number of job failures
due to ssh connection failures because it's picking up the forks = 15
in rpcd/playbooks/ansible.cfg.

We also set env var ANSIBLE_SSH_RETRIES=3 when running deploy.sh or
openstack-ansible.  This env var is ansible 2.x and above only, so will
have no affect on liberty and mitaka branches.

Connects https://github.com/rcbops/u-suk-dev/issues/1406

### DIFF
--- a/pipeline-steps/common.groovy
+++ b/pipeline-steps/common.groovy
@@ -68,6 +68,20 @@ def venvPlaybook(Map args){
   } //withenv
 } //venvplaybook
 
+def calc_ansible_forks(){
+  def forks = sh (script: """#!/bin/bash
+    CPU_NUM=\$(grep -c ^processor /proc/cpuinfo)
+    if [ \${CPU_NUM} -lt "10" ]; then
+      ANSIBLE_FORKS=\${CPU_NUM}
+    else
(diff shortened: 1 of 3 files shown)


mattt416:
@hughsaunders still testing locally, I'll fire off a job on CIT jenkins and add the URL when done.  Because we're passing around env vars it's hard to see it actually working, short of the job failing on an ssh error or not.  :)

mattt416:
https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_liberty-ceph-periodic/16/
https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_master-ceph-periodic/22/

PR #75, opened 2017-03-17
Use ssh_retry in ansible 1.9

hughsaunders:
Upstream ansible doesn't have functionality for retrying failed ssh
connections before v2.

As we often see ssh connection failures, use a fork of ansible 1.9
that contains the ssh_retry connection plugin.

This only affects mitaka and below as newton and above use ansible v2.

Connects rcbops/u-suk-dev#1406


hughsaunders:
This is a migration of vars that are [currently used in jenkins-rpc / jenkins.propter.net](https://github.com/rcbops/jenkins-rpc/blob/master/scripts/gating_bash_lib.sh#L362-L363)

PR #80, opened 2017-03-18
Use axis specific variables for user_vars

hughsaunders:
This prevents one axis from overriding another's value. The values from
each axis are combined to form the default value for the parameter.

This resolves an issue where tempest_service_available_swift was not set
for pr jobs because the pr axis overrides the user variables set by
the ceph axis.

Related rcbops/rpc-gating#66


mattt416:
Looks like this has already been pushed to jenkins, and I'm now seeing timer jobs failing because user_zzz_gating_variables.yml has this in it:

```
...
tempest_tempest_conf_overrides:
    volume-feature-enabled:
        snapshot: true

\n\n
```

PR #82, opened 2017-03-22
Add python/container artifact build jobs

odyssey4me:
Connects https://github.com/rcbops/u-suk-dev/issues/1432

### DIFF
--- a/rpc-jobs/RPC-Artifact-Build.yml
+++ b/rpc-jobs/RPC-Artifact-Build.yml
@@ -1,0 +1,121 @@
+- project:
+    name: 'RPC-Artifact_Build-Jobs'
+    # Note: branch is the branch for periodics to build
+    #       branches is the branch pattern to match for PR Jobs.
+    series:
+      - artifacts:
+          branch: artifacts-14.0
+          branches: "artifacts-.*"
+    image:
+      - trusty:
+          IMAGE: "Ubuntu 14.04 LTS (Trusty Tahr) (PVHVM)"
+    context:
+      - Python
+      - Container
+    ztrigger:
+      - pr:
+          CRON: ""
+          STAGES: "Allocate Resources, Connect Slave, Build {context} Artifacts, Cleanup"
+      - periodic:
+          branches: "do_not_build_on_pr"
+          STAGES: "Allocate Resources, Connect Slave, Build Python Artifacts, Build Container Artifacts, Cleanup"
+          PUSH_TO_MIRROR: yes
+    jobs:
+      - 'RPC-Artifact_Build-{series}-{image}-{context}-{ztrigger}'
+
+- job-template:
+    # DEFAULTS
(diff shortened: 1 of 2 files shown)


rpc-jobs/RPC-Artifact-Build.yml
```diff
@@ -0,0 +1,88 @@
...
+      # See params.yml
+      - rpc_params:
+         RPC_BRANCH: "{branch}"
+      - rpc_gating_params
+      - single_use_slave_params:
+         IMAGE: "{IMAGE}"
+         FLAVOR: "performance2-15"
+      - string:
+          name: PUSH_TO_MIRROR
+          default: "NO"
```
hughsaunders: Probably should be ```{PUSH_TO_MIRROR}``` That variable also needs to be added to defaults at the top of job-template.

rpc-jobs/RPC-Artifact-Build.yml
```diff
@@ -0,0 +1,88 @@
...
+          description: "Set this to YES if you want to push the newly built artifacts to rpc-repo."
+      - string:
+          name: STAGES
+          default: "{STAGES}"
+          description: |
+            Pipeline stages to run CSV. Note that this list does not influence execution order.
+            Options:
+              Allocate Resources
+              Connect Slave
+              Build Artifacts
```
odyssey4me: There is actually no such stage defined as 'Build Artifacts'. Adding the groovy bits to make it seems like an unnecessary abstraction here. Thoughts?

rpc-jobs/RPC-Artifact-Build.yml
odyssey4me: Thanks, will adjust it.

rpc-jobs/RPC-Artifact-Build.yml
hughsaunders: Yeah, theres no need to use a stage for your script here. It won't work with the runonpubcloud abstraction as that also uses stages, and stages can't currently be nested. If you want to use your own stage you'd have to use a lower level method to get the single use instance. (see pubcloud.groovy).

rpc-jobs/RPC-Artifact-Build.yml
```diff
@@ -0,0 +1,110 @@
...
+    image:
+      - trusty:
+          IMAGE: "Ubuntu 14.04 LTS (Trusty Tahr) (PVHVM)"
+    context:
+      - Python
+      - Container
+    ztrigger:
+      - pr:
+          CRON: ""
+          STAGES: "Allocate Resources, Connect Slave, Build {context} Artifacts, Cleanup"
```
odyssey4me: Will using {context} here work?

odyssey4me: Apparently the JJB renders, so I guess it might: https://rpc.jenkins.cit.rackspace.net/job/Jenkins-Job-Builder/313/console

PR #90, opened 2017-03-22
Retain logs for 7 days

mattt416:
This is not ideal, but we need to keep this number low until we have
more storage allocated to the CIT instance or figure out a way to
send logs elsewhere.

### DIFF
--- a/rpc-jobs/RPC-AIO.yml
+++ b/rpc-jobs/RPC-AIO.yml
@@ -87,7 +87,7 @@
     concurrent: true
     properties:
       - build-discarder:
-          num-to-keep: 30
+          num-to-keep: 7
     parameters:
       # See params.yml
       - rpc_params:


rpc-jobs/RPC-AIO.yml
```diff
@@ -87,7 +87,7 @@
     concurrent: true
     properties:
       - build-discarder:
-          num-to-keep: 30
+          num-to-keep: 7
```
mkam: This should actually be `days-to-keep` and not `num-to-keep` if we want to do it by days. Not sure if that's what it was originally supposed to be or not.

mattt416: You are right, I was thinking days.  However, 7 days is still going to be too much once we flip on per commit gating.  I need to rethink this a little.

mattt416:
Closing, need to rethink this.

PR #93, opened 2017-03-22
Add AIO prep to Artifact builds

odyssey4me:
In order to ensure that the correct checkout is
on the remote node we need to use the AIO prep
groovy script.

Connects https://github.com/rcbops/u-suk-dev/issues/1432

### DIFF
--- a/rpc-jobs/RPC-Artifact-Build.yml
+++ b/rpc-jobs/RPC-Artifact-Build.yml
@@ -12,22 +12,37 @@
     context:
       - Python
       - Container
+      - Pipeline
     ztrigger:
       - pr:
           CRON: ""
-          STAGES: "Allocate Resources, Connect Slave, Build {context} Artifacts, Cleanup"
+          STAGES: "Allocate Resources, Connect Slave, Prepare Deployment, Build {context} Artifacts, Cleanup"
       - periodic:
           branches: "do_not_build_on_pr"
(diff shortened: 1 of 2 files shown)


odyssey4me:
Testing: https://rpc.jenkins.cit.rackspace.net/job/RPC-Artifact_Build-artifacts-trusty-Python-pr/8/

rpc-jobs/RPC-Artifact-Build.yml
```diff
@@ -12,22 +12,37 @@
...
 
 - job-template:
     # DEFAULTS
     IMAGE: "Ubuntu 14.04 LTS (Trusty Tahr) (PVHVM)"
+    DEPLOY_SWIFT: "yes"
+    DEPLOY_CEPH: "yes"
+    DEPLOY_ELK: "yes"
     CRON: "H H(9-21) * * 1-5"
-    STAGES: "Allocate Resources, Connect Slave, Build Python Artifacts, Build Container Artifacts, Cleanup"
+    CONTEXT_USER_VARS: ""
```
hughsaunders: You don't need to create these if you aren't using them. They are used in the AIO job because we set user vars from multiple Axis. However there's no harm in creating them if you might need to set user vars in future.

odyssey4me: I thought that they were required as they're included by default in rpc_params with no global defaults?

PR #94, opened 2017-03-23
Add GHPRB trigger to AIO jobs

hughsaunders:
The github pull request builder plugin can now trigger pipelines,
in a much simpler way than using a jenkinsfile/orgfolder as a parent
job.

connects rcbops/u-suk-dev#1379

### DIFF
--- a/rpc-jobs/RPC-AIO.yml
+++ b/rpc-jobs/RPC-AIO.yml
@@ -129,9 +129,29 @@
               Cleanup
     triggers:
       - timed: "{CRON}"
+      - github-pull-request:
+          org-list:
+            - rcbops
+          github-hooks: true
+          trigger-phrase: '.*recheck_cit_all.*|.*recheck_cit_{context}.*'
+          only-trigger-phrase: false
+          white-list-target-branches:
+            - "{branches}"
+          auth-id: "github_account_rpc_jenkins_svc"
+          status-context: 'CIT/{context}'
+
     dsl: |
       // CIT Slave node
       node() {{
+        /* if job is triggered by PR, then we need to set RPC_REPO and
+          RPC_BRANCH using the env vars supplied by ghprb. Ztrigger is a
+          jjb variable so uses single braces.
+        */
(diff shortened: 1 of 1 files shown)


hughsaunders:
recheck_lint

hughsaunders:
hmm, realised this needs more work, because the env vars provided by ghprb a different to those provided by org folder.

hughsaunders:
Added var translation block, but needs testing.

rpc-jobs/RPC-AIO.yml
```diff
@@ -129,9 +129,29 @@
...
     dsl: |
       // CIT Slave node
       node() {{
+        /* if job is triggered by PR, then we need to set RPC_REPO and
+          RPC_BRANCH using the env vars supplied by ghprb. Ztrigger is a
+          jjb variable so uses single braces.
+        */
+        if ("{ztrigger}" == "pr" ){{
+          env.RPC_REPO = env.ghprbAuthorRepoGitUrl
+          env.RPC_BRANCH = env.ghprbSourceBranch
```
hughsaunders: Gah this isn't right as we need to merge with targetbranch

rpc-jobs/RPC-AIO.yml
```diff
@@ -129,9 +129,29 @@
               Cleanup
     triggers:
       - timed: "{CRON}"
+      - github-pull-request:
+          org-list:
+            - rcbops
+          github-hooks: true
+          trigger-phrase: '.*recheck_all.*|.*recheck_{context}.*'
```
mattt416: How long do we want to run old and new jenkins side-by-side?  If not an immediate switch-over (which I'd recommend, quotas allowing), should we change these to `recheck_cit_*`?

hughsaunders: good point

rpc-jobs/RPC-AIO.yml
hughsaunders: Done

PR #95, opened 2017-03-23
1443 rpc gating lint

hughsaunders:


hughsaunders:
recheck_lint

hughsaunders:
recheck_lint

hughsaunders:
recheck_lint

hughsaunders:
recheck_lint

hughsaunders:
recheck_lint

hughsaunders:
it works :shipit:

rpc-jobs/jjb-setup.yml
```diff
@@ -11,17 +11,16 @@
...
-        - string:
-            name: JJB_REPO
-            default: https://github.com/rcbops/rpc-gating
-        - string:
-            name: JJB_BRANCH
-            default: master
+        - bool:
+            name: DELETE_OLD
+            description: "Remove jobs that are no longer defined."
+            default: false
```
odyssey4me: Personally I would think that you'd want this to default to 'true' so that jobs are always cleaned up. This enforces that all jobs are always cleaned up unless you specifically don't want to do that.

rpc-jobs/jjb-setup.yml
```diff
@@ -1,4 +1,24 @@
...
+        - git:
+            url: "https://github.com/rcbops/rpc-gating"
+            branches:
+                - master
+    properties:
+      - rpc-gating-github
+    triggers:
+      - github # triggered post merge, not on PR
+    builders:
+    builders:
```
mkam: Is there any reason why there's two `builders:` or is it just a typo?

rpc-jobs/jjb-setup.yml
```diff
@@ -1,4 +1,24 @@
 - job:
+    name: 'Merge-Trigger-JJB'
+    project-type: freestyle
```
mkam: The trigger job I wrote was a pipeline/workflow type instead of a freestyle (https://github.com/rcbops/rpc-gating/pull/100), but I know you were having issues making this one a workflow, so I might change mine over to a freestyle too just so that we have a consistent format for trigger jobs.

hughsaunders: This job wasnt working when I left work, so worth testing it with Matt's typo fix before basing anything else on this.

rpc-jobs/jjb-setup.yml
hughsaunders: Just a typo, should have spotted that!

rpc-jobs/jjb-setup.yml
hughsaunders: I'd prefer to use pipeline but couldn't work out how to specify a branch for post merge trigger with a pipeline job.

mkam: We can merge this as a freestyle and I can work on updating it to a pipeline!

The JJB failures was because there was an old job named Merge-Trigger-JJB that was a workflow job, so it failed when it tried to update the job as a freestyle one. I renamed the old one to Merge-Trigger-JJB-pipeline and it ran successfully.

hughsaunders: Ahhhah thanks for spotting that.

PR #98, opened 2017-03-23
Fix inline comment

mattt416:


mattt416:
Closing in favour of https://github.com/rcbops/rpc-gating/pull/99/.

PR #100, opened 2017-03-24
Trigger multi-node AIO on post merge

mkam:
Because pipeline jobs don't have an scm section where you can specify which branch and repo to watch, I made a separate trigger job to call the omna job. Pipelines use the repo and branch of the last build, so that git info is hardcoded into the trigger jobs.

Connects https://github.com/rcbops/u-suk-dev/issues/1358

### DIFF
--- a/rpc-jobs/multi-node-aio.yml
+++ b/rpc-jobs/multi-node-aio.yml
@@ -47,6 +47,15 @@
             build job: 'OnMetal_Multi_Node_AIO_{series}-xenial-post-merge'
           }}
         }} // stage
+        stage('Push to rpc-openstack'){{
+         sshagent (credentials:['rpc-jenkins-svc-github-ssh-key']){{
+           sh """
+             mkdir -p ~/.ssh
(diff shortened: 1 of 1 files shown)


rpc-jobs/multi-node-aio.yml
```diff
@@ -15,21 +15,50 @@
...
           days-to-keep: 7
+      - rpc-openstack-github
     concurrent: true
     triggers:
-      - timed: "H H(0-8) * * 1-5"
+      - github
+    dsl: |
+      node(){{
+        stage('OnMetal Multi-Node AIO') {{
+          git branch: "{branch}", url: "https://github.com/rcbops/rpc-openstack"
```
hughsaunders: Will this trigger on merges to all branches, or just the branch specified in `{branch}` ?

mkam: It will only trigger on the specified branch because it tracks the branch used by the last build.

PR #101, opened 2017-03-24
Create defaults file

mattt416:
This PR moves a bunch of job defaults to global defaults.

There is a huge net benefit of doing this, however it does mean we can
remove the raft of default variables set in RPC-AIO.yml and
RPC-Artifact-Build.yml and should we need to change REGION quickly, for
example, we can do it on place versus 3.

This needs to be re-tested once the lint job is triggering again.

Connects https://github.com/rcbops/u-suk-dev/issues/1493

### DIFF
--- a/rpc-jobs/RPC-AIO.yml
+++ b/rpc-jobs/RPC-AIO.yml
@@ -70,15 +70,6 @@
 
 - job-template:
     # DEFAULTS
-    IMAGE: "Ubuntu 14.04 LTS (Trusty Tahr) (PVHVM)"
-    DEPLOY_SWIFT: "yes"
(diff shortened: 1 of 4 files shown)


mattt416:
This needs a touch of testing before we merge, please hold off for now!

rpc-jobs/RPC-AIO.yml
```diff
@@ -1,5 +1,6 @@
 - project:
     name: 'RPC-AIO-Jobs'
+    defaults: global
```
mkam: The jobs should all use global by default so you shouldn't have to specify it.

mattt416: Removed!

PR #107, opened 2017-03-28
Timeout multi-node AIO prep

mkam:
Prevents jobs from hanging forever waiting for the VMs to come up.

Connects https://github.com/rcbops/u-suk-dev/issues/1467


mkam:
recheck_lint

PR #109, opened 2017-03-28
Add PR trigger to artifact build jobs

odyssey4me:
Currently the Artifact build jobs are only
triggered manually or periodically. This
adds PR triggers.

https://github.com/rcbops/u-suk-dev/issues/1413

### DIFF
--- a/rpc-jobs/RPC-Artifact-Build.yml
+++ b/rpc-jobs/RPC-Artifact-Build.yml
@@ -81,6 +81,16 @@
               Cleanup
     triggers:
       - timed: "{CRON}"
+      - github-pull-request:
+          org-list:
+            - rcbops
+          github-hooks: true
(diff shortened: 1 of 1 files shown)


rpc-jobs/RPC-Artifact-Build.yml
```diff
@@ -81,6 +81,16 @@
...
+      - github-pull-request:
+          org-list:
+            - rcbops
+          github-hooks: true
+          trigger-phrase: '.*recheck_cit_all.*|.*recheck_cit_{image}_{context}.*'
+          only-trigger-phrase: false
+          white-list-target-branches:
+            - "{branches}"
+          auth-id: "github_account_rpc_jenkins_svc"
+          status-context: 'CIT/{image}-{context}'
```
mkam: Should probably add "artifact" to this so that it's different from the AIO test statuses.

PR #110, opened 2017-03-28
[WIP] Skip teardown if meta key set

mattt416:
Currently, if you add the Cleanup stage to a build there is no way
clean way of preventing the teardown should you end up needing to
preserve the instance for investigation.

With this commit, you can now skip the teardown of an instance by
running the following mid-way through the build:

```
nova meta <instance> set preserve=yes
```

or

```
nova meta <instance> set preserve=true
```

For this to work, we've had to move the slave teardown to a separate
playbook so we can check for the existence of an ansible fact.  We also
create an ansible.cfg to write facts to disk so we can use these facts
in destroy-jenkins-slave.yml.  A new playbook called write_facts.yml is
added so we explicitly gather facts and write to disk.  We could
probably shimmy these into another playbook, but keeping it separate
will make it easier when we start to deploy to non-cloud instances.

Lastly, we update setup-jenkins-slave.yml by removing a bunch of defunct
tasks which never get executed as the playbook is called with a --limit
that does not include localhost.


mattt416:
Note that this will still tear down MaaS entities, etc.  I suspect if we allow those to be preserved then we'll forget to clean them up when we tear down the instance.

mattt416:
https://rpc.jenkins.cit.rackspace.net/job/Single_Use_Slave_Example/60/console (will set meta tag)
https://rpc.jenkins.cit.rackspace.net/job/Single_Use_Slave_Example/61/ (leaving unset)

mkam:
Those builds failed from what seemed to be unrelated issues, but I reran the jobs against the existing servers and 60 wasn't deleted and 61 was as expected! The only other thing I noticed is that the server is still removed as a Jenkins node. It's not difficult to add it back, but I wasn't sure if we wanted to have that skipped too or not. We don't run that script in ansible, so it'll take more effort to skip that part, though.

mkam:
I had time to do full reruns:

Cleaned: https://rpc.jenkins.cit.rackspace.net/job/Single_Use_Slave_Example/63/console
Preserved: https://rpc.jenkins.cit.rackspace.net/job/Single_Use_Slave_Example/67/console

mattt416:
@mkam not deleting the instance from jenkins is a good idea, especially if you want to re-run stages through jenkins itself.  I've started to make changes to accommodate this, but they're not quite working yet.  :)  I'll remove the WIP label once I'm done.

hughsaunders:
Cleanup is normally an optional stage, so can be disabled by removing it from the STAGES parameter. Is the aim of this change to allow the user to decide to preserve the instance after the build is initiated?

mattt416:
@hughsaunders correct!  I often find myself spinning something up to test, forgetting to remove Cleanup or thinking it won't be needed and then realising I need to do additional troubleshooting after the run has finished.  I thought this was going to be a trivial change but keeping the slave connected seems to be a bit more work.

mattt416:
# Preserving instance:

https://rpc.jenkins.cit.rackspace.net/job/Single_Use_Slave_Example/70/console

# Standard run:

https://rpc.jenkins.cit.rackspace.net/job/Single_Use_Slave_Example/71/console

mattt416:
I was hoping that you could later run the Cleanup stage in isolation and it'd tear down the slave/instance correctly, but it looks like further work will need to be done to facilitate that.  :(  Let's not merge this for a bit, still working through a few things.

b3rnard0:
@mattt416 Does this have a connected parent issue?

mattt416:
@b3rnard0 nope, this was a 1 line change that spiralled out of control.  I'm not working on this any more as there was no time allocated to it so will close it off for now and considering adding a card to backlog to revisit in the future.

PR #112, opened 2017-03-30
Add timeout to AIO Job

hughsaunders:
Timeout is set to 8 hours as upgrade jobs are typically <6hrs, so that
gives us 2 hours of buffer.

This is an absolute timeout as inactivity timeout isn't currently
supported for pipelines.

See: https://issues.jenkins-ci.org/browse/JENKINS-26521
Connects: rcbops/u-suk-dev#1492

### DIFF
--- a/rpc-jobs/RPC-AIO.yml
+++ b/rpc-jobs/RPC-AIO.yml
@@ -144,59 +144,61 @@
 
     dsl: |
       // CIT Slave node
-      node() {{
-        /* if job is triggered by PR, then we need to set RPC_REPO and
-          RPC_BRANCH using the env vars supplied by ghprb. Ztrigger is a
-          jjb variable so uses single braces.
-        */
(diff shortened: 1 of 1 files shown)


rpc-jobs/RPC-AIO.yml
```diff
@@ -144,59 +144,61 @@
...
+                "DEPLOY_ELK=${{DEPLOY_ELK}}",
+                ]
+              aio_prepare.prepare()
+              deploy.deploy_sh(environment_vars: environment_vars)
               deploy.addChecksumRule()
               maas.deploy()
               maas.verify()
               tempest.tempest()
               holland.holland()
+              if (env.STAGES.contains("Upgrade")) {{
```
hughsaunders: Not sure whats up with the diff here, I only added the timeout block and indented everything by 1.

PR #113, opened 2017-03-30
Artifact job refactor

odyssey4me:
This patch aims to refactor the Artifact jobs so
that they more effectively use stages, and share
a common structure. The detailed steps are all in
the groovy file, whereas the pipeline logic is
formed into stages in the job dsl.

An additional job is also added to build apt
artifacts. This job runs on a jenkins slave rather
than a single-use node.

A common script to get the credentials into the
environment is used for all jobs.

Connects https://github.com/rcbops/u-suk-dev/issues/1451

### DIFF
--- a/pipeline-steps/artifact_build.groovy
+++ b/pipeline-steps/artifact_build.groovy
@@ -1,71 +1,100 @@
-def python(Map args) {
+def get_rpc_repo_creds(){
+  return [
+    string(
+      credentialsId: "RPC_REPO_IP",
+      variable: "REPO_HOST"
+    ),
+    string(
+      credentialsId: "RPC_REPO_SSH_USERNAME_TEXT",
+      variable: "REPO_USER"
+    ),
+    file(
+      credentialsId: "RPC_REPO_SSH_USER_PRIVATE_KEY_FILE",
+      variable: "REPO_USER_KEY"
+    ),
+    file(
+      credentialsId: "RPC_REPO_SSH_HOST_PUBLIC_KEY_FILE",
+      variable: "REPO_HOST_PUBKEY"
+    ),
+    file(
+      credentialsId: "RPC_REPO_GPG_SECRET_KEY_FILE",
+      variable: "GPG_PRIVATE"
+    ),
+    file(
+      credentialsId: "RPC_REPO_GPG_PUBLIC_KEY_FILE",
+      variable: "GPG_PUBLIC"
+    )
+  ]
+}
+
+def apt() {
   common.conditionalStage(
-    stage_name: "Build Python Artifacts",
+    stage_name: "Build Apt Artifacts",
     stage: {
-      withEnv(args.environment_vars) {
-        withCredentials([
-          string(
-            credentialsId: "RPC_REPO_IP",
-            variable: "REPO_HOST"
-          ),
-          string(
-            credentialsId: "RPC_REPO_SSH_USERNAME_TEXT",
-            variable: "REPO_USER"
-          ),
-          file(
-            credentialsId: "RPC_REPO_SSH_USER_PRIVATE_KEY_FILE",
-            variable: "REPO_USER_KEY"
-          ),
-          file(
-            credentialsId: "RPC_REPO_SSH_HOST_PUBLIC_KEY_FILE",
-            variable: "REPO_HOST_PUBKEY"
-          )
-        ]){
-          ansiColor('xterm') {
-            dir("/opt/rpc-openstack/") {
-              sh """#!/bin/bash
-              scripts/artifacts-building/python/build-python-artifacts.sh
-              """
-            } // dir
-          } // ansiColor
-        } // withCredentials
-      } // withEnv
+      withCredentials(get_rpc_repo_creds()) {
+        prepareRpcGit()
+        ansiColor('xterm') {
+          dir("/opt/rpc-openstack/") {
+            sh """#!/bin/bash
+            scripts/artifacts-building/apt/build-apt-artifacts.sh
+            """
+          } // dir
+        } // ansiColor
+      } // withCredentials
     } // stage
   ) // conditionalStage
 }
 
-def container(Map args) {
+def python() {
+  common.conditionalStage(
+    stage_name: "Build Python Artifacts",
+    stage: {
+      pubcloud.runonpubcloud {
+        try {
+          withCredentials(get_rpc_repo_creds()) {
+            prepareRpcGit()
+            ansiColor('xterm') {
+              dir("/opt/rpc-openstack/") {
+                sh """#!/bin/bash
+                scripts/artifacts-building/python/build-python-artifacts.sh
+                """
+              } // dir
+            } // ansiColor
+          } // withCredentials
+        } catch (e) {
+          print(e)
+          throw e
+        } finally {
+          common.archive_artifacts()
+        }
+      } // pubcloud slave
+    } // stage
+  ) // conditionalStage
+}
+
+def container() {
   common.conditionalStage(
     stage_name: "Build Container Artifacts",
     stage: {
-      withEnv(args.environment_vars) {
-        withCredentials([
-          string(
-            credentialsId: "RPC_REPO_IP",
-            variable: "REPO_HOST"
-          ),
-          string(
-            credentialsId: "RPC_REPO_SSH_USERNAME_TEXT",
-            variable: "REPO_USER"
-          ),
-          file(
-            credentialsId: "RPC_REPO_SSH_USER_PRIVATE_KEY_FILE",
-            variable: "REPO_USER_KEY"
-          ),
-          file(
-            credentialsId: "RPC_REPO_SSH_HOST_PUBLIC_KEY_FILE",
-            variable: "REPO_HOST_PUBKEY"
-          )
-        ]){
-          ansiColor('xterm') {
-            dir("/opt/rpc-openstack/") {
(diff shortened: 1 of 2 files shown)


rpc-jobs/RPC-Artifact-Build.yml
```diff
@@ -138,3 +138,93 @@
...
+          default: "{branch}"
+          description: |
+            Target branch - the branch to be tested (sha1 param) will
+            be rebased against this. Also overridden by github pull request builder plugin.
+      - string:
+          name: RPC_REPO
+          default: "https://github.com/rcbops/rpc-openstack"
+          description: "RPC REPO URL"
+      - string:
+          name: RPC_REPO_BRANCH
```
hughsaunders: For consistency, we're using RPC_REPO and RPC_BRANCH in CIT jenkins.

rpc-jobs/RPC-Artifact-Build.yml
```diff
@@ -138,3 +138,93 @@
...
+      - string:
+          name: PUSH_TO_MIRROR
+          default: "YES"
+          description: "Set this to NO if you don't want to push the published local folder, the public gpg key, and the current db to the rpc mirror."
+    wrappers:
+      - ansicolor
+      - timestamps
+      - credentials-binding:
+        - text:
+            credential-id: RPC_REPO_IP
```
hughsaunders: These credentials will probably need to be added. Please add all credentials to passwordsafe and jenkins.

rpc-jobs/RPC-Artifact-Build.yml
```diff
@@ -138,3 +138,93 @@
...
+    node: ArtifactBuilder2
+    concurrent: false
+    build-discarder:
+      daysToKeep: 30
+    triggers:
+      - timed: "H 4 * * *"
+    parameters:
+      - pr-params
+      - string:
+          name: ghprbTargetBranch
```
hughsaunders: This is a parameter provided by the github pull request builder plugin, but the ghprb trigger is not enabled for this job, is that intentional?

rpc-jobs/RPC-Artifact-Build.yml
```diff
@@ -138,3 +138,93 @@
...
+- job:
+    name: JJB-artifacts-apt
+    node: ArtifactBuilder2
+    concurrent: false
+    build-discarder:
+      daysToKeep: 30
+    triggers:
+      - timed: "H 4 * * *"
+    parameters:
+      - pr-params
```
hughsaunders: As mentioned by the lintjob this macro doesn't exist, it can be added to params.yml or you can remove the line as ghprb is not listed in triggers.

rpc-jobs/RPC-Artifact-Build.yml
odyssey4me: I've added them to Jenkins. As far as I know @mattt416 has a card to ensure that they're appropriately setup in passwordsafe.

rpc-jobs/RPC-Artifact-Build.yml
odyssey4me: I just copied the old job verbatim. If it's not needed I'll remove it.

rpc-jobs/RPC-Artifact-Build.yml
odyssey4me: OK, I'll adjust that.

rpc-jobs/RPC-Artifact-Build.yml
mattt416: I have a card open to make sure they get added to the password sync tool, but you'll need to add them to passwordsafe as I have no idea what the unencrypted values are.

odyssey4me: Someone's going to have to clue me in to what passwordsafe is then. :)

rpc-jobs/RPC-Artifact-Build.yml
```diff
@@ -92,7 +96,15 @@
           auth-id: "github_account_rpc_jenkins_svc"
           status-context: 'CIT/artifact-{image}-{context}'
     dsl: |
-      // CIT Slave node
+      node('ArtifactBuilder2') {{
+        dir("rpc-gating") {{
+            git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
+            common = load 'pipeline-steps/common.groovy'
+            aio_prepare = load 'pipeline-steps/aio_prepare.groovy'
```
mattt416: Could you import aio_prepare in artifact_build.groovy since that's where it's used?  And is there even a need to run aio_prepare.prepare() for this apt piece?

pipeline-steps/artifact_build.groovy
```diff
@@ -1,71 +1,105 @@
...
-              """
-            } // dir
-          } // ansiColor
-        } // withCredentials
-      } // withEnv
+      pubcloud.runonpubcloud {
+        try {
+          withEnv(args.environment_vars) {
+            get_rpc_repo_creds() {
+              ansiColor('xterm') {
```
mattt416: This doesn't run an aio_prepare.prepare() like the other two functions do, meaning this stage cannot be run by itself.  I'm not sure if that's a problem at all (not all our stages work in isolation) but just mentioning it while I see it.

rpc-jobs/RPC-Artifact-Build.yml
mattt416: Grab me tomorrow and I'll get you squared away!

pipeline-steps/artifact_build.groovy
odyssey4me: Yeah, that was an oversight - thanks for picking it up!

rpc-jobs/RPC-Artifact-Build.yml
odyssey4me: Yeah, we need the rpc-openstack repo to be there. I've submitted https://github.com/rcbops/rpc-gating/pull/114 to break out the git repo prep as that's all we want for all the artifact build jobs.

odyssey4me: I think it's better to keep all the imports in the job definition instead of spreading them all around in the groovy files. Well, I think we should stick to one standard as that makes reviewing and maintaining a lot easier IMO.

pipeline-steps/artifact_build.groovy
```diff
@@ -1,71 +1,106 @@
-def python(Map args) {
+def get_rpc_repo_creds(){
```
hughsaunders: I don't think this will work as is.  I see two options:

1. Rename to with_rpc_repo_creds, accept a body closure and execute that within the withCredentials block (see [runonpubcloud](https://github.com/rcbops/rpc-gating/blob/master/pipeline-steps/pubcloud.groovy#L140))
1. Return the array of arguments to be used in withCredentials by the caller.

odyssey4me: @hughsaunders What do you mean by 'accept a body closure' ?

pipeline-steps/artifact_build.groovy
```diff
@@ -1,35 +1,81 @@
...
+  withCredentials([
```
hughsaunders: I was thinking more like..
```
def get_rpc_repo_creds(){
  return [
    string(
      credentialsId: "RPC_REPO_IP",
      variable: "REPO_HOST"
    ),
    string(
      credentialsId: "RPC_REPO_SSH_USERNAME_TEXT",
      variable: "REPO_USER"
    ),
    file(
      credentialsId: "RPC_REPO_SSH_USER_PRIVATE_KEY_FILE",
      variable: "REPO_USER_KEY"
    ),
    file(
      credentialsId: "RPC_REPO_SSH_HOST_PUBLIC_KEY_FILE",
      variable: "REPO_HOST_PUBKEY"
    ),
    file(
      credentialsId: "RPC_REPO_GPG_SECRET_KEY_FILE",
      variable: "GPG_PRIVATE"
    ),
    file(
      credentialsId: "RPC_REPO_GPG_PUBLIC_KEY_FILE",
      variable: "GPG_PUBLIC"
    )
  ]
}
```

pipeline-steps/artifact_build.groovy
```diff
@@ -1,35 +1,73 @@
...
+      variable: "GPG_PUBLIC"
+    )
+  ]
+}
+
+def apt(Map args) {
+  common.conditionalStage(
+    stage_name: "Build Apt Artifacts",
+    stage: {
+      withEnv(args.environment_vars, get_rpc_repo_creds()) {
```
hughsaunders: The creds need to be used in withCredentials not withEnv.

pipeline-steps/artifact_build.groovy
evrardjp: @hughsaunders can't we just have everything returned? I'm not sure if I'm the right person to review this... So different than usual!

hughsaunders: @evrardjp I'm not sure what you mean? can you give an example?

odyssey4me: @evrardjp It is now returning all the credentials.

odyssey4me:
@evrardjp Those are intentionally left out as they are not common vars. Once this merges and is confirmed to work, I'll evolve it to include those.

odyssey4me:
@evrardjp FYI I've added the key control vars in https://github.com/rcbops/rpc-gating/pull/117

PR #115, opened 2017-03-31
[WIP] Allow formatted Stages/Steps to be unconditional

odyssey4me:
The conditionalStage/conditionalStep functions appear
to largely be used to consistently format the output
when executing them, rather than to actually make them
conditional.

This is evidenced by the fact that the conditional
stages when executing on the public cloud instances
are absolutely not conditional and instead are required.

This patch renames the two functions, making the default
for each function to be to run the Stage/Step
unconditionally and providing a parameter to change it
for the stage.

TODO:
 - this needs a review to agree on which steps/stages are actually conditional or not
 - once each stage is appropriately aligned, the jobs can be cleaned up to remove the steps/stages from the stages list which aren't conditional

### DIFF
--- a/pipeline-steps/aio_prepare.groovy
+++ b/pipeline-steps/aio_prepare.groovy
@@ -1,6 +1,7 @@
 def prepare(){
-  common.conditionalStage(
+  common.runStage(
     stage_name: "Prepare Deployment",
+    conditional: True,
     stage: {
       prepareRpcGit()
       ansiColor('xterm'){
--- a/pipeline-steps/artifact_build.groovy
+++ b/pipeline-steps/artifact_build.groovy
@@ -1,6 +1,7 @@
 def python(Map args) {
-  common.conditionalStage(
+  common.runStage(
     stage_name: "Build Python Artifacts",
+    conditional: True,
     stage: {
       withEnv(args.environment_vars) {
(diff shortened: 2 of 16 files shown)


odyssey4me:
@hughsaunders Interesting. Under what conditions would you do something like that? Surely that's the purpose of the 'Pause' stage?

hughsaunders:
@odyssey4me yeah pause may well be used in conjunction, eg a job is run including pause, it deploys fine but tempest fails, a new job could be kicked off that skips allocate and connect, but re-runs tempest. After that the first job can be resumed to clear up. There can be issues with state missing as the two jobs will have separate workspaces, but for some things it works.

hughsaunders:
@odyssey4me the main use of pause is for manual investigation; a build fails, rerun with pause then login and poke it. @mattt416 is [adding functionality](https://github.com/rcbops/rpc-gating/pull/110) to skip cleanup based on instance metadata which can be added during the build so that the re-run isn't required.

odyssey4me:
OK, thanks. I'd love to be able to hide these so that they're always included, but don't *have* to be specified in the jobs (they clutter the job definitions somewhat), but I understand the purpose of having them a little better now. Thanks. :)

PR #123, opened 2017-04-04
Cancel builds on updated PRs and pin requirements

mkam:
I had to upgrade jenkins-job-builder to 2.0.0.0b1 so we could use `cancel-builds-on-update` and pinned some other dependencies while I was at it. I also had to include `stevedore<1.20.0` to avoid pbr version conflicts.

If we want to include an entire `pip freeze` for the constraints, I can make that update.

Connects https://github.com/rcbops/u-suk-dev/issues/1465


mattt416:
@hughsaunders I did the upgrade locally and just ensured jenkins still loaded, etc.  :)  I was able to fire off a test job without issue.

requirements.txt
```diff
@@ -0,0 +1,2 @@
+ansible
+jenkins-job-builder
```
mattt416: Could you dump the contents of constraints.txt here and just use this file for `-c` and `-r` to pip?  I vaguely recall something like that being done in openstack-ansible.

mkam:
I added the jenkinsapi package and also added some missing uses of the constraints.txt. Still looking into what's the best practice for constraints files to address whether we need two separate files or not.

PR #124, opened 2017-04-05
Deploy ARA

hughsaunders:
Deploy ARA and submit task results as unit tests to jenkins.

Connects rcbops/u-suk-dev#318


hughsaunders:
This job failed on MaaS and correctly reported that via ARA: https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_master-experimental-ondemand/9/testReport/

Sadly this PR needs a rebase, so I'll do that and then run another test against the rebased version.

pipeline_steps/common.groovy
```diff
@@ -27,10 +29,69 @@ def install_ansible(){
...
+        )
+
+        // remove submitted junit files and clear database
+        sh """#!/bin/bash -x
+          rm -f junit.xml ||:
+          rm -rf ${db_file}
+        """
+      }
+    } else {
+      print("submit_ara skipped as ${env.WORKSPACE}/${db_file} doesn't exist")
```
git-harry: Under what circumstances will this function be called when there is no ara db file?

pipeline_steps/common.groovy
```diff
@@ -27,10 +29,69 @@ def install_ansible(){
...
+  dir(env.WORKSPACE){
+    db_file="ansible.sqlite"
+    if (fileExists(file: db_file)){
+      withEnv(common.get_deploy_script_env()){
+
+        // remove previously generated junit files and generate a new one.
+        sh """#!/bin/bash -x
+          . .venv/bin/activate
+          rm -f junit.xml ||:
+          ara generate junit junit.xml ||:
```
git-harry: If generating the file failed, how is that reported? What is the benefit of ignoring the error? Does the junit function below given some form of useful error?

pipeline_steps/artifact_build.groovy
```diff
@@ -80,13 +80,11 @@ def python() {
         try {
           withCredentials(get_rpc_repo_creds()) {
             common.prepareRpcGit(branch: env.RPC_BRANCH)
-            ansiColor('xterm') {
```
hughsaunders: Good spot, fixed.

pipeline_steps/common.groovy
```diff
@@ -188,7 +188,11 @@ def conditionalStage(Map args){
   stage(args.stage_name){
     if (env.STAGES.contains(args.stage_name)){
         print "Stage Start: ${args.stage_name}"
-        args.stage()
+        ansiColor('xterm') {
```
hughsaunders: While those are indeed duplicate calls, I think they are worth leaving in as those functions may be used outside of a conditional step or stage.

pipeline_steps/common.groovy
hughsaunders: Its not reported explicitly, however the lack of test results displayed in jenkins will alert users to the problem. Also its important these instrumentation functions don't cause build failures as they are not part of the product under test and just there to aid in debugging of legitimate failures. 

The aim of this PR is to reduce the time taken to determine what caused a build to fail, I don't want to introduce more failure cases. If this mechanism fails, and the job itself also fails than debugging can be done as per usual by reading the log.

pipeline_steps/common.groovy
hughsaunders: https://github.com/rcbops/rpc-gating/blob/bug/318/rpc_jobs/rpc_aio.yml#L212 for a Mitaka or below build.

hughsaunders:
Pushed new commits to remove two instances of ansiColor as pointed out by @git-harry

pipeline_steps/common.groovy
```diff
@@ -188,7 +266,11 @@ def conditionalStage(Map args){
...
+          timestamps(){
+            args.stage()
+          } //timestamps
+        } //ansiColor
         print "Stage Complete: ${args.stage_name}"
```
mattt416: Why not put everything inside timestamps, including the logging/debugging messages?

pipeline_steps/ssh_slave.groovy
```diff
@@ -6,61 +6,51 @@
  *  - playbooks/inventory/hosts
  */
 def connect(){
-  common.conditionalStage(
```
mattt416: This change seems a bit out of place here.  If we're going to nuke these stages/steps then we probably need to clean up Connect Slave / Destroy Slave from STAGES.

hughsaunders: This may have been a rebase fail.. I'll check that current head.

hughsaunders: Fixed

pipeline_steps/common.groovy
hughsaunders: I thought it made sense to just time the stage thats being executed. If we want timestamps over everything, the call should be at the top level of every job (as timeout is currently).

hughsaunders:
Pushed new commits to address a rebase fail in ssh_slave.groovy pointed out by @mattt416

rpc_jobs/rpc_aio.yml
```diff
@@ -146,6 +159,9 @@
             RPC_BRANCH using the env vars supplied by ghprb. Ztrigger is a
             jjb variable so uses single braces.
           */
+          if (env.STAGES.contains("Clear Workspace")){{
```
mkam: What use cases are there for not wanting to clear the workspace every time?

mkam:
As a follow up, can we also put this in the multi-node AIO?

mattt416:
I think this is ready to merge but I'd rather avoid doing it at 11 PM incase something blows up.  @git-harry, can you merge it in the morning when @hughsaunders is in?  :)

rpc_jobs/rpc_aio.yml
hughsaunders: The idea was to allow rerunning some stages and re-using state in the workspace. I know that isn't guaranteed to work though as a different workspace may be allocated :/

hughsaunders:
@mkam added rcbops/u-suk-dev#1581 for that.

PR #127, opened 2017-04-06
Update pip after creating a virtualenv

mkam:


pipeline-steps/common.groovy
```diff
@@ -21,6 +21,7 @@ def install_ansible(){
     source .venv/bin/activate
 
     # These pip commands cannot be combined into one.
+    pip install -U pip
     pip install -U six packaging appdirs
     pip install -U setuptools pip
```
hughsaunders: double pip install pip (4x pip?)

mkam: Oh, didn't notice it was already updated later. I'll remove it!

PR #129, opened 2017-04-06
Use COPY instead of ADD

mkam:
I forgot to use COPY instead of ADD. ADD is more likely to cause cache invalidations, so that's why Docker kept having to reinstall the python packages.

https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy

Connects https://github.com/rcbops/u-suk-dev/issues/1539

PR #131, opened 2017-04-07
Add Periodic Cleanup Job

hughsaunders:
This job deletes instances that are older than a specified time (Default
48 hours). This is useful to keep us from hitting quota if an individual
job has a cleanup failure or a user forgets to remove an instance that
was kept for investigation.

Connects rcbops/u-suk-dev#1380

### DIFF
--- a/rpc_jobs/periodic_cleanup.yml
+++ b/rpc_jobs/periodic_cleanup.yml
@@ -1,0 +1,57 @@
+- job:
+    name: 'Periodic-Cleanup'
+    project-type: workflow
+    parameters:
+      - string:
+          name: "INSTANCE_AGE_LIMIT"
+          default: "48"
+          description: |
+            Hours. Instances older than this will be removed.
+      - string:
+          name: "INSTANCE_PREFIX"
+          default: "ra"
+          description: |
+            Only instances whose names match the supplied prefix will be cleaned
+            up.
+      - string:
+          name: "REGION"
+          default: "DFW"
+          description: |
+            Only instances in the specified region will be cleaned up.
+      - rpc_gating_params
+    triggers:
+      - timed: "H * * * *"
+    properties:
+      - build-discarder:
+          days-to-keep: 3
+    dsl: |
+      node(){
+        deleteDir()
+        stage("Prepare"){
+          dir("rpc-gating") {
+              git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
+              common = load 'pipeline_steps/common.groovy'
+              common.docker_cache_workaround()
+              container = docker.build env.BUILD_TAG.toLowerCase()
+          }
+        }
+        container.inside {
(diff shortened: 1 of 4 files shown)


hughsaunders:
I'll run some tests before marking this as ready for review.

git-harry:
A successful job - https://rpc.jenkins.cit.rackspace.net/job/Periodic-Cleanup/32/console

git-harry:
Rebased on top of https://github.com/rcbops/rpc-gating/pull/129

rpc_jobs/periodic_cleanup.yml
```diff
@@ -0,0 +1,57 @@
...
+          description: |
+            Hours. Instances older than this will be removed.
+      - string:
+          name: "INSTANCE_PREFIX"
+          default: "ra"
+          description: |
+            Only instances whose names match the supplied prefix will be cleaned
+            up.
+      - string:
+          name: "OS_REGION_NAME"
```
mattt416: We call this REGION everywhere else.   You could standardise, but I'm not concerned enough to want to block the review.

scripts/periodic_cleanup.py
```diff
@@ -0,0 +1,60 @@
...
+import os
+
+import pyrax
+
+
+def get_env_vars():
+    env_vars = {}
+
+    env_vars["age_limit"] = int(os.environ.get("INSTANCE_AGE_LIMIT", 48))
+    env_vars["instance_prefix"] = os.environ.get("INSTANCE_PREFIX", "ra")
```
mattt416: Doesn't this set env_vars["instance_prefix"] to "ra" if not set, which means the check below never fails?

hughsaunders: it get set to "ra" if its not set, but not if it's set to ""

rpc_jobs/periodic_cleanup.yml
hughsaunders: Good point, I was using OS_REGION_NAME because it matched the other vars used by nova client, but this PR no longer uses nova client, so using REGION to be consistent with the rest of the repo would be better.

rpc_jobs/periodic_cleanup.yml
```diff
@@ -0,0 +1,57 @@
...
+        }
+        container.inside {
+          stage("Checkout"){
+            git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
+          }
+          stage("Public Cloud Cleanup"){
+            withCredentials([
+              string(
+                credentialsId: "dev_pubcloud_username",
+                variable: "OS_USERNAME"
```
git-harry: @mattt416 @hughsaunders along the same lines as the comments on REGION, I assume it makes sense for these to be PUBCLOUD_USERNAME and PUBCLOUD_API_KEY?

PR #132, opened 2017-04-07
Make openstack-ansible-ops parameters re-usable

mattt416:
This commit makes the openstack-ansible-ops params into the params.yml
file so we can re-use these by other jobs.  We also fix some incorrect
indenting in multi_node_aio.yml.

### DIFF
--- a/rpc_jobs/multi_node_aio.yml
+++ b/rpc_jobs/multi_node_aio.yml
@@ -77,23 +77,15 @@
           name: "RPC_BRANCH"
           default: "{branch}"
       - tempest_params:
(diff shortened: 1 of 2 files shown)


mattt416:
Updated OnMetal-Multi-Node-AIO_master-trusty-periodic [here](https://rpc.jenkins.cit.rackspace.net/job/Jenkins-Job-Builder/25/console) and ran a test build [here](https://rpc.jenkins.cit.rackspace.net/job/OnMetal-Multi-Node-AIO_master-trusty-periodic/8/).

PR #133, opened 2017-04-07
Update multi-node parameters

mattt416:
This commit moves more params in multi_node_aio.yml to params.yml, and
uses the rpc_repo_params macro instead of redefining within the job
template.


mattt416:
Job update [here](https://rpc.jenkins.cit.rackspace.net/job/Jenkins-Job-Builder/27/) and job build [here](https://rpc.jenkins.cit.rackspace.net/job/OnMetal-Multi-Node-AIO_master-trusty-periodic/9/).

PR #134, opened 2017-04-07
First pass at a LEM multi-node AIO

mattt416:

- adds MULTI_NODE_AIO_DIR and DATA_DISK_DEVICE params to osa_ops_params
  and overrides these in the respective jobs
- removes old code from setup_jenkins_slave.yml that is no longer
  used and changes a reference from ansible_default_ipv4.address to
  ansible_host (this is necessary because we may try to use a host
  behind a firewall and not wish to use the host's
  ansible_default_ipv4.address address)
- creates rpc_multi_node_aio.yml which is specifically for building the
  multi-node AIO on LEM nodes

Connects https://github.com/rcbops/u-suk-dev/issues/1368

### DIFF
--- a/pipeline_steps/multi_node_aio_prepare.groovy
+++ b/pipeline_steps/multi_node_aio_prepare.groovy
@@ -2,7 +2,7 @@ def prepare() {
   dir("openstack-ansible-ops") {
     git url: env.OSA_OPS_REPO, branch: env.OSA_OPS_BRANCH
   }
-  dir("openstack-ansible-ops/multi-node-aio") {
+  dir("openstack-ansible-ops/${env.MULTI_NODE_AIO_DIR}") {
     common.conditionalStage(
       stage_name: 'Prepare Multi-Node AIO',
       stage: {
@@ -25,7 +25,8 @@ def prepare() {
               "VM_IMAGE_CREATE=true",
               "DEPLOY_OSA=true",
               "PRE_CONFIG_OSA=true",
-              "RUN_OSA=false"]
+              "RUN_OSA=false",
+              "DATA_DISK_DEVICE=${env.DATA_DISK_DEVICE}"]
           ) //run_script
         } //timeout
       } //stage
--- a/rpc_jobs/lem_multi_node_aio.yml
+++ b/rpc_jobs/lem_multi_node_aio.yml
@@ -1,0 +1,135 @@
+- project:
+    name: 'LEM-Multi-Node-AIO-Jobs'
+    # Note: branch is the branch for periodics to build
+    #       branches is the branch pattern to match for PR Jobs.
+    series:
+      - mitaka:
+          branch: mitaka-13.1
+          OPENSTACK_ANSIBLE_BRANCH: "stable/mitaka"
+      - newton:
+          branch: newton-14.0
+          OPENSTACK_ANSIBLE_BRANCH: "stable/newton"
+      - master:
+          branch: master
+          OPENSTACK_ANSIBLE_BRANCH: "stable/newton"
+    context:
+      - trusty:
+          DEFAULT_IMAGE: "14.04.5"
+      - xenial:
+          DEFAULT_IMAGE: "16.04.2"
+
+    # NOTE: Hugh tested this and found that ztrigger overrides series and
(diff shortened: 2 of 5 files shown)


mattt416:
recheck_lint

mattt416:
recheck_lint

mattt416:
New job:

https://rpc.jenkins.cit.rackspace.net/job/RPC-Multi-Node-AIO_master-xenial-periodic/31/console

Still need to validate that this hasn't broken the existing multi-node in any way!

mattt416:
Any compelling reason to move some of the bits and pieces in the JJB out to a separate groovy script?

mkam:
For the JJB vs a groovy script stuff, while it would be nice to reduce duplicate code between the two multi-node jobs, I'd prefer to leave it how it is. I feel like if we abstract it any further, then the job configurations won't have very much useful information left in them in terms of what steps are happening in the job.

mattt416:
@mkam that's a fair comment -- i've renamed the JJB file and jobs to reflect these are for LEM deployments.

rpc_jobs/params.yml
```diff
@@ -54,6 +54,18 @@
           name: PARTITION_HOST
           default: true
           description: Enable partitioning of host data disk device
+      - string:
+          name: DATA_DISK_DEVICE
+          default: "{DATA_DISK_DEVICE}"
```
mattt416: The OnMetal multi-node and LEM hardware multi-node use different data disks (sdb, and a logical volume respectively), which means we cannot set a default in the macro.  Instead, we include the macro in the job definition and then pass DATA_DISK_DEVICE through as a parameter.

rpc_jobs/rpc_multi_node_aio.yml
```diff
@@ -0,0 +1,143 @@
...
+          OPENSTACK_ANSIBLE_BRANCH: "stable/liberty"
+      - mitaka:
+          branch: mitaka-13.1
+          OPENSTACK_ANSIBLE_BRANCH: "stable/mitaka"
+      - newton:
+          branch: newton-14.0
+          OPENSTACK_ANSIBLE_BRANCH: "stable/newton"
+      - master:
+          branch: master
+          branches: "master"
```
mattt416: That's not needed, will remove!

rpc_jobs/rpc_multi_node_aio.yml
```diff
@@ -0,0 +1,143 @@
...
+      - series: liberty
+        context: xenial
+      - series: mitaka
+        context: xenial
+    jobs:
+      - 'RPC-Multi-Node-AIO_{series}-{context}-{ztrigger}'
+
+- job-template:
+    # DEFAULTS
+    branch: master
```
mattt416: Yeah, I believe these aren't needed as I just tested commenting them out and the file still test correctly.  I think we'd only need to set a default if we have a matrix where the variable doesn't necessarily get set in each combination, or if it's referenced in the job template (ie. NUM_TO_KEEP).  I'll get those two vars removed from the template.

pipeline_steps/multi_node_aio_prepare.groovy
```diff
@@ -25,7 +25,8 @@ def prepare() {
               "VM_IMAGE_CREATE=true",
               "DEPLOY_OSA=true",
               "PRE_CONFIG_OSA=true",
-              "RUN_OSA=false"]
+              "RUN_OSA=false",
+              "DATA_DISK_DEVICE=${env.DATA_DISK_DEVICE}"]
```
hughsaunders: Trailing `,` is allowed.

rpc_jobs/lem_multi_node_aio.yml
```diff
@@ -0,0 +1,135 @@
+- project:
+    name: 'LEM-Multi-Node-AIO-Jobs'
+    # Note: branch is the branch for periodics to build
+    #       branches is the branch pattern to match for PR Jobs.
+    series:
+      - mitaka:
+          branch: mitaka-13.1
```
hughsaunders: This should be named RPC_BRANCH but I realise this is following a pattern so have created rcbops/u-suk-dev#1585 to address that.

PR #135, opened 2017-04-10
Push artifacts to Cloud Files

mkam:
- Had to pin requests because of an issue in pyrax where requests==2.11.0 throws an InvalidHeader when it sets the TTL for the file in the header 
- Moved all the cloud account credentials into a common method
- Hardcoded DFW as the upload region since the CIT nodes are in DFW and we're currently only uploading from there

Connects https://github.com/rcbops/u-suk-dev/issues/1438

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -270,7 +270,7 @@ def archive_artifacts(){
     done < <(lxc-ls)
 
     # compress to reduce storage space requirements
-    tar cvjf "\$d".tar.bz2 \$d > artifact_index.txt
+    tar cjf "\$d".tar.bz2 \$d
     """
   } catch (e){
     print(e)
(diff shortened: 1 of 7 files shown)


mkam:
Tested just uploading and publishing: https://rpc.jenkins.cit.rackspace.net/job/mkam-RPC-AIO/10/console
AIO Build: https://rpc.jenkins.cit.rackspace.net/job/mkam-RPC-AIO/11/console
Multi-node AIO Build: https://rpc.jenkins.cit.rackspace.net/job/OnMetal-Multi-Node-AIO/2/console

pipeline_steps/common.groovy
```diff
@@ -278,13 +278,54 @@ def archive_artifacts(){
   } finally{
     // still worth trying to archiveArtifacts even if some part of
     // artifact collection failed.
-    archiveArtifacts(
-      allowEmptyArchive: true,
-      artifacts: "artifacts_${env.BUILD_TAG}.tar.bz2,artifact_index.txt"
-    )
+    stash(
+      name: "artifacts",
+      includes: "artifacts_${env.BUILD_TAG}.tar.bz2")
```
mkam: The venvPlaybook methods that we use to execute ansible has an assumption that it's running from the CIT node, which is a different OS than the created slave nodes. But it's probably better that I make that method more general rather than having to pass the file between nodes, so I'll make that update.

pipeline_steps/common.groovy
```diff
@@ -278,13 +278,54 @@ def archive_artifacts(){
...
+    node(){
+      sh "rm -f index.html"
```
mkam: My thought process for this was that the playbook would do just uploading and getting the file to the proper location would be done outside of that in case we want to reuse the playbook in a different way. I'll update the playbook to make the destination a parameter, though.

pipeline_steps/common.groovy
```diff
@@ -278,13 +278,54 @@ def archive_artifacts(){
   } finally{
     // still worth trying to archiveArtifacts even if some part of
     // artifact collection failed.
-    archiveArtifacts(
-      allowEmptyArchive: true,
-      artifacts: "artifacts_${env.BUILD_TAG}.tar.bz2,artifact_index.txt"
```
mkam: The index is so that we can see what files are in the tar without having to download it. With the old archive artifacts method you could view artifact_index.txt without having to download it, which isn't the case anymore, so I didn't include it. I could try putting the contents of artifacts_index.txt into index.html so we keep that functionality.

mkam: Added an optional "description_file" variable to the playbook and passed it artifacts_index.txt.

pipeline_steps/common.groovy
mkam: It turns out that there was only one fix needed to get the rax module to work on the cloud  slave, and that was to install libssl-dev. It's been a while since I last tried, so I thought it was going to be more work than that. Also, since we're uploading directly from the cloud node now, there's less of a need to remove the old artifacts since most nodes are for single use, so I omitted that part in my update.

pipeline_steps/common.groovy
hughsaunders: #124 removes artifact_index because it isn't useful. Its too large to be a good summary, if you're going to download that, you may as well download the full archive.

mkam:
Updated it and retested here:
https://rpc.jenkins.cit.rackspace.net/job/mkam-RPC-AIO/61/

git-harry:
This seems to work as expected in terms of uploading the data and creating the link. What happens when the job is deleted? How are artefacts removed or will they be kept indefinitely?

mkam:
The artifacts themselves will expire automatically after 30 days. Job or build deletion will only lose the links.

https://github.com/rcbops/rpc-gating/pull/135/files#diff-621286eb1d8b3532c2094c4d0fd6743dR17

PR #136, opened 2017-04-11
Create separate slave step/stage

mattt416:
This commit moves the step / stage for connecting and destroying a
jenkins slave from pubcloud.groovy to ssh_slave.groovy so that these
steps can be properly included in a non pubcloud context.

Connects https://github.com/rcbops/u-suk-dev/issues/1368


mattt416:
https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_master-ceph-periodic/45/

PR #138, opened 2017-04-11
Run flake8 as part of lint script

mkam:
Connected https://github.com/rcbops/u-suk-dev/issues/1537

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,9 @@ RUN apt-get update && apt-get install -y groovy2 python-pip build-essential pyth
 RUN apt-get install -y libffi-dev
 RUN apt-get install -y sudo
 COPY requirements.txt /requirements.txt
+COPY test-requirements.txt /test-requirements.txt
 COPY constraints.txt /constraints.txt
 RUN pip install -c /constraints.txt -r /requirements.txt
+RUN pip install -c /constraints.txt -r /test-requirements.txt
 RUN useradd jenkins --shell /bin/bash --create-home --uid 500
 RUN echo "jenkins ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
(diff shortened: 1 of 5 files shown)


mkam:
recheck_lint

lint.sh
```diff
@@ -73,7 +72,7 @@ check_bash(){
 check_python(){
   while read script
   do
-    python -m py_compile  $script \
+    python -m py_compile  $script  && flake8 $script \
```
hughsaunders: I don't think both are needed as flake8 will catch compilation errors.

Dockerfile
```diff
@@ -3,7 +3,9 @@ RUN apt-get update && apt-get install -y groovy2 python-pip build-essential pyth
 RUN apt-get install -y libffi-dev
 RUN apt-get install -y sudo
 COPY requirements.txt /requirements.txt
+COPY constraints.txt /test-requirements.txt
```
hughsaunders: I think you meant test-requirements instead of constraints?

git-harry:
@mkam I've added the line
```
Connected https://github.com/rcbops/u-suk-dev/issues/1537
```
to the description so that the pull request and issue are linked in Waffle.

If you add this line to the commit message, the description get populated with it automatically linking the two. In addition, this means the repo history has a record of the issue which can be useful for troubleshooting etc.

mkam:
recheck_lint

lint.sh
```diff
@@ -73,7 +72,7 @@ check_bash(){
 check_python(){
   while read script
   do
-    python -m py_compile  $script \
+    flake8 $script \
       && echo "Python syntax ok: $script" \
       || { echo "Python syntax fail $script"; rc=1; }
   done < <(find ${fargs[@]} -iname \*.py)
```
mattt416: Any downside to just running flake8 on the entire dir, rather than on each found py script?

mattt416:
Removing @hughsaunders' review since he's away for some time, plus his concerns have already been addressed.

PR #139, opened 2017-04-13
Reduce Timeout Age

hughsaunders:
Some jobs are still failing on quota, so we need to reduce the amount
of time that instances can idle before they are cleaned up.


b3rnard0:
@hughsaunders Does this have a connected parent issue?

PR #142, opened 2017-04-18
Run all tempest tests for multi-node AIO

mkam:
Successful run of all tests against newton-14.0: https://rpc.jenkins.cit.rackspace.net/job/OnMetal-Multi-Node-AIO/22/
[In progress] Master: https://rpc.jenkins.cit.rackspace.net/job/OnMetal-Multi-Node-AIO/23/
[In progress] Mitaka: https://rpc.jenkins.cit.rackspace.net/job/OnMetal-Multi-Node-AIO/25/

Connects https://github.com/rcbops/u-suk-dev/issues/1490

### DIFF
--- a/playbooks/vars/all.yml
+++ b/playbooks/vars/all.yml
@@ -4,6 +4,7 @@ gating_overrides_file: "/etc/openstack_deploy/user_zzz_gating_variables.yml"
 default_gating_overrides:
   maas_notification_plan: npTechnicalContactsEmail
   nova_virt_type: qemu
+  nova_cross_az_attach: True
   rackspace_cloud_auth_url: "https://identity.api.rackspacecloud.com/v2.0/"
   rackspace_cloud_tenant_id: "{{lookup('env', 'PUBCLOUD_TENANT_ID')}}"
   rackspace_cloud_username: "{{lookup('env', 'PUBCLOUD_USERNAME')}}"
@@ -21,3 +22,4 @@ default_gating_overrides:
       reseller_admin_role: ResellerAdmin
     compute-feature-enabled:
       personality: false
+      attach_encrypted_volume: false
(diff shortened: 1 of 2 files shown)


mattt416:
I've created a [separate commit](https://github.com/rcbops/rpc-gating/commit/462b0b9161066aaabe7f304d3250452d5e3c6d11) to enable cinder-backup across the board as these are currently skipped on the multi node and when we deploy the AIO w/ swift.  I'll give this some testing and will create the PR once I know it works correctly.

mkam:
The master build all passed, and the mitaka build had one known failure and one flaky test failure. Reran the scenario test failure and it passed on the second run:
https://rpc.jenkins.cit.rackspace.net/view/Multi-Node%20AIO/job/OnMetal-Multi-Node-AIO/26/testReport/

playbooks/vars/all.yml
```diff
@@ -4,6 +4,7 @@ gating_overrides_file: "/etc/openstack_deploy/user_zzz_gating_variables.yml"
 default_gating_overrides:
   maas_notification_plan: npTechnicalContactsEmail
   nova_virt_type: qemu
+  nova_cross_az_attach: True
```
mattt416: @mkam i was having a quick look at this before approving, and noticed that rpco explicitly disables `nova_cross_az_attach`.  I'm not entirely sure why that is but I think we best keep that behaviour so we're more closely testing what is getting deployed.  If tests fail because of this then we'll have to remove those tests from being run.  :(

mattt416: Looks like there was an attempt to [remove](https://github.com/rcbops/rpc-openstack/pull/2098) this override, however support wanted it left as is.  As such, I think we should remove this override from gating and exclude any failing tests from running.  Perhaps running `all` tests is not going to be possible?

PR #147, opened 2017-04-20
Skip ARA if build succeeded to prevent false negatives

mkam:
Might not be the ideal solution for the long term, but it should at least prevent the gates from blocking PRs that had no problems.

We also adjust the path to the .venv when running jenkins_node.py, which ensures the ssh slave is torn down correctly.

Connects https://github.com/rcbops/u-suk-dev/issues/1587

### DIFF
--- a/rpc_jobs/rpc_aio.yml
+++ b/rpc_jobs/rpc_aio.yml
@@ -207,9 +207,9 @@
               }}
             }} catch (e) {{
               print(e)
+              common.submit_ara()
               throw e
             }} finally {{
-              common.submit_ara()
               common.archive_artifacts()
             }}
           }} //pubcloud slave
--- a/pipeline_steps/ssh_slave.groovy
+++ b/pipeline_steps/ssh_slave.groovy
@@ -50,7 +50,7 @@ def destroy(){
       ]){
         dir("rpc-gating/scripts"){
           sh """
-            . ../playbooks/.venv/bin/activate
+            . ${env.WORKSPACE}/.venv/bin/activate
             pip install 'pip==9.0.1'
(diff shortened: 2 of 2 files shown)


mkam:
recheck_lint

mkam:
Tested the logic with a test job, fully testing here: https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_master-swift-periodic/58/

mattt416:
I went to check on the build above and noticed that after 3+ hours it was still deploying nova and the build dir was massive.  I made another small change to this PR, updated the JJB for the job, and submit https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_master-swift-periodic/59/.  I'll need to check on this in the morning to see how it does.

rpc_jobs/rpc_aio.yml
```diff
@@ -205,11 +205,14 @@
...
                 holland.holland()
               }}
+              currentBuild.result = "SUCCESS"
             }} catch (e) {{
               print(e)
               throw e
             }} finally {{
-              common.submit_ara()
+              if(currentBuild.result != "SUCCESS"){{
+                common.submit_ara()
```
git-harry: This line could have been moved into the `catch` section which would save having to set the result etc.

hughsaunders: 👍

hughsaunders:
Sadly this is the best way to solve the unstable problem for now. 👍 With the modification suggested by @git-harry

mkam:
Closing in favor of https://github.com/rcbops/rpc-gating/pull/148

PR #148, opened 2017-04-21
Revert ARA

mattt416:
This PR reverts the following two PRs:

https://github.com/rcbops/rpc-gating/pull/146
https://github.com/rcbops/rpc-gating/pull/124

We have started to see some jenkins instability which we suspect is related to #124.  Ideally we would like to troubleshoot further however rpco users are currently blocked so reverting and testing #124 further before rolling out may be the best way forward.

Connects https://github.com/rcbops/u-suk-dev/issues/1589


mattt416:
https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_master-swift-periodic/61/ to test this revert

evrardjp:
Surgical. Fine for me.

evrardjp:
Maybe this PR should be squash merged instead of just merged to avoid having multiple revert in tree?

git-harry:
This looks to be reverting the intended commits. Is the intention to wait for the test job first before merging? If all the jobs are currently broken merging this can hardly make things worse.

PR #151, opened 2017-04-25
Checkout PR repository correctly

odyssey4me:
When a job is triggered by a PR, the checkout must
universally include the change introduced by the PR.

Currently only the AIO job inherits the correct PR
checkout. This patch moves the PR resolving to the
general RPC-O repo checkout so that it becomes
universal.

Connects https://github.com/rcbops/u-suk-dev/issues/1607

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -352,18 +352,34 @@ def prepareConfigs(Map args){
   } //withCredentials
 }
 
-def prepareRpcGit(Map args){
+def prepareRpcGit(branch = "auto"){
   dir("/opt/rpc-openstack"){
+
+    if (branch == "auto"){
+      /* if job is triggered by PR, then we need to set RPC_REPO and
+         RPC_BRANCH using the env vars supplied by ghprb.
+      */
+      if ( env.ghprbPullId != null ){
+        env.RPC_REPO = "https://github.com/${env.ghprbGhRepository}.git"
+        branch = "origin/pr/${env.ghprbPullId}/merge"
+        print("Triggered by PR: ${env.ghprbPullLink}")
+      } else {
+        branch = env.RPC_BRANCH
+      }
+    }
+
+    print("Repo: ${env.RPC_REPO} Branch: ${branch}")
+
     // checkout used instead of git as a custom refspec is required
     // to checkout pull requests
     checkout([$class: 'GitSCM',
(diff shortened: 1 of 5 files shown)


pipeline_steps/common.groovy
```diff
@@ -354,6 +354,16 @@ def prepareConfigs(Map args){
 
 def prepareRpcGit(Map args){
   dir("/opt/rpc-openstack"){
+
+    /* if job is triggered by PR, then we need to set RPC_REPO and
+       RPC_BRANCH using the env vars supplied by ghprb.
+    */
+    if ( ghprbPullId != null  ){{
```
hughsaunders: Double braces are used in JJB jobs to escape the jjb template mechanism. In direct groovy code single braces should be used.

pipeline_steps/common.groovy
```diff
@@ -354,6 +354,16 @@ def prepareConfigs(Map args){
 
 def prepareRpcGit(Map args){
   dir("/opt/rpc-openstack"){
+
+    /* if job is triggered by PR, then we need to set RPC_REPO and
+       RPC_BRANCH using the env vars supplied by ghprb.
+    */
+    if ( env.ghprbPullId != null ){
+      env.RPC_REPO = "https://github.com/${env.ghprbGhRepository}.git"
+      env.RPC_BRANCH = "origin/pr/${env.ghprbPullId}/merge"
```
hughsaunders: This isn't used in the checkout below, args.branch is. Also we still need to be able to override the branch for checking out the upgrade_from version. 

How about:
```
def prepareRpcGit(branch="auto"){
  if (branch == "auto"){
    if (env.ghprbPullId != null){
      env.RPC_REPO = "https://github.com/${env.ghprbGhRepository}.git"
      branch = "origin/pr/${env.ghprbPullId}/merge"
    }
    branch = env.RPC_BRANCH
  }

 ...

```
Note that `args.branch` will need to change to `branch` and `branch: env.RPC_BRANCH` will need to be removed from all the calls to `prepareRpcGit`. The upgrade call will be `common.prepareRpcGit(env.UPGRADE_FROM_REF)`

hughsaunders: gah `branch = env.RPC_BRANCH` will need to be above the inner if, otherwise it will override the branch from the pull request.

PR #149, opened 2017-04-24
Add support to deploy and test from deploy VM node

mkam:
Adds the VM parameter to multiple method so that we can run tasks on the newly added deploy node of the multi-node AIO environment.

I used named parameters instead of `Map args` with this card in mind: https://github.com/rcbops/u-suk-dev/issues/1531

Connects https://github.com/rcbops/u-suk-dev/issues/1578

PR #152, opened 2017-04-25
Add LEM stage and cleanup stages

mattt416:

1. Adds a new stage for creating the logical volume for VMs, which
   effectively allows you to re-move the stage from a job if the host
   already has the logical volume created (running again will generate
   a failure)
2. Cleans up STAGES and loads the kibana tests

Connects https://github.com/rcbops/u-suk-dev/issues/1369

### DIFF
--- a/rpc_jobs/lem_multi_node_aio.yml
+++ b/rpc_jobs/lem_multi_node_aio.yml
@@ -59,19 +59,22 @@
           TESTR_OPTS: ""
       - string:
           name: STAGES
-          default: "Connect Slave, Prepare Multi-Node AIO, Prepare RPC Configs, Deploy RPC w/ Script, Destroy Slave"
+          default: "Connect Slave, Prepare LEM AIO, Prepare Multi-Node AIO, Prepare RPC Configs, Deploy RPC w/ Script, Install Tempest, Tempest Tests, Prepare Kibana Selenium, Kibana Tests, Holland, Destroy Slave"
           description: |
             Pipeline stages to run CSV. Note that this list does not influence execution order.
             Options:
               Connect Slave
-              Prepare Deployment
+              Prepare LEM AIO
+              Prepare Multi-Node AIO
+              Prepare RPC Configs
               Deploy RPC w/ Script
               Setup MaaS
               Verify MaaS
               Install Tempest
               Tempest Tests
+              Prepare Kibana Selenium
+              Kibana Tests
               Holland (test holland mysql backup)
-              Upgrade
               Destroy Slave
       - string:
           name: INSTANCE_NAME
@@ -99,6 +102,7 @@
               tempest = load 'pipeline_steps/tempest.groovy'
               holland = load 'pipeline_steps/holland.groovy'
               maas = load 'pipeline_steps/maas.groovy'
+              kibana = load 'pipeline_steps/kibana.groovy'
           }}
           dir("rpc-gating/playbooks") {{
             writeFile file: "inventory/hosts", text: "[job_nodes]\n${{INSTANCE_NAME}} ansible_host=${{INSTANCE_IP}}\n"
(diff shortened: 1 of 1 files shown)


rpc_jobs/lem_multi_node_aio.yml
```diff
@@ -99,16 +102,19 @@
...
           dir("rpc-gating/playbooks") {{
             writeFile file: "inventory/hosts", text: "[job_nodes]\n${{INSTANCE_NAME}} ansible_host=${{INSTANCE_IP}}\n"
             common.install_ansible()
           }}
           ssh_slave.connect()
           node(INSTANCE_NAME) {{
-            sh """
-               lvcreate -n aio -L250G lxc
-            """
+            if (env.STAGES.contains("Prepare LEM AIO")) {{
```
git-harry: The other uses of this seem to be when a stage has been created with `stage()`. Is this a "real" stage? I assume that means it won't appear in the table for the job as a stage. It seems strange that `STAGES` can include a stage that effectively doesn't exist. I'm not really expressing an opinion here, just trying to understand things better.

rpc_jobs/lem_multi_node_aio.yml
```diff
@@ -59,19 +59,22 @@
           TESTR_OPTS: ""
       - string:
           name: STAGES
-          default: "Connect Slave, Prepare Multi-Node AIO, Prepare RPC Configs, Deploy RPC w/ Script, Destroy Slave"
+          default: "Connect Slave, Prepare LEM AIO, Prepare Multi-Node AIO, Prepare RPC Configs, Deploy RPC w/ Script, Setup MaaS, Verify MaaS, Install Temepst, Tempest Tests, Prepare Kibana Selenium, Kibana Tests, Holland, Destroy Slave"
```
git-harry: I thought `, Setup MaaS, Verify MaaS,` was being left to https://github.com/rcbops/u-suk-dev/issues/1370. I would have thought this wouldn't work because the prepare stage is missing.

mattt416: You are right, I will remove so these can be added in https://github.com/rcbops/u-suk-dev/issues/1370.

rpc_jobs/lem_multi_node_aio.yml
mattt416: Actually this is a good question and has me now thinking whether what I'm doing is right.  Give me a sec to think about this a bit more!

Side note, `Prepare Multi-Node AIO` does end up appearing as a stage on https://rpc.jenkins.cit.rackspace.net/job/LEM-Multi-Node-AIO_master-xenial-periodic/.

git-harry: Yes but `Prepare Multi-Node AIO` is a stage https://github.com/rcbops/rpc-gating/blob/master/pipeline_steps/multi_node_aio_prepare.groovy#L7

mattt416: Heh, ignore me ... confusing that with `Prepare LEM AIO`.

mattt416:
@git-harry tested this `Prepare LEM AIO` stage here:

https://rpc.jenkins.cit.rackspace.net/job/LEM-Multi-Node-AIO_master-xenial-periodic/23/console

rpc_jobs/lem_multi_node_aio.yml
```diff
@@ -99,16 +102,22 @@
...
           dir("rpc-gating/playbooks") {{
             writeFile file: "inventory/hosts", text: "[job_nodes]\n${{INSTANCE_NAME}} ansible_host=${{INSTANCE_IP}}\n"
             common.install_ansible()
           }}
           ssh_slave.connect()
           node(INSTANCE_NAME) {{
-            sh """
-               lvcreate -n aio -L250G lxc
-            """
+            common.conditionalStage(
```
git-harry: Did you consider adding this in a pipeline file? While this probably won't be used anywhere else it's likely nicer to read.

mattt416: I did, but I think the job template will need a fair bit of refactoring once https://github.com/rcbops/rpc-gating/pull/149 lands so I figured we could extract some of the function calls to separate groovy scripts once we perform those updates.

mattt416:
@git-harry squashed commits!

PR #154, opened 2017-04-26
Configure apt artifact source when necessary

odyssey4me:
When a more restricted setup is used to ensure
that an artifacted deployment only uses the
prepared artifacts, the jobs fail as the jenkins
slave cannot install the JRE which comes from
the ubuntu updates repo.

This patch ensures that the slave is prepared
with the apt artifact repo source prior to
trying to install anything.

For the ``artifacts-14.0`` branch, this should
setup the artifacted apt repo and key. Without
doing so the job will fail to install the apt
packages. For any other branches (currently)
it should skip those tasks.

Connects https://github.com/rcbops/u-suk-dev/issues/1627

### DIFF
--- a/playbooks/setup_jenkins_slave.yml
+++ b/playbooks/setup_jenkins_slave.yml
@@ -1,7 +1,74 @@
 ---
 - hosts: job_nodes
   user: root
+  vars:
+    rpc_release_file_name: "{{ lookup('env', 'WORKSPACE') ~ '/rpc-openstack/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml' }}"
+    rpco_mirror_base_url: "http://rpc-repo.rackspace.com"
+    rpco_mirror_apt_filename: "rpco"
+    rpco_gpg_key_location: "{{ rpco_mirror_base_url }}/apt-mirror/"
+    rpco_gpg_key_name: "rcbops-release-signing-key.asc"
+    rpco_gpg_key_id: 52AA252F
   tasks:
+
+    # Check whether a file exists to help determine the
+    # rpc_release, as the job may not include a checkout
+    # of rpc-openstack.
+    - name: Check if release file name exists
+      stat:
+        path: "{{ rpc_release_file_name }}"
+      register: _release_file
+      delegate_to: localhost
+
+    # We then want to check the value of the rpc_release
+    # variable in the file. It may be a bit of jinja, or
+    # not exist, or be a real value. These edge cases are
+    # needed to protect against due to the many releases
+    # tested by the gate and the varied approaches across
+    # them all.
+    # If the value is of the form r<major>.<minor>.<patch>
+    # and optionally including a release candidate suffix
+    # then we have a valid version.
+    - name: Determine rpc_release value
+      shell: |
+        RPC_RELEASE=$(awk '/^rpc_release/ { print $2; }' {{ rpc_release_file_name }} | sed 's/"//g')
+        if [[ ! ${RPC_RELEASE} =~ ^r[0-9]+\.[0-9]+\.[0-9]+(rc[0-9]+)?$ ]]; then
+          RPC_RELEASE="none"
+        fi
+        echo ${RPC_RELEASE}
+      args:
+        executable: "/bin/bash"
+      register: _release_value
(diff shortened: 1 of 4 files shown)


playbooks/setup_jenkins_slave.yml
```diff
@@ -1,7 +1,49 @@
...
+    # needed to protect against due to the many releases
+    # tested by the gate and the varied approaches across
+    # them all.
+    # If the value is of the form r<major>.<minor>.<patch>
+    # and optionally including a release candidate suffix
+    # then we have a valid version.
+    - name: Determine rpc_release value
+      set_fact:
+        rpc_release_value: |-
+          lookup('pipe', 'awk "/^rpc_release/ { print $2; }" ' ~ rpc_release_file_name) | replace('"', '')
```
odyssey4me: There is probably a better way to grab this value. I'm open to suggestions!

mattt416:
```
root@ramsp-206-9c3e:/opt/rpc-openstack/rpcd# cat etc/openstack_deploy/user_rpco_variables_defaults.yml | awk "/^rpc_release/ { print $2; }"
rpc_release: "{{ lookup('pipe', 'cd /opt/rpc-openstack && git describe --tags --abbrev=0') }}"
root@ramsp-206-9c3e:/opt/rpc-openstack/rpcd#
```

Is that correct?

playbooks/setup_jenkins_slave.yml
```diff
@@ -1,7 +1,49 @@
 ---
 - hosts: job_nodes
   user: root
+  vars:
+    rpc_release_file_name: "{{ lookup('env', 'WORKSPACE') ~ 'rpc-openstack/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml' }}"
   tasks:
+
+    # We first check whether the file exists, as the job may
+    # not include a checkout of rpc-openstack
+    - name: Check if release file name exists
```
mkam: Most if not all jobs will run this setup before they would checkout rpc-openstack since the CIT Jenkins slave generally only needs to pull rpc-gating. Additionally, the checkouts would be on the CIT slave instead of the the public cloud server that is currently being configured, so these tasks should be delegated to localhost instead of running on the job_nodes hosts.

playbooks/setup_jenkins_slave.yml
odyssey4me: Yes - if you look in the artifacts-14.0 branch you'll see a static value, which is what we're changing the whole release process to use. We don't use the value unless it matches a version string - see the conditional in the apt_repo task.

mattt416: I see, but will that work when run against rpc-o master for example?

odyssey4me: Odd that it gave you the entire line though. You should have gotten ``"{{"``

playbooks/setup_jenkins_slave.yml
```diff
@@ -1,7 +1,74 @@
...
+        - _release_file.stat.exists
+
+    # If the release file exists, and we have a valid value
+    # then add an apt key & source for the artifacts repo for
+    # that version.
+    # This ensures that builds using the restricted starting
+    # apt configuration for an artifacted build are able to
+    # install the JRE package below which is only available
+    # in the ubuntu-updates repo.
+    - name: Configure the apt artifact repo key
```
hughsaunders: The ansible version on the CIT slave is always >= 2, so could this and the following task be in a block to save the repeated when?

odyssey4me: Sure it could, although I find using conditionals for blocks makes it harder to read and provides no other benefit. Blocks are great for try/rescue but not so great for doing conditional blocks.

odyssey4me:
Tested on the rpc-o master branch to validate that it does not break the non-artifacted AIO build: https://rpc.jenkins.cit.rackspace.net/job/odyssey4me-RPC-AIO/4/console

Tested on the rpc-o artifacts branch to validate that it does what it's supposed to do for an artifacted build: https://rpc.jenkins.cit.rackspace.net/job/odyssey4me-RPC-Artifact-Build/5/console

PR #155, opened 2017-04-26
Use MaaS token and API URL

git-harry:
The MaaS playbooks can either use a set of public cloud credentials or a
token and API URL. Customer deployments always require a token and API
URL. This commit adjusts how MaaS is deployed so that the existing
public cloud credentials are used to generate a token and API URL which
are then used by the playbooks instead.

This changes applies to all jobs deploying MaaS. Given that the use of
public cloud credentials in the playbooks is only for the benefit of
developers it makes sense for that to be the untested path.

This change does not directly address multi-physical-node deployments
using LEM however that can build on this work when it becomes a reality.

Connected https://github.com/rcbops/u-suk-dev/issues/1370

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -340,14 +340,20 @@ def prepareConfigs(Map args){
       } //dir
       dir("rpc-gating/playbooks"){
         common.install_ansible()
-        common.venvPlaybook(
-          playbooks: ["aio_config.yml"],
-          venv: ".venv",
-          args: [
-            "-i inventory",
-            "--extra-vars \"@vars/${args.deployment_type}.yml\""
-          ]
-        )
+        withCredentials(common.get_cloud_creds()) {
+          String venv = "${env.WORKSPACE}/rpc-gating/playbooks/.venv"
+          List maas_vars = maas.get_maas_token_and_url(env.PUBCLOUD_USERNAME, env.PUBCLOUD_API_KEY, env.REGION, venv)
+          withEnv(maas_vars) {
+            common.venvPlaybook(
+              playbooks: ["aio_config.yml"],
+              venv: ".venv",
+              args: [
+                "-i inventory",
+                "--extra-vars \"@vars/${args.deployment_type}.yml\""
+              ]
+            )
+          }
+        }
       } //dir
   } //withCredentials
 }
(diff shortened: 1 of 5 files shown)


mattt416:
Hi @git-harry , going to look at this now -- but looks like it needs a rebase and also are there any tests you ran that you can link here for us to see?  That'll save us having to spin something up to test.

pipeline_steps/common.groovy
```diff
@@ -340,14 +340,20 @@ def prepareConfigs(Map args){
...
-            "--extra-vars \"@vars/${args.deployment_type}.yml\""
-          ]
-        )
+        withCredentials(common.get_cloud_creds()) {
+          String venv = "${env.WORKSPACE}/rpc-gating/playbooks/.venv"
+          List maas_vars = maas.get_maas_token_and_url(env.PUBCLOUD_USERNAME, env.PUBCLOUD_API_KEY, env.REGION, venv)
+          withEnv(maas_vars) {
+            common.venvPlaybook(
+              playbooks: ["aio_config.yml"],
+              venv: ".venv",
```
mattt416: This is the default, it can be omitted.

git-harry:
@mattt416 I've got a lem one that hasn't managed to complete yet and I ran some rpc_aio's that were successful:
- [lem] https://rpc.jenkins.cit.rackspace.net/job/git-harry2_master-xenial/10/
- [rpc_aio] https://rpc.jenkins.cit.rackspace.net/job/git-harry_master-xenial/21/

mattt416:
OK cool, I'm testing a LEM one here too:

https://rpc.jenkins.cit.rackspace.net/job/LEM-Multi-Node-AIO_master-xenial-periodic/27/console

mattt416:
My previous test failed, and a [subsequent](https://rpc.jenkins.cit.rackspace.net/job/LEM-Multi-Node-AIO_master-xenial-periodic/32/console) one failed also.  I actually think this is a result of the LEM nodes and not this commit itself, so I'll need to re-test on an OnMetal AIO.

mattt416:
Tested an OnMetal here:

https://rpc.jenkins.cit.rackspace.net/job/OnMetal-Multi-Node-AIO_master-trusty-periodic/29/console

PR #158, opened 2017-04-27
added jobs for generic irr testing

cloudnull:

### DIFF
--- a/pipeline_steps/irr_role_tests.groovy
+++ b/pipeline_steps/irr_role_tests.groovy
@@ -1,0 +1,72 @@
+def run_irr_tests() {
+  common.conditionalStage(
+    stage_name: "Run IRR Tests",
+    stage: {
+      pubcloud.runonpubcloud {
+        try {
+          ansiColor('xterm') {
+            dir("${env.WORKSPACE}/${env.ghprbGhRepository}") {
+              print("Triggered by PR: ${env.ghprbPullLink}")
+              checkout([$class: 'GitSCM',
+                branches: [[name: "origin/pr/${env.ghprbPullId}/merge"]],
+                doGenerateSubmoduleConfigurations: false,
+                extensions: [[$class: 'CleanCheckout']],
+                submoduleCfg: [],
+                userRemoteConfigs: [
+                  [
+                    url: "https://github.com/${env.ghprbGhRepository}.git",
+                    refspec: '+refs/pull/*:refs/remotes/origin/pr/* +refs/heads/*:refs/remotes/origin/*'
+                  ]
+                ]
+              ]) // checkout
+              sh """#!/bin/bash
+              bash ./run_tests.sh
+              """
+            } // dir
+          } // ansiColor
+        } catch (e) {
+          print(e)
+          throw e
+        } finally {
+          irr_archive_artifacts()
+        }
+      } // pubcloud slave
+    } // stage
+  ) // conditionalStage
+}
+
+def irr_archive_artifacts(){
+  print("Running Log Collection")
+  try{
+    sh """#!/bin/bash
+    d="artifacts_\${BUILD_TAG}"
+    mkdir -p "${WORKSPACE}/logs"
+    pushd "${WORKSPACE}/logs"
+      touch "\$d".marker
+    popd
+    tar -C "${WORKSPACE}/logs" -cjf "${env.WORKSPACE}/\$d".tar.bz2 .
+    """
+  } catch (e){
+    print(e)
+    throw(e)
+  } finally{
+    // still worth trying to archiveArtifacts even if some part of
+    // artifact collection failed.
+      pubcloud.uploadToCloudFiles(
+        container: "jenkins_logs",
+        src: "${env.WORKSPACE}/artifacts_${env.BUILD_TAG}.tar.bz2",
+        html_report_dest: "${env.WORKSPACE}/artifacts_report/index.html")
+      publishHTML(
+        allowMissing: true,
+        alwaysLinkToLastBuild: true,
+        keepAll: true,
+        reportDir: 'artifacts_report',
+        reportFiles: 'index.html',
+        reportName: 'Build Artifact Links'
+      )
+    sh """
+    rm -rf "${WORKSPACE}/logs"
+    rm -f artifacts_${env.BUILD_TAG}.tar.bz2
+    """
+  }
+}
--- a/rpc_jobs/irr_role_test.yml
+++ b/rpc_jobs/irr_role_test.yml
@@ -1,0 +1,92 @@
+- project:
+    name: 'RPC-IRR-Jobs'
+    repo:
+      # IRR REPOS ARE ADDED HERE
+      - rpc_maas:
+          irr_repo_url: https://github.com/rcbops/rpc-maas
+    series:
+      # Note: branches is the branch pattern to match for PR Jobs.
+      - master:
+          branches: "master"
+    image:
+      - xenial:
+          IMAGE: "Ubuntu 16.04 LTS (Xenial Xerus) (PVHVM)"
+      - trusty:
+          IMAGE: "Ubuntu 14.04 LTS (Trusty Tahr) (PVHVM)"
+    context:
+      - base
+    jobs:
+      - 'RPC-IRR_{repo}-{series}-{image}-{context}'
+
+- job-template:
+    # DEFAULTS
+    STAGES: >
+      Allocate Resources,
+      Connect Slave,
+      Prepare Deployment,
+      Run IRR Tests,
+      Cleanup,
+      Destroy Slave
+    NUM_TO_KEEP: 30
+    # TEMPLATE
+    name: 'RPC-IRR_{repo}-{series}-{image}-{context}'
+    project-type: workflow
+    concurrent: true
+    properties:
+      - build-discarder:
+          num-to-keep: "{NUM_TO_KEEP}"
+      - github:
+          url: "{irr_repo_url}"
+    parameters:
+      - rpc_gating_params
+      - single_use_slave_params:
+          IMAGE: "{IMAGE}"
+          FLAVOR: "{FLAVOR}"
+          REGION: "{REGION}"
+      - string:
+          name: IRR_CONTEXT
+          default: "{context}"
+      - string:
+          name: IRR_IMAGE
+          default: "{image}"
+      - string:
+          name: IRR_SERIES
+          default: "{series}"
+      - string:
+          name: STAGES
+          default: "{STAGES}"
+          description: |
+            Pipeline stages to run CSV. Note that this list does not influence execution order.
+            Options:
+              Allocate Resources
+              Connect Slave
+              Prepare Deployment
+              Run IRR Tests
+              Cleanup
+              Destroy Slave
+    triggers:
+      - github-pull-request:
+          org-list:
+            - rcbops
+          github-hooks: true
+          trigger-phrase: '.*recheck_cit_all.*|.*recheck_cit_{repo}_{series}_{image}_{context}.*'
+          only-trigger-phrase: false
+          white-list-target-branches:
+            - "{branches}"
+          auth-id: "github_account_rpc_jenkins_svc"
+          status-context: 'CIT/{image}_{context}'
+          cancel-builds-on-update: true
+
+    dsl: |
+      // CIT Slave node
+      timeout(time: 2, unit: 'HOURS'){{
+        node() {{
+          dir("rpc-gating") {{
+              git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
+              common = load 'pipeline_steps/common.groovy'
+              pubcloud = load 'pipeline_steps/pubcloud.groovy'
+              irr_role_tests = load 'pipeline_steps/irr_role_tests.groovy'
+          }}
+          irr_role_tests.run_irr_tests()
+        }} // cit node
+      }} // timeout


pipeline_steps/run_tests.groovy
```diff
@@ -0,0 +1,17 @@
+def run(Map args) {
+  common.conditionalStage(
+    stage_name: "Run IRR Tests",
+    stage: {
+      environment_vars = args.environment_vars + common.get_deploy_script_env()
+      withEnv(environment_vars) {
+        ansiColor('xterm') {
+          dir("/opt/rpc-openstack/") {
```
cloudnull: I'd like this path to be more dynamic

rpc_jobs/irr_role_test.yml
```diff
@@ -0,0 +1,128 @@
...
+            environment_vars = [
+              "DEPLOY_HAPROXY=yes",
+              "DEPLOY_AIO=no",
+              "DEPLOY_MAAS=no",
+              "DEPLOY_TEMPEST=no",
+              "DEPLOY_SWIFT=${{DEPLOY_SWIFT}}",
+              "DEPLOY_CEPH=${{DEPLOY_CEPH}}",
+              ]
+            aio_prepare.prepare()
+            deploy.deploy_sh(environment_vars: environment_vars)
```
cloudnull: this should likley go away

rpc_jobs/irr_role_test.yml
```diff
@@ -0,0 +1,128 @@
...
+            deploy.addChecksumRule()
+            run_tests.run()
+          }} catch (e) {{
+            print(e)
+            throw e
+          }} finally {{
+            common.archive_artifacts()
```
cloudnull: This can be removed

cloudnull: done

rpc_jobs/irr_role_test.yml
cloudnull: done

rpc_jobs/irr_role_test.yml
```diff
@@ -0,0 +1,101 @@
...
+    #       branches is the branch pattern to match for PR Jobs.
+    series:
+      - master:
+          branch: master
+          branches: "master"
+    image:
+      - xenial
+      - trusty
+    context:
+      - base
```
mattt416: What does base mean, and since there's just a single element do we need it?

rpc_jobs/irr_role_test.yml
```diff
@@ -0,0 +1,101 @@
...
+    ztrigger:
+      - pr:
+          CRON: ""
+      - periodic:
+          branches: "do_not_build_on_pr"
+          NUM_TO_KEEP: 10
+    exclude: []
```
mattt416: You can remove the exclude if there's nothing to exclude.

pipeline_steps/run_tests.groovy
```diff
@@ -0,0 +1,39 @@
...
+                submoduleCfg: [],
+                userRemoteConfigs: [
+                  [
+                    url: "https://github.com/${env.ghprbGhRepository}.git",
+                    refspec: '+refs/pull/*:refs/remotes/origin/pr/* +refs/heads/*:refs/remotes/origin/*'
+                  ]
+                ]
+              ]) // checkout
+              sh """#!/bin/bash
+              # Backticks used because Groovy does not like dollar sign and parenthesizes
```
mattt416: \$( ... ) ?

cloudnull: Groovy fails to run this it tries to interpret the $(...)

rpc_jobs/irr_role_test.yml
cloudnull: done.

rpc_jobs/irr_role_test.yml
cloudnull: reserved for other scenarios later.

pipeline_steps/run_tests.groovy
```diff
@@ -0,0 +1,39 @@
...
+              pushd `basename ${env.ghprbGhRepository}`
+                run_tests.sh
```
mattt416: rpc-maas/run_tests.sh is not executable.

cloudnull: done.

pipeline_steps/run_tests.groovy
mattt416: Have a look at https://github.com/rcbops/rpc-gating/blob/master/pipeline_steps/horizon.groovy#L54

pipeline_steps/run_tests.groovy
```diff
@@ -0,0 +1,39 @@
...
+                extensions: [[$class: 'CleanCheckout']],
+                submoduleCfg: [],
+                userRemoteConfigs: [
+                  [
+                    url: "https://github.com/${env.ghprbGhRepository}.git",
+                    refspec: '+refs/pull/*:refs/remotes/origin/pr/* +refs/heads/*:refs/remotes/origin/*'
+                  ]
+                ]
+              ]) // checkout
+              sh """#!/bin/bash
```
mattt416: I'd probably separate the clone and execution of the tests into two functions, certainly not necessary but may make it easier to read/update in the future.

pipeline_steps/run_tests.groovy
cloudnull: done.

rpc_jobs/irr_role_test.yml
```diff
@@ -0,0 +1,100 @@
+- project:
+    name: 'RPC-IRR-Jobs'
+    # Note: branch is the branch for periodics to build
+    #       branches is the branch pattern to match for PR Jobs.
+    series:
+      - master:
+          branch: master
+          branches: "master"
+    image:
+      - xenial
```
mattt416: You'll probably need to set IMAGE here for xenial/trusty, ie:

```IMAGE: "Ubuntu 16.04 LTS (Xenial Xerus) (PVHVM)"```

Comment (cloudnull) [inline rpc_jobs/irr_role_test.yml]: done.

File: rpc_jobs/irr_role_test.yml [hunk 10]
```diff
@@ -0,0 +1,100 @@
...
+          only-trigger-phrase: false
+          white-list-target-branches:
+            - "{branches}"
+          auth-id: "github_account_rpc_jenkins_svc"
+          status-context: 'CIT/{image}_{context}'
+          cancel-builds-on-update: true
+
+    dsl: |
+      // CIT Slave node
+      timeout(time: 8, unit: 'HOURS'){{
```
cloudnull: done.

pipeline_steps/run_tests.groovy
```diff
@@ -0,0 +1,38 @@
...
+              pushd \$(basename ${env.ghprbGhRepository})
+                bash ./run_tests.sh
```
hughsaunders: -x ?

pipeline_steps/run_tests.groovy
```diff
@@ -0,0 +1,38 @@
...
+                submoduleCfg: [],
+                userRemoteConfigs: [
+                  [
+                    url: "https://github.com/${env.ghprbGhRepository}.git",
+                    refspec: '+refs/pull/*:refs/remotes/origin/pr/* +refs/heads/*:refs/remotes/origin/*'
+                  ]
+                ]
+              ]) // checkout
+              sh """#!/bin/bash
+              pushd \$(basename ${env.ghprbGhRepository})
```
hughsaunders: I don't think `checkout()` will create a subdir for the repo; `env.workspace` will become the checkout, so this pushd will probably fail. I think its generally a good idea to checkout to a subdir so that other repos can be checked out to their own subdirs if required in future.

rpc_jobs/irr_role_test.yml
```diff
@@ -0,0 +1,102 @@
...
+      # IRR REPOS ARE ADDED HERE
+      - github:
+          url: https://github.com/rcbops/rpc-maas
+    parameters:
+      - rpc_gating_params
+      - single_use_slave_params:
+         IMAGE: "{IMAGE}"
+         FLAVOR: "{FLAVOR}"
+         REGION: "{REGION}"
+         IRR_CONTEXT: "{context}"
```
hughsaunders: single_use_slave_params doesn't make use of context, image, series.

rpc_jobs/irr_role_test.yml
```diff
@@ -0,0 +1,102 @@
...
+      // CIT Slave node
+      timeout(time: 2, unit: 'HOURS'){{
+        node() {{
+          dir("rpc-gating") {{
+              git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
+              common = load 'pipeline_steps/common.groovy'
+              pubcloud = load 'pipeline_steps/pubcloud.groovy'
+              run_tests = load 'pipeline_steps/run_tests.groovy'
+          }}
+          // try within pubcloud node so we can archive archive_artifacts
```
hughsaunders: This comment is misleading as the try it is after does not run on a pubcloud node. The whole try block could be removed as it doesn't do anything useful on catch or finally.

rpc_jobs/irr_role_test.yml
```diff
@@ -0,0 +1,102 @@
...
+    series:
+      - master:
+          branch: master
+          branches: "master"
+    image:
+      - xenial:
+          IMAGE: "Ubuntu 16.04 LTS (Xenial Xerus) (PVHVM)"
+      - trusty:
+          IMAGE: "Ubuntu 14.04 LTS (Trusty Tahr) (PVHVM)"
+    context:
```
hughsaunders: Is this axis still useful? Kept in so that additional scenarios can be added later?

pipeline_steps/run_tests.groovy
```diff
@@ -0,0 +1,38 @@
...
+                bash ./run_tests.sh
+              popd
+              """
+            } // dir
+          } // ansiColor
+        } catch (e) {
+          print(e)
+          throw e
+        } finally {
+          common.archive_artifacts()
```
hughsaunders: This may not be relevant, depends what run_tests.sh does.

rpc_jobs/irr_role_test.yml
```diff
@@ -0,0 +1,102 @@
+- project:
```
hughsaunders: Would it make sense to have a repo axis? Then a couple of  jobs can be generated per repo, which will keep the histories independent. It looks like this JJB project is aimed at generating a set of jobs, but that one set of jobs will deal with every IRR?

rpc_jobs/irr_role_test.yml
```diff
@@ -0,0 +1,102 @@
...
+    NUM_TO_KEEP: 30
+
+    # TEMPLATE
+    name: 'RPC-IRR_{series}-{image}-{context}-{ztrigger}'
+    project-type: workflow
+    concurrent: true
+    properties:
+      - build-discarder:
+          num-to-keep: "{NUM_TO_KEEP}"
+      # IRR REPOS ARE ADDED HERE
```
hughsaunders: I think adding repos here will cause all the generated jobs to run against all the repos, which may or may not be whats anticipated - see comment above about repo axis.

rpc_jobs/irr_role_test.yml
odyssey4me: Yeah, there are plans to add a ceph context at some point, and perhaps others. This serves as the 'scenario' to test.

rpc_jobs/irr_role_test.yml
odyssey4me: The intent here was to provide the context, image and series to the environment the test runs in so that run_tests.sh could perhaps do things based on their values. If there's a better way to do this (without making them editable params on the job) then please let us know how.

pipeline_steps/run_tests.groovy
odyssey4me: Why not just execute ``./run_tests.sh`` ?

rpc_jobs/irr_role_test.yml
odyssey4me: What difference would it make? Will this execute a job against all repositories for every PR to one of those repositories right now? Or will it only execute against the repository a PR is requested against.

rpc_jobs/irr_role_test.yml
odyssey4me: Fair point.

pipeline_steps/run_tests.groovy
odyssey4me: Is there any harm in running it?

pipeline_steps/run_tests.groovy
odyssey4me: Personally I don't think it'll make that much difference as it all stands now. It may be an idea later to change the rpccheckout method to be more generic and then re-use it, but for now I think that's probably overkill.

pipeline_steps/run_tests.groovy
odyssey4me: Perhaps the ``dir(env.WORKSPACE)`` should then be changed to ``dir("${env.WORKSPACE}/${env.ghprbGhRepository}")`` and the basename can be removed from this line?

rpc_jobs/irr_role_test.yml
odyssey4me: Ah, if that is the case then there will need to be a repo axis. I don't think that the property can be set in the axis though, can it? If not then how would one implement the axis so that it filters properly?

pipeline_steps/run_tests.groovy
cloudnull: done.

pipeline_steps/run_tests.groovy
cloudnull: Debug levels will be handled in the test script.

pipeline_steps/run_tests.groovy
hughsaunders: Because the `sh` call is within the same `dir` block as `checkout` the working directory for the shell will be the root of the clone, so trying to cd into that dir will fail. If dir is changed as @odyssey4me suggests then this pushd can be removed.

pipeline_steps/run_tests.groovy
cloudnull: We hope to gather artifacts from the build for further review after a gate job. I'm inclined to leave this as is for now.

rpc_jobs/irr_role_test.yml
cloudnull: ++ We only have one context but more will be added later.

pipeline_steps/run_tests.groovy
hughsaunders: The form `bash script.sh` will ignore opts set in the script's `#!` line, but if the script sets options via the `set` command then this will be fine.

rpc_jobs/irr_role_test.yml
cloudnull: done.

rpc_jobs/irr_role_test.yml
cloudnull: Done.

rpc_jobs/irr_role_test.yml
cloudnull: ++ as @odyssey4me stated, the intention was to ensure these vars were passed into the environment. If there's a better way to do this please let me know.

rpc_jobs/irr_role_test.yml
cloudnull: done.

pipeline_steps/run_tests.groovy
hughsaunders: No harm in running it, more of a heads up that, `common.archive_artifacts` is not a general artifact collection function, it is specifically aimed at collecting logs and config from rpc deploys, so additional archiving may be required for tests that aren't a straightforward rpc deploy.

rpc_jobs/irr_role_test.yml
hughsaunders: Scenario is fine, it was called `context` because thats what github calls the labels next to statuses in PRs.

rpc_jobs/irr_role_test.yml
hughsaunders: This doesn't do what you want it to do. The dict under single_use_slave params is passed to the single_use_slave_params macro, which doesn't make use of the additional values so they will be ignored. If you want to ensure those template vars get passed to the job environment then additional parameters will need to be defined. These can be defined in the job, or in params.yml

rpc_jobs/irr_role_test.yml
```diff
@@ -0,0 +1,94 @@
+- project:
+    name: 'RPC-IRR-Jobs'
+    # Note: branch is the branch for periodics to build
+    #       branches is the branch pattern to match for PR Jobs.
+    repo:
+      # IRR REPOS ARE ADDED HERE
+      - github:
+          url: https://github.com/rcbops/rpc-maas
```
odyssey4me: I *think* that it will need to be something like this instead:

```
  repo:
    # IRR REPOS ARE ADDED HERE
    - rpc_maas:
      properties:
        - github:
            url: https://github.com/rcbops/rpc-maas
```

rpc_jobs/irr_role_test.yml
```diff
@@ -0,0 +1,94 @@
...
+    context:
+      - base
+    ztrigger:
+      - pr:
+          CRON: ""
+      - periodic:
+          branches: "do_not_build_on_pr"
+          NUM_TO_KEEP: 10
+    jobs:
+      - 'RPC-IRR_{series}-{image}-{context}-{ztrigger}'
```
odyssey4me: This will need to change to:

```
- 'RPC-IRR_{repo}-{series}-{image}-{context}-{ztrigger}'
```

rpc_jobs/irr_role_test.yml
```diff
@@ -0,0 +1,95 @@
...
+    properties:
+      - build-discarder:
+          num-to-keep: "{NUM_TO_KEEP}"
+    parameters:
+      - rpc_gating_params
+      - single_use_slave_params:
+         IMAGE: "{IMAGE}"
+         FLAVOR: "{FLAVOR}"
+         REGION: "{REGION}"
+      - irr_params:
```
odyssey4me: Change to:

```
- string:
  name: IRR_CONTEXT
  default: "{context}"
- string:
  name: IRR_IMAGE
  default: "{{ image }}"
- string:
  name: IRR_SERIES
  default: "{{ series }}"
```

rpc_jobs/irr_role_test.yml
cloudnull: done.

rpc_jobs/irr_role_test.yml
cloudnull: done.

rpc_jobs/irr_role_test.yml
```diff
@@ -0,0 +1,102 @@
...
+        properties:
+          - github:
+              url: https://github.com/rcbops/rpc-maas
+    series:
+      - master:
+          branch: master
+          branches: "master"
+    image:
+      - xenial:
+          IMAGE: "Ubuntu 16.04 LTS (Xenial Xerus) (PVHVM)"
```
odyssey4me: I'm confused as to why this can't just inherit the image definition from the defaults?

rpc_jobs/irr_role_test.yml
hughsaunders: In the project:
```
repo:
  - rpc_maas:
      irr_repo_url: https://github.com/rcbops/rpc-maas
```

In the job-template:
```
properties:
  - github:
      url: {{irr_repo_url}}
```

rpc_jobs/irr_role_test.yml
odyssey4me: Ah, I see. trusty is the default image, so we only need to override the image for the ``xenial`` axis like so:

```
      - trusty
      - xenial:
```

hughsaunders: Yeah

rpc_jobs/irr_role_test.yml
```diff
@@ -0,0 +1,102 @@
+- project:
+    name: 'RPC-IRR-Jobs'
+    # Note: branch is the branch for periodics to build
+    #       branches is the branch pattern to match for PR Jobs.
+    repo:
```
hughsaunders: In the project (here):
```
  repo:
    - rpc_maas:
       irr_repo_url: https://github.com/rcbops/rpc-maas
```
in the job-template:
```
  properties:
    - github:
      url: {irr_repo_url}
```

rpc_jobs/irr_role_test.yml
```diff
@@ -0,0 +1,102 @@
...
+              Prepare Deployment
+              Run IRR Tests
+              Cleanup
+              Destroy Slave
+    triggers:
+      - github-pull-request:
+          org-list:
+            - rcbops
+          github-hooks: true
+          trigger-phrase: '.*recheck_cit_all.*|.*recheck_cit_{image}_{context}.*'
```
odyssey4me: Should this now change to include the whole set of ``{repo}-{series}-{image}-{context}``?

rpc_jobs/irr_role_test.yml
cloudnull: done.

rpc_jobs/irr_role_test.yml
cloudnull: done.

rpc_jobs/irr_role_test.yml
cloudnull: done.

pipeline_steps/run_tests.groovy
```diff
@@ -0,0 +1,71 @@
...
+}
+
+def irr_archive_artifacts(){
+  dir("${env.WORKSPACE}/logs") {
+    print("Running Log Collection")
+    try{
+      sh """#!/bin/bash
+      d="artifacts_\${BUILD_TAG}"
+      touch "\$d".marker
+      tar -C "${WORKSPACE}/logs" -cjf "${env.WORKSPACE}/\$d".tar.bz2 .
```
odyssey4me: Will this try to tarball itself? Maybe add an exclude for the file being created.

odyssey4me: Never mind, I see that it's creating a tarball in the parent directory.

cloudnull: yes. good catch. changed.

pipeline_steps/run_tests.groovy
```diff
@@ -0,0 +1,72 @@
+def run_irr_tests() {
```
hughsaunders: Last thing, can you call this irr_tests.groovy or something similar, so that the file name reflects the functions defined within.

cloudnull: done.

pipeline_steps/irr_role_tests.groovy
```diff
@@ -0,0 +1,72 @@
...
+  common.conditionalStage(
+    stage_name: "Run IRR Tests",
+    stage: {
+      pubcloud.runonpubcloud {
+        try {
+          ansiColor('xterm') {
+            dir("${env.WORKSPACE}/${env.ghprbGhRepository}") {
+              print("Triggered by PR: ${env.ghprbPullLink}")
+              checkout([$class: 'GitSCM',
+                branches: [[name: "origin/pr/${env.ghprbPullId}/merge"]],
```
hughsaunders: This won't work for periodic jobs

cloudnull: I don't intend this to work with a periodic job at this point. only PRs for now.

rpc_jobs/irr_role_test.yml
```diff
@@ -0,0 +1,96 @@
...
+    name: 'RPC-IRR-Jobs'
+    # Note: branch is the branch for periodics to build
+    #       branches is the branch pattern to match for PR Jobs.
+    repo:
+      # IRR REPOS ARE ADDED HERE
+      - rpc_maas:
+          irr_repo_url: https://github.com/rcbops/rpc-maas
+    series:
+      - master:
+          branch: master
```
hughsaunders: This should be removed because it isn't used, same for `branch:` in the template defaults.

odyssey4me: How does ``white-list-target-branches`` get populated then?

rpc_jobs/irr_role_test.yml
```diff
@@ -0,0 +1,95 @@
...
+      Allocate Resources,
+      Connect Slave,
+      Prepare Deployment,
+      Run IRR Tests,
+      Cleanup,
+      Destroy Slave
+    NUM_TO_KEEP: 30
+    # TEMPLATE
+    name: 'RPC-IRR_{repo}-{series}-{image}-{context}'
+    branch: master
```
odyssey4me: The ``branch`` variable is never used.

PR #160, opened 2017-04-28
Temporarily disable cinder-backup from being tested

mattt416:
The cinder-backup service is usable on an AIO deployment, but
deploying on a multi-node deployment seems to result in the service not
working.  I'm not a storage expert, but from looking briefly it appears
that the open-iscsi package is missing on the storage nodes, which
prevents the service from backing up a volume.  The package open-iscsi
gets installed by nova on the compute nodes, and since an AIO deploys
nova-compute, cinder-volume, and cinder-backup to the host itself this
explains why cinder-backup is functional there.

I will create an upstream bug and fix for this but in the mean time
it's probably best to just disable this from being tested.

Connects https://github.com/rcbops/u-suk-dev/issues/1643

### DIFF
--- a/playbooks/vars/all.yml
+++ b/playbooks/vars/all.yml
@@ -27,7 +27,11 @@ default_gating_overrides:
       attach_encrypted_volume: false
     volume-feature-enabled:
       snapshot: true
-  cinder_service_backup_program_enabled: true
+  # NOTE(mattt): This functionality is currently broken on multi-node
+  #              deployments, disabling until this is addressed in
+  #              openstack-ansible-os_cinder
+  cinder_service_backup_program_enabled: false
+  #
   # This is being increased from the default of 85 as the default value may be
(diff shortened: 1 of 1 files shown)


BjoernT:
Iscsi is not working since we install it inside a container (netlink bug). Only Swift and NFS driver work, we did only test with swift driver.

mattt416:
Hi @BjoernT, our tests should have been using the swift driver for cinder backup since that is the [default](https://github.com/openstack/openstack-ansible-os_cinder/blob/stable/newton/defaults/main.yml#L138), however for whatever reason it still seemed to want `/etc/iscsi/initiatorname.iscsi`.  I'll need to double-check my work but please provide any additional information you may have on hand as like I say I'm not overly familiar w/ the storage piece (when it comes to iscsi etc.)

Thanks!

--Matt

PR #164, opened 2017-05-02
Add Jenkins node deletion to periodic cleanup

mkam:
Deleting offline nodes: https://rpc.jenkins.cit.rackspace.net/job/mkam-Periodic-Cleanup/4/console

I tested that a node coming online didn't get deleted by manually running the script since I couldn't force the exact timing of that situation using my Jenkins job.

Connects https://github.com/rcbops/u-suk-dev/issues/1605

### DIFF
--- a/scripts/periodic_cleanup.py
+++ b/scripts/periodic_cleanup.py
@@ -10,6 +10,7 @@ from dateutil.tz import tzutc
 import os
 
 import pyrax
+import jenkins_node
 
 
 def get_env_vars():
@@ -56,5 +57,11 @@ if __name__ == "__main__":
     cleanup_instances(
         cs_client=cs,
         age_limit=args["age_limit"],
+        instance_prefix=args["instance_prefix"],
+    )
+
+    jenkins_client = jenkins_node.get_jenkins_client()
+    jenkins_node.delete_inactive_nodes(
+        jenkins=jenkins_client,
         instance_prefix=args["instance_prefix"],
     )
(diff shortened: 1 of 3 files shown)


scripts/periodic_cleanup.py
```diff
@@ -46,6 +53,16 @@ def cleanup_instances(cs_client, age_limit, instance_prefix):
             server.delete()
 
 
+def cleanup_jenkins_nodes(jenkins_client, instance_prefix):
+    for node_id, node in jenkins_client.get_nodes().iteritems():
+        # Ignore nodes that are coming online for the first time
+        # which won't have an offline cause.
+        if (node_id.startswith(instance_prefix) and not node.is_online()
+                and node._data.get("offlineCauseReason")):
```
git-harry: Reaching into the class for `_data` is not ideal. I think it would be better to use an alternative if possible. I had a skim through the module and it looks like `node.poll().get("offlineCauseReason")` should work.

mkam:
Changed to using poll() to get the offline reason and reran here: https://rpc.jenkins.cit.rackspace.net/job/mkam-Periodic-Cleanup/7/console

mattt416:
@mkam could you not just call jenkins_node.py for each cloud server that we delete?  The benefit there is slightly less code, the downside is we don't do a wider cleanup sweep.

mkam:
@mattt416 I started doing that originally, but I was thinking that we've had a lot of situations where we manually cleaned up the servers because we couldn't wait for the periodic time limit to pass. In those cases, then we'd still to manually delete a bunch of offline nodes again.

mattt416:
@mkam makes sense!  What about moving this functionality to jenkins_node.py then?

mkam:
Cleanup: https://rpc.jenkins.cit.rackspace.net/job/mkam-Periodic-Cleanup/10/console
Connect/Delete Slave build: https://rpc.jenkins.cit.rackspace.net/job/mkam-RPC-AIO/5/console

PR #166, opened 2017-05-03
Clean up new RPC-IRR instances

mattt416:
This commit updates periodic_cleanup.py so we can pass a pattern to use
when cleaning up instances.  We also update the periodic cleanup job to
cleanup ra|ri instances.

Connects https://github.com/rcbops/u-suk-dev/issues/1655

### DIFF
--- a/scripts/periodic_cleanup.py
+++ b/scripts/periodic_cleanup.py
@@ -1,13 +1,14 @@
 #!/usr/bin/env python
 
-# This script deletes instances whose name starts with "ra" if they are in
-# error state or older than AGE_LIMIT hours.
+# This script deletes instances whose name matches a supplied pattern if they
+# are in error state or older than AGE_LIMIT hours.
 
 
 import datetime
 import dateutil.parser
 from dateutil.tz import tzutc
 import os
+import re
 
 import pyrax
 import jenkins_node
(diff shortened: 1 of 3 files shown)


mattt416:
Not tested yet -- please hold off on merging until this has been reviewed and tested.  :)

mattt416:
Tested here: https://rpc.jenkins.cit.rackspace.net/job/Periodic-Cleanup/555/console

Didn't clean up anything as far as I can tell, but didn't blow anyway anything either.  :)

mkam:
Can you update jenkins_node.py too now that it matches prefixes there as well?

mkam:
Rebased and added the regex matching to the jenkins node deletion as well:

https://rpc.jenkins.cit.rackspace.net/job/Periodic-Cleanup/585/console

scripts/periodic_cleanup.py
```diff
@@ -34,7 +35,7 @@ def cleanup_instances(cs_client, age_limit, instance_prefix):
```
git-harry: It's probably worth removing the default of `"ra"` on line 21.

mkam: Done!

PR #169, opened 2017-05-05
Check Groovy syntax with JJB templates

hughsaunders:
This commit extracts groovy from JJB templates so they can be
checked for syntax. Previously JJB templates were only checked for
yaml syntax and JJB semantics so groovy errors could slip in.

This commit also reduces lint time by scanning all groovy files
with a single JVM rather than starting one up for each file.

### DIFF
--- a/rpc_jobs/lem_multi_node_aio.yml
+++ b/rpc_jobs/lem_multi_node_aio.yml
@@ -146,14 +146,15 @@
             ) // deploy_sh
             multi_node_aio_prepare.multi_node_aio_networking()
             node(){{
-            maas.prepare(instance_name: instance_name)
-            maas.deploy()
-            maas.verify()
-            tempest.tempest("infra1", "deploy1")
-            kibana.kibana("deploy1")
-            holland.holland()
-          }} // hardware node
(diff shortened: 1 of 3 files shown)


rpc_jobs/lem_multi_node_aio.yml
```diff
@@ -146,14 +146,15 @@
...
-            maas.verify()
-            tempest.tempest("infra1", "deploy1")
-            kibana.kibana("deploy1")
-            holland.holland()
-          }} // hardware node
-          ssh_slave.destroy()
-          maas.entity_cleanup(instance_name: instance_name)
-        }} // cit node
-      }} // timeout
+              maas.prepare(instance_name: instance_name)
```
hughsaunders: This job contained a syntax error, so had to be fixed in order to not break the gate with the new lint job.

hughsaunders: While I fixed the syntax, the job still looks broken, we shouldn't be trying to run maas/tempeset etc on the CIT node.

hughsaunders:
Closing in favour of #170 as that one has m04r kanban.

PR #173, opened 2017-05-08
Fix doc skip for artifact jobs

odyssey4me:
The doc skip check was misplaced and therefore not
skipping all jobs properly.


evrardjp:
Got it!

PR #175, opened 2017-05-08
Support liberty Kibana testing

mkam:
Takes into account the different location of the credentials and uses branches of the kibana-selenium repo that correspond to the version of RPC.

AIO upgrade build: https://rpc.jenkins.cit.rackspace.net/job/mkam-RPC-AIO_mitaka-upgrade/5/
AIO, non-upgrade mitaka build: https://rpc.jenkins.cit.rackspace.net/job/mkam-RPC-AIO/10/
Multi-node AIO build: https://rpc.jenkins.cit.rackspace.net/job/mkam-OnMetal-Multi-Node-AIO/4/

Connects https://github.com/rcbops/u-suk-dev/issues/1663

### DIFF
--- a/pipeline_steps/kibana.groovy
+++ b/pipeline_steps/kibana.groovy
@@ -1,21 +1,21 @@
-def kibana(vm=null){
+def kibana(branch, vm=null){
   common.conditionalStage(
     stage_name: "Prepare Kibana Selenium",
     stage: {
-      kibana_prep()
+      kibana_prep(branch)
     }
   )
   common.conditionalStage(
     stage_name: "Kibana Tests",
     stage: {
-      kibana_tests(vm)
+      kibana_tests(branch, vm)
     }
   )
 }
 
-def kibana_prep(){
+def kibana_prep(branch){
   dir("kibana-selenium") {
-    git url: "https://github.com/rcbops-qe/kibana-selenium.git", branch: "master"
+    git url: "https://github.com/rcbops-qe/kibana-selenium.git", branch: "${branch}"
 
     sh """#!/bin/bash
       # The phantomjs package on 16.04 is buggy, see:
@@ -42,10 +42,10 @@ def kibana_prep(){
   }
 }
 
-def kibana_tests(vm=null){
+def kibana_tests(branch, vm=null){
   try {
     dir("kibana-selenium") {
-      git url: "https://github.com/rcbops-qe/kibana-selenium.git", branch: "master"
+      git url: "https://github.com/rcbops-qe/kibana-selenium.git", branch: "${branch}"
 
       if(vm != null){
         sh """#!/bin/bash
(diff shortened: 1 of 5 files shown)


pipeline_steps/kibana.groovy
```diff
@@ -1,4 +1,4 @@
-def kibana(vm=null){
+def kibana(branch="master", vm=null){
```
git-harry: Given `branch` is always specified and using this function without specifying it could cause the wrong code to be tested, I think it would be better not to specify a default.

pipeline_steps/kibana.groovy
```diff
@@ -62,7 +62,9 @@ def kibana_tests(vm=null){
         export PATH=\$PATH:./phantomjs-2.1.1-linux-x86_64/bin
         # Remove any existing screenshots from old runs
         rm -f *.png
-        python conf-gen.py --secure --password-file /etc/openstack_deploy/user_rpco_secrets.yml \
+
+        export PASSWORD=\$(grep -Ir kibana_password /etc/openstack_deploy/ | head -1 | sed 's/.*\\: //')
```
git-harry: This seems to work as expected. It looks like you could have changed this to the following to achieve the same thing:
`python conf-gen.py --secure --password-file /etc/openstack_deploy`

I would have thought `kibana_password` would only be set once, your use of `head -1` implies otherwise. If it was defined multiple times I think the last would have been used by Ansible and not the first. That said, it would probably be better if it failed when the password was set multiple times given that shouldn't happen.

pipeline_steps/kibana.groovy
```diff
@@ -42,10 +42,10 @@ def kibana_prep(){
   }
 }
 
-def kibana_tests(vm=null){
+def kibana_tests(branch="master", vm=null){
```
git-harry: Same as above, I think it would be better not to set a default branch.

pipeline_steps/kibana.groovy
mkam: For liberty to mitaka upgrades, I was seeing the kibana password in user_extras_secrets.yml and user_rpco_secrets.yml, which were identical files. The first one is what liberty uses and the second file is standard for mitaka and newton. I'll look into why there are duplicate files.

mkam: The upgrade script creates `user_rpco_secrets.yml` based on `user_extras_secrets.yml` as part of the upgrade process, but doesn't delete `user_extras_secrets.yml` afterwards. @cjloader suggested that it might be left as a backup.

https://github.com/rcbops/rpc-openstack/blob/mitaka-13.1/scripts/test-upgrade.sh#L49
https://github.com/rcbops/rpc-openstack/blob/mitaka-13.1/scripts/update-secrets.sh

I'm going to change the `head -1` to `tail -1`, though, for now in case we want to stay with this method.

pipeline_steps/kibana.groovy
```diff
@@ -8,7 +8,7 @@ def kibana(vm=null){
   common.conditionalStage(
     stage_name: "Kibana Tests",
     stage: {
-      kibana_tests(vm)
+      kibana_tests(branch, vm)
     }
   )
 }
```
git-harry: What about kibana_prep? The requirements that get installed into the venv are always from the master branch. By the looks of it they are currently the same but that is not guaranteed.

mkam: Added now!

git-harry:
@mkam sorry, one more thing - rpc_jobs/lem_multi_node_aio.yml calls the function `kibana` but hasn't been updated.

mkam:
@git-harry No worries, it's my bad for forgetting it in the first place! Added it to that job as well.

PR #179, opened 2017-05-10
UG-575 Add Xenial archive location

mkam:
Updated to check for artifacts in locations for both Trusty and Xenial and added some logging for the archive steps.

Xenial: https://rpc.jenkins.cit.rackspace.net/job/mkam-RPC-AIO/21/
Trusty: https://rpc.jenkins.cit.rackspace.net/job/mkam-RPC-AIO/25/

Connects https://github.com/rcbops/u-suk-dev/issues/1591


hughsaunders:
Checked the archives for both jobs linked and they look good, thanks @mkam

PR #177, opened 2017-05-10
UG-586 Add Jira Notification for PR job results

hughsaunders:
Github pull request builder automatically updates PRs, but we now also
need to provide that information to the relevant JIRA issue.

Connects rcbops/u-suk-dev#1664

PR #183, opened 2017-05-16
UG-586 Add functions for JIRA notification

hughsaunders:
Github pull request builder automatically updates PRs, but we now also
need to provide that information to the relevant JIRA issue.

Connects rcbops/u-suk-dev#1664

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -439,4 +439,49 @@ def is_doc_update_pr(git_dir) {
   return is_doc_update_pr
 }
 
+/* Look for JIRA issue key in commit messages for commits in the source branch
+ * that aren't in the target branch.
+ * This function uses environment variables injected by github pull request
+ * builder and so can only be used for PR triggered jobs
+ */
+def get_jira_issue_key(repo_path="rpc-openstack"){
+  def key_regex = "^[a-zA-Z][a-zA-Z0-9_]+-[1-9][0-9]*"
+  dir(repo_path){
+    commit_titles = sh(
+      returnStdout: true,
+      script: "git log --pretty=%s origin/${ghprbTargetBranch}..${ghprbSourceBranch}").split('\n')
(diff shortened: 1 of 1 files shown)


hughsaunders:
This is a continuation of #177 but divded up, so we don't change all jobs with a single PR. This way we can merge the common functions, then gradually add into the jobs.

mancdaz:
@git-harry could you please review

hughsaunders:
@git-harry fyi the code hasn't changed since 177.

pipeline_steps/common.groovy
```diff
@@ -439,4 +439,48 @@ def is_doc_update_pr(git_dir) {
...
+    for (def i=0; i<=commit_titles.size(); i++){
+      try{
+        key = (commit_titles[i] =~ key_regex)[0]
+        print ("Found Jira Issue Key: ${key}")
+        return key
+      } catch (e){
+        continue
+      }
+    }
+    throw new Exception("JIRA Issue key not found in any of the commits"
```
git-harry: If someone read this in the log would they know what to do? Maybe this exception should be a bit more verbose, providing sufficient detail by itself to allow someone to understand how to fix the issue.

pipeline_steps/common.groovy
```diff
@@ -439,4 +439,48 @@ def is_doc_update_pr(git_dir) {
...
+    print ("Not a PR job, so not attempting to add a Jira comment")
+    return
+  }
+  try{
+    key = get_jira_issue_key(repo_path)
+    jiraComment(issueKey: key,
+                body: body)
+    print "Jira Comment: [${key}] ${body}"
+  } catch (e){
+    print (e)
```
git-harry: It might be useful to prefix the error string with a standard bit of blurb to explain, in general, what is failing.

hughsaunders: Added location context to the exception print.

PR #187, opened 2017-05-16
UG-586 Add Jira comments to AIO Jobs

hughsaunders:
Depends on #183
Connects rcbops/u-suk-dev#1664

### DIFF
--- a/rpc_jobs/rpc_aio.yml
+++ b/rpc_jobs/rpc_aio.yml
@@ -169,6 +169,7 @@
 
     dsl: |
       // CIT Slave node
+      currentBuild.result = 'SUCCESS'
       timeout(time: 8, unit: 'HOURS'){{
         node() {{
           dir("rpc-gating") {{
@@ -226,9 +227,11 @@
               }}
             }} catch (e) {{
               print(e)
+              currentBuild.result = 'FAILURE'
               throw e
             }} finally {{
               common.archive_artifacts()
+              common.safe_jira_comment("${{currentBuild.result}}: [${{env.BUILD_TAG}}|${{env.BUILD_URL}}]")
             }}
           }} //pubcloud slave
         }} // cit node


hughsaunders:
recheck_dev_lint

rpc_jobs/rpc_aio.yml
```diff
@@ -169,6 +169,7 @@
 
     dsl: |
       // CIT Slave node
+      currentBuild.result = 'SUCCESS'
```
git-harry: Does this mean if a failure happens between here and line 199 the build will be marked as `SUCCESS`?

hughsaunders: No, an exception will change the build result, without having to change it explicitly. The only reason to set it manually is so that it can be read.

git-harry: That makes sense. Does that mean `delPubCloudSlave` failures could cause an inconsistency? Are you counting on the likelihood of failure there being low?

hughsaunders: Failures in delPubCloudSlave should be caught by the catch on line 227, so that shouldn't cause an inconsistency, whats the situation you are concerned about?

git-harry: Maybe I misunderstand how this works, `runonpubcloud` takes the argument `body` which I'd assumed was lines 198-235 which would mean that `delPubCloudSlave` runs afterwards. If that's not the case, what is `body`?

hughsaunders: I see what you mean. Errors in delPubCloudSlave won't be caught because the try block is within pubcloud.runonpubcloud. That could be fixed by using a try/catch within the current runonpubcloud finally block. That could log and squash the exception so that teardown failures don't cause builds to fail, however thats an existing bug and should probably be fixed separately.

PR #190, opened 2017-05-17
UG-528 Improve security of webhooks node

git-harry:
The webhooks host is long-lived, to improve security the following
changes are applied:
- Restrict SSH access to specific IP addresses
- Replace use of lineinfile with template for modifying sshd_config to
  prevent conflicting values for the same configuration option
- Enable the automatic installation of APT security package update

### DIFF
--- a/playbooks/webhooks.yml
+++ b/playbooks/webhooks.yml
@@ -1,31 +1,15 @@
 ---
-- name: Configure Webhook Proxy Node
+- name: Secure Webhook Proxy Node
   hosts: job_nodes
   user: root
   vars:
-    github_cidr: "192.30.252.0/22"
+    ip_address_whitelist: "{{ lookup('env', 'SSH_IP_ADDRESS_WHITELIST') }}"
   pre_tasks:
-    - name: "Configure SSHD"
-      lineinfile:
-        dest: /etc/ssh/sshd_config
-        regexp: "{{item.regexp}}"
-        line: "{{item.line}}"
-      with_items:
-        - regexp: "GatewayPorts"
-          line: "GatewayPorts no"
-        - regexp: "PasswordAuthentication"
-          line: "PasswordAuthentication no"
-        - regexp: "ClientAliveInterval"
-          line: "ClientAliveInterval 15"
-
     - name: Install apt packages
       apt:
-        pkg: "{{ item }}"
+        pkg: "ufw"
         state: installed
         update_cache: yes
-      with_items:
-        - apache2-utils
-        - ufw
 
     - name: Reset UFW
       ufw:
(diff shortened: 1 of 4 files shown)


playbooks/setup_jenkins_slave.yml
```diff
@@ -2,6 +2,78 @@
...
+        port: 22
+        src: "{{ item }}"
+      with_items: "{{ ip_address_whitelist }}"
+
+    - name: Enable firewall
+      ufw:
+        state: enabled
+        policy: deny
+
+    - name: Add RPC public keys to authorized_keys
```
hughsaunders: This fails here, because it tries to add a key to the jenkins user that hasn't been created yet.

odyssey4me:
As an additional measure, it may be worth adding an install of fail2ban. It blocks IP's that do successive failed attempts at connecting via SSH.

playbooks/setup_jenkins_slave.yml
git-harry: Given `jenkins` doesn't appear to use SSH I've removed it from the list.

git-harry:
@odyssey4me the firewall is already restricting access to specific IP addresses so I'm not sure fail2ban adds any additional protection.

hughsaunders:
Build using 77e4653:
  https://rpc.jenkins.cit.rackspace.net/view/AIO/job/RPC-AIO_master-swift-periodic/98/console

hughsaunders:
Next test build: https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_master-swift-periodic/99/console

git-harry:
This seems to break something, I'll need to investigate further. https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_master-swift-periodic/101

playbooks/setup_jenkins_slave.yml
```diff
@@ -2,6 +2,77 @@
...
+      authorized_key: "user={{ item }} key='{{ lookup('file', lookup('env', 'WORKSPACE')+'/rpc-gating/keys/rcb.keys') }}' state=present"
```
evrardjp: probably worth defaulting the lookup env.
Also, please use yaml notation. Far more readable.

evrardjp:
We want to ensure the ssh is what we need, not only pick a few lines, so I think it's a generally good addition. I didn't go deeper into details though.

git-harry:
recheck_lint

git-harry:
recheck_lint

git-harry:
@mattt416 this is ready for review.

playbooks/webhooks.yml
```diff
@@ -35,6 +19,72 @@
...
+      AcceptEnv: "LANG LC_*"
+      Subsystem: "sftp /usr/lib/openssh/sftp-server"
+      UsePAM: "yes"
+      GatewayPorts: "no"
+      ClientAliveInterval: 15
+      UseDNS: "no"
+      Compression: "yes"
+      CompressionLevel: 6
+    - role: jnv.unattended-upgrades
+      unattended_automatic_reboot_time: "00:00"
```
mattt416: Is this necessary when `unattended_automatic_reboot` defaults to `false`?  Also, would we ever want to automatically reboot?

git-harry: Bah, I missed `unattended_automatic_reboot`. The request was for automatic updates of security packages, how would you prefer this was dealt with when a package requires a restart?

playbooks/webhooks.yml
```diff
@@ -35,6 +19,72 @@
...
+      PrintLastLog: "no"
+      TCPKeepAlive: "yes"
+      AcceptEnv: "LANG LC_*"
+      Subsystem: "sftp /usr/lib/openssh/sftp-server"
+      UsePAM: "yes"
+      GatewayPorts: "no"
+      ClientAliveInterval: 15
+      UseDNS: "no"
+      Compression: "yes"
+      CompressionLevel: 6
```
mattt416: A lot of these are defaults -- have you just explicitly set them incase the defaults change in the role?  Side note, I didn't see anything for `CompressionLevel` in the role.

git-harry: I copied this from how it's used in OSA and merged in the existing config being used by the host. I didn't check whether any of them were already defaults set by the role but am happy to strip them out.

playbooks/webhooks.yml
mattt416: I'd say don't enable automatic reboots and we handle reboots manually.  Keep in mind that this is only upgrading packages w/ security updates, so the # of updates should be quite minimal.

playbooks/webhooks.yml
mattt416: I think it's fine to be explicit, but I'd probably just remove the things like `CompressionLevel` that don't exist in the role so we're not carrying around invalid vars.

git-harry:
Sorry about that @mattt416, I was sure I'd tested it. It should be fixed now.

PR #191, opened 2017-05-17
UG-599 Add JIRA issue creation

hughsaunders:
This adds the functionality to create JIRA issues for failed builds
but doesn't integrate it with any jobs, that will be added in a
subsequent PR.

Test job: https://rpc.jenkins.cit.rackspace.net/job/scratchpipeline/176/
Issue created by test job: https://rpc-openstack.atlassian.net/browse/UG-603

PR #192, opened 2017-05-17
UG-599 Add Jira Issue Creation to Multinode AIO

hughsaunders:
Create jira issues instead of github issues when a post-merge mnaio
build fails.

- [x] Depends on #191


evrardjp:
Title is confusing. ``Use JIRA issues instead of Github``, or ``Move to JIRA issues`` is far more explicit.

PR #193, opened 2017-05-18
UG-605 Cleanup docker images and workspaces

hughsaunders:
Additions to periodic cleanup:
  * Jenkins workspaces not modified in the last 7 days
  * Docker objects (images, containers)

### DIFF
--- a/rpc_jobs/periodic_cleanup.yml
+++ b/rpc_jobs/periodic_cleanup.yml
@@ -59,4 +59,10 @@
             }
           }
         }
+        stage("Cleanup Docker Objects"){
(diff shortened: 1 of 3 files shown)


rpc_jobs/periodic_cleanup.yml
```diff
@@ -59,4 +59,10 @@
             }
           }
         }
+        stage("Cleanup Docker Objets"){
```
dankolbrs: Not to nitpick, but missing a `c` in `Objects`

hughsaunders: Good spot, done :)

PR #196, opened 2017-05-19
UG-605 Fix script not found in periodic cleanup

hughsaunders:
Periodic cleanup has the following rough structure:
```
node(){
 clone rpc-gating
 ...
 docker stuffs
 ...
 cleanup shell scripts
}
```

The problem is that the workspace used after "docker stuffs" is not the
same workspace that is used for the initial clone. That means that the
cleanup shell scripts can't be found. As a workaround, this commit
moves the cleanup shell scripts to before the docker step.

### DIFF
--- a/rpc_jobs/periodic_cleanup.yml
+++ b/rpc_jobs/periodic_cleanup.yml
@@ -27,10 +27,18 @@
     dsl: |
       node(){
         deleteDir()
-        stage("Prepare"){
-          dir("rpc-gating") {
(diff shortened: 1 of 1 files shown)


hughsaunders:
As there were problems last time, I have already updated the Periodic-Cleanup job. [Example of a successful run](https://rpc.jenkins.cit.rackspace.net/job/Periodic-Cleanup/931/)

PR #198, opened 2017-05-19
Install gating python packages in isolated mode

mkam:
The packages that are installed for gating purposes should not use the pip configuration set for RPC deployments.

AIO partial build: https://rpc.jenkins.cit.rackspace.net/job/mkam-RPC-AIO/2/console

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -22,10 +22,10 @@ def install_ansible(){
     source .venv/bin/activate
 
     # These pip commands cannot be combined into one.
-    pip install -U six packaging appdirs
-    pip install -U setuptools
-    pip install 'pip==9.0.1'
+    pip install --index-url https://pypi.python.org/simple pip==9.0.1 setuptools==33.1.1
+    pip install --isolated -c rpc-gating/constraints.txt six packaging appdirs
     pip install \
+      --isolated \
       -U \
       -c rpc-gating/constraints.txt \
       -r rpc-gating/requirements.txt


pipeline_steps/common.groovy
```diff
@@ -22,10 +22,11 @@ def install_ansible(){
     source .venv/bin/activate
 
     # These pip commands cannot be combined into one.
-    pip install -U six packaging appdirs
-    pip install -U setuptools
     pip install 'pip==9.0.1'
```
odyssey4me: In OSA we've found it best to update pip and setuptools at the same time:

pip install 'pip==9.0.1 setuptools==33.1.1'

A later version of setuptools causes nonsense.

odyssey4me: You may have to set the index-url here to ensure this comes from pypi though, so:

pip install --index-url https://pypi.python.org/simple 'pip==9.0.1 setuptools==33.1.1'

pipeline_steps/common.groovy
```diff
@@ -22,10 +22,11 @@ def install_ansible(){
...
+    pip --isolated install -U six packaging appdirs
+    pip --isolated install -U setuptools
```
odyssey4me: setuptools should be pinned

pipeline_steps/common.groovy
```diff
@@ -22,10 +22,11 @@ def install_ansible(){
     source .venv/bin/activate
 
     # These pip commands cannot be combined into one.
-    pip install -U six packaging appdirs
-    pip install -U setuptools
     pip install 'pip==9.0.1'
+    pip --isolated install -U six packaging appdirs
```
odyssey4me: this install should use constraints for pinning - not sure why they're not in requirements/constraints?

pipeline_steps/common.groovy
```diff
@@ -22,10 +22,10 @@ def install_ansible(){
     source .venv/bin/activate
 
     # These pip commands cannot be combined into one.
-    pip install -U six packaging appdirs
-    pip install -U setuptools
-    pip install 'pip==9.0.1'
+    pip install --index-url https://pypi.python.org/simple 'pip==9.0.1 setuptools==33.1.1'
```
odyssey4me: this fails when trying it on trusty with the python-pip package installed

so it looks like this will have to do:

-    pip install -U six packaging appdirs		 +    pip install --index-url https://pypi.python.org/simple 'pip==9.0.1 setuptools==33.1.1'
  
            
  Write  Preview

this fails when trying it on trusty with the python-pip package installed

so it looks like this will have to do:

pip install -U --index-url https://pypi.python.org/simple six packaging appdirs
pip install -U --index-url https://pypi.python.org/simple 'setuptools==33.1.1'
pip install --index-url https://pypi.python.org/simple 'pip==9.0.1'
...

odyssey4me: Actually, I was wrong. The issue is the quotes. This works:

pip install --index-url https://pypi.python.org/simple pip==9.0.1 setuptools==33.1.1
pip install --isolated -c rpc-gating/constraints.txt six packaging appdirs

odyssey4me:
I tested this on a trusty host to ensure that we have a worst case scenario:

```
# If pip is not installed, we install it
which pip &> /dev/null || { apt-get update && apt-get install -y python-pip; }

# These packages are required to be on the host for the things in requirements.txt
apt-get install -y libssl-dev libffi-dev python-dev

# These can be added to requirements.txt instead of being installed separately without constraints
for req in "six packaging appdirs"; do
  echo ${req} >> requirements.txt
done

# Note:
# We use --index-url just in case the version of pip we are using
# is too old to understand --isolated. We need to get up to pip>9
# to ensure that it understands constraints.

pip install --index-url https://pypi.python.org/simple pip==9.0.1 setuptools==33.1.1
pip install --isolated -r rpc-gating/requirements -c rpc-gating/constraints.txt
```

odyssey4me:
One step further is for how pip gets there. This is how to get the latest pip without installing python-pip. This can be done inside a venv.

```
CURL_CMD="curl --silent --show-error --retry 5"
OUTPUT_FILE="$(mktmp)"
${CURL_CMD} https://bootstrap.pypa.io/get-pip.py > ${OUTPUT_FILE} ||\
    ${CURL_CMD} https://raw.githubusercontent.com/pypa/get-pip/master/get-pip.py > ${OUTPUT_FILE}
python ${OUTPUT_FILE} --isolated "pip==9.0.1" "setuptools==33.1.1"
```

mkam:
I think the changes for installing pip and the other system packages should be taken care of in a separate PR, especially since there's UG-612 for pulling that installation out of the slave setup playbook.

I tested the changes of just the pip package installs here: https://rpc.jenkins.cit.rackspace.net/job/mkam-RPC-AIO/5/

odyssey4me:
@mkam Agreed. I've added this information into UG-612 as I think that's the right place to handle how pip gets there and what the minimal package set is.

PR #199, opened 2017-05-19
[WIP] UG-520: Add OSA SHA bump PR proposal job

mkam:
Remaining work: 
- [ ] Test the PR is created properly
- [ ] Decide on the cadence this will run (poll on OSA changes, run daily, etc.)
- [ ] If not polling on changes, don't create duplicate PRs if OSA hasn't changed
- [ ] Address how to bump role requirements and SHA variable references
- [ ] Delete workspace at end of run (haven't added it yet for testing purposes)
- [ ] Close older PR if there's a new SHA before the older one has merged?
- [ ] Move job to run in docker instead?

Test run of current changes, minus the create_pull_request: https://rpc.jenkins.cit.rackspace.net/job/mkam-test/36/console

### DIFF
--- a/rpc_jobs/osa_sha_update.yml
+++ b/rpc_jobs/osa_sha_update.yml
@@ -1,0 +1,92 @@
+- job:
+    name: 'OSA-SHA-Update-PR'
+    project-type: workflow
+    logrotate:
+      daysToKeep: 14
+    parameters:
+        - string:
+            name: OSA_REPO
+            default: "https://github.com/openstack/openstack-ansible"
+        - string:
+            name: OSA_BRANCH
+            default: "stable/newton"
+        - string:
+            name: "RPC_REPO"
+            default: "https://github.com/rcbops/rpc-openstack"
+        - string:
+            name: "RPC_BRANCH"
+            default: "master"
+        - rpc_gating_params
+    dsl: |
+      node(){
+        deleteDir()
+        dir("rpc-gating") {
+          git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
+          common = load 'pipeline_steps/common.groovy'
+        }
+
+        def OSA_SHA;
+        def RPC_OSA_SHA;
+        stage('Compare OSA SHA') {
+          dir("rpc-openstack") {
+            common.prepareRpcGit(env.RPC_BRANCH, env.WORKSPACE)
+            OSA_SHA = sh(
+              script: "git ls-remote ${env.OSA_REPO} ${env.OSA_BRANCH} | awk '{ print \$1 }' | tr -d '\n'",
+              returnStdout: true)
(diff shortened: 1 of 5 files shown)


mkam:
The osa-differ is taking a considerably longer time to run on the CIT node than when I was testing it locally. It may be running into the git clone timeout issue we've been seeing.

evrardjp:
@mkam Do you need help on this connectivity issue?

evrardjp:
I'd be in favor, if we have a discussion on the "when", to do it in either once in a "release train schedule" (1 per month), or alternatively (more risky) do it every time there is a new OSA tag to be consumed.

But that's definitely something we can iterate on, as long as we have the tooling.

mattt416:
I'm running osa-differ on my laptop and the initial run is taking an eternity cloning all the repos in ~/.osa-differ.  I image a slow initial run is expected, and then subsequent runs should be much faster.  This is on the jenkins slave, which seems reasonable (since everything had already been cloned):

```
(.venv)-bash-4.1$ time osa-differ ba6ee18960e1d09ef1431431e05882ef023c8dab 3c78cd2d466d4c02e03782ded43eda37566639c7 -d /var/lib/jenkins/workspace/mkam-test@2/osa-differ --file osa_diff.md

Report written to file: osa_diff.md

real	0m13.871s
user	0m9.811s
sys	0m3.969s
(.venv)-bash-4.1$
```

As such, we probably do not want to wipe the workspace on each build, unless we want to have osa-differ use the default ~/.osa-differ as its output directory?

rpc_jobs/osa_sha_update.yml
```diff
@@ -0,0 +1,92 @@
...
+        - string:
+            name: "RPC_REPO"
+            default: "https://github.com/rcbops/rpc-openstack"
+        - string:
+            name: "RPC_BRANCH"
+            default: "master"
+        - rpc_gating_params
+    dsl: |
+      node(){
+        deleteDir()
```
mattt416: So this here is problematic -- I guess we'll need to remove this or have osa-differ just write outside of the workspace?

rpc_jobs/osa_sha_update.yml
```diff
@@ -0,0 +1,92 @@
...
+              source /opt/rh/python27/enable
+              virtualenv --python=/opt/rh/python27/root/usr/bin/python .venv
+            fi
+            source .venv/bin/activate
+            pip install osa-differ
+            osa-differ "${RPC_OSA_SHA}" "${OSA_SHA}" -d "${env.WORKSPACE}/osa-differ" --file osa_diff.md
+          """
+          osa_diff = sh(
+            script: "cat osa_diff.md",
+            returnStdout: true)
```
mattt416: Did you want to print out `osa_diff`?  I modified the job to do so but the job's [console output](https://rpc.jenkins.cit.rackspace.net/job/mkam-test/38/console) is quite clunky.

PR #201, opened 2017-05-22
UG-613 Change TMPDIR on pip installs

dankolbrs:
Changes TMPDIR env var to use /root rather than pubcloud tmp
to resolve out of space issues on pip installs

Connects https://rpc-openstack.atlassian.net/browse/UG-613


dankolbrs:
Closing this out after cleaning up `/tmp` space with the fix pushed out on Friday seems to have resolved the issue.

PR #203, opened 2017-05-23
Update artifact/AIO jobs for newton-14.1

odyssey4me:
This adds the artifact build tests to the
master and newton-14.1 branches and cleans
up the references to the old artifacts
branch.

The AIO image is set to one that will never
change to ensure that the apt artifacts are
always newer than the existing packages on
the host which is being used.

As these images now contain the ubuntu-updates
repo, the steps to configure the updates/artifacts
apt repo are no longer necessary and have been
removed.

Connects https://github.com/rcbops/u-suk-dev/issues/1735


odyssey4me:
recheck_it_all

mattt416:
This looks OK to me, but will need to be tested before merging since it will affect all RPC AIO jobs.

Also, do we need to test both newton-14.0 and newton-14.1?

odyssey4me:
@mattt416 Yeah, until we do an initial release of newton-14.1 we cannot retire the newton-14.0 branch.

rpc_jobs/rpc_aio.yml
```diff
@@ -184,17 +184,6 @@
...
             }}
-            // We need to checkout the rpc-openstack repo on the CIT Slave
-            // so that we can cater for builds on the more restricted images
-            // used for artifact-based builds.
-            if (env.STAGES.contains("Upgrade")) {{
-              common.prepareRpcGit(env.UPGRADE_FROM_REF, env.WORKSPACE)
-            }} else {{
-              common.prepareRpcGit("auto", env.WORKSPACE)
-            }} // if
-            if(common.is_doc_update_pr("${{env.WORKSPACE}}/rpc-openstack")){{
```
git-harry: Why is this being removed? This breaks the doc update shortcut.

odyssey4me: Ah, I see that now. I was effectively trying to revert https://github.com/rcbops/rpc-gating/commit/253cff09c8ac2601c5203b8a8a246cbf75442987 but I see now that the use of the CIT clone has evolved for this job.

mattt416:
Testing this here: https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_newton141-xenial-periodic/1/

Side note, when this merges the existing newton jobs are going to be re-created (as newton140) so we'll want to go and manually remove the old jobs to conserve space.

odyssey4me:
it looks like https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_newton141-xenial-periodic/2/console failed because of some magic in the gating that's trying to derive and match branch names in another repo. It appears that https://github.com/rcbops-qe/kibana-selenium will require a newton-14.1 branch, and perhaps something else - the second failure isn't clear. @dankolbrs can you assist with making the new branch for that repo?

mattt416:
Triggered https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_newton141-xenial-periodic/4/ since kibana-selenium now has a neuton-14.1 branch.

rpc_jobs/rpc_artifact_build.yml
```diff
@@ -124,16 +127,12 @@
...
-        // We need to checkout the rpc-openstack repo on the CIT Slave
-        // so that we can cater for builds on the more restricted images
-        // used for artifact-based builds.
         dir("rpc-gating") {{
           git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
           common = load 'pipeline_steps/common.groovy'
           pubcloud = load 'pipeline_steps/pubcloud.groovy'
           artifact_build = load 'pipeline_steps/artifact_build.groovy'
         }}
-        common.prepareRpcGit("auto", env.WORKSPACE)
```
evrardjp: why is this removed?

odyssey4me: It's no longer needed. It was only there for the repo changes which have been removed.

evrardjp:
Jobs worked.

PR #204, opened 2017-05-23
UG-615 Disable cinder-backup

dankolbrs:
Disables cinder-backup on AIO as L->M has upstream bug that is
causing gating failures.

https://github.com/rcbops/rpc-openstack/issues/2258

Connects https://rpc-openstack.atlassian.net/browse/UG-615

### DIFF
--- a/playbooks/vars/aio.yml
+++ b/playbooks/vars/aio.yml
@@ -30,4 +30,6 @@ gating_overrides:
   raw_multi_journal: false
   journal_size: 1024
   osd_directory: true
-  cinder_service_backup_program_enabled: true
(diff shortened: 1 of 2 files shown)


alextricity25:
I ran a few mitaka+upgrade jobs with ``RPC_GATING_BRANCH`` set to ``UG-615`` and ``TEMPEST_TEST_SETS`` set to ``scenario defcore``.
The job numbers are 79-81: https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_mitaka-upgrade-periodic/
I say that if those tests pass, then we should be good to go.

PR #205, opened 2017-05-24
UG-616 Make kibana-selenium repo a parameter

mattt416:


mattt416:
Testing this here: https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_mitaka-ceph-periodic/68/

pipeline_steps/kibana.groovy
```diff
@@ -15,7 +15,7 @@ def kibana(branch, vm=null){
 
 def kibana_prep(branch){
   dir("kibana-selenium") {
-    git url: "https://github.com/rcbops-qe/kibana-selenium.git", branch: "${branch}"
+    git url: env.KIBANA_SELENIUM_REPO, branch: "${branch}"
```
git-harry: Why does this needs parameterising? What is this blocking? What consideration was given to parameterising the branch used?

mattt416: This _was_ blocking https://github.com/rcbops/rpc-gating/pull/203 since https://github.com/rcbops-qe/kibana-selenium didn't have a branch for newton-14.1.  Since we weren't sure how long it'd take to get that branch cut, we thought it'd make sense to make https://github.com/rcbops-qe/kibana-selenium configurable so that we could fork and branch ourselves.

Anyway, this is no longer blocked so there's no urgency in getting this merged.

git-harry: The ability to test implies to me the ability to be able to modify the branch as well as the repo.

It appears the real issue here was that kibana-selenium was out of sync with rpc-openstack, how have processes been improved to prevent this in the future?

The code itself seems fine however I from what I understand of this issue I suspect it will prove to be of limited use. I am happy to see it merged but would first like to understand how the root cause of this issue has been addressed.

mattt416:
recheck_lint

mattt416:
recheck_lint

mattt416:
recheck_lint

mattt416:
recheck_lint

pipeline_steps/kibana.groovy
mattt416: Updated to allow branch to be specified as a param, build to test: https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_mitaka-ceph-periodic/73/

@alextricity25 do you have a runbook of sorts when cutting a new branch on rpc-openstack?  If so, can you update it to ensure that prior to cutting a release that https://github.com/rcbops-qe/kibana-selenium gets a new branch cut as well?

hughsaunders: We should create a job for that, then anything else that needs to happen at the same time can be added to it.

PR #206, opened 2017-05-24
 [DO NOT MERGE]Prepare the jobs built off master for Ocata

alextricity25:
Ocata doesn't support trusty, and therefore all trusty jobs can be
excluded from the master+<context> matrix. This will leave only
the Xenial job to run off the master branch, which deploys swift.

NOTE: There is still work to be done to figure out how we are going
to be testing external Ceph deployments Ocata onwards. That work is
being tracked here:
https://github.com/rcbops/u-suk-dev/issues/1640

Connects https://github.com/rcbops/u-suk-dev/issues/1639

### DIFF
--- a/rpc_jobs/lem_multi_node_aio.yml
+++ b/rpc_jobs/lem_multi_node_aio.yml
@@ -27,6 +27,9 @@
       # Xenial builds are run for newton and above.
       - series: mitaka
         context: xenial
+      # The master+trusty must be exlucded in preperation for Ocata
+      - series: master
+        context: trusty
     jobs:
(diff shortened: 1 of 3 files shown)


mattt416:
👍 for @odyssey4me's suggestion.  I think having an `image` axis is probably the way forward.

odyssey4me:
@alextricity25 I've implemented a broader, more flexible test matrix in https://github.com/rcbops/rpc-gating/pull/208 - adding the ocata series will be much simpler if that merges.

alextricity25:
@odyssey4me Sounds good! Closing this in favor of https://github.com/rcbops/rpc-gating/pull/208

PR #207, opened 2017-05-24
Add Leapfrog upgrade AIO jobs

evrardjp:
In order to test the leapfrog upgrades from
kilo to newton, jobs have been added to allow
the deployment of kilo and the leapfrog
upgrade to the newton-14.1 branch.

The job for the kilo deployment is to test
any new PR's to Kilo, if need be, and also
to validate via a periodic job that Kilo can
actually deploy. This will help to differentiate
between failures due to a kilo deployment, or
failures due to the leapfrog upgrade process.

The stages in the leapupgrade jobs are
deliberately reduced to a minimum set. This
is a starting target - once the job is showing
success more stages will be added to find
other potential bugs.

### DIFF
--- a/pipeline_steps/aio_prepare.groovy
+++ b/pipeline_steps/aio_prepare.groovy
@@ -2,7 +2,7 @@ def prepare(){
   common.conditionalStage(
     stage_name: "Prepare Deployment",
     stage: {
-      if (env.STAGES.contains("Upgrade")) {
+      if (env.STAGES.contains("Major Upgrade") || env.STAGES.contains("Leapfrog Upgrade")) {
         common.prepareRpcGit(env.UPGRADE_FROM_REF)
       } else {
         common.prepareRpcGit()
--- a/pipeline_steps/deploy.groovy
+++ b/pipeline_steps/deploy.groovy
@@ -34,7 +34,7 @@ def deploy_sh(Map args) {
             for ( e in environment_vars ) {
               export_vars += "export ${e}; "
             }
-            sh """#!/bin/bash 
+            sh """#!/bin/bash
             sudo ssh -T -oStrictHostKeyChecking=no ${args.vm} \
               '${export_vars} cd /opt/rpc-openstack; scripts/deploy.sh'
             """
@@ -45,11 +45,11 @@ def deploy_sh(Map args) {
   ) // conditionalStage
 }
 
-def upgrade(Map args) {
+def upgrade(String stage_name, String upgrade_script, List env_vars) {
   common.conditionalStage(
-    stage_name: "Upgrade",
+    stage_name: stage_name,
     stage: {
-      environment_vars = args.environment_vars + common.get_deploy_script_env()
+      environment_vars = env_vars + common.get_deploy_script_env()
       withEnv(environment_vars){
         dir("/opt/rpc-openstack/openstack-ansible"){
           sh "git reset --hard"
@@ -57,12 +57,20 @@ def upgrade(Map args) {
         common.prepareRpcGit()
         dir("/opt/rpc-openstack"){
           sh """
-            scripts/test-upgrade.sh
+            scripts/$upgrade_script
           """
         } // dir
       } // withEnv
     } // stage
   ) // conditionalStage
+}
+
+def upgrade_major(Map args) {
+  upgrade("Major Upgrade", "test-upgrade.sh", args.environment_vars)
+}
+
+def upgrade_leapfrog(Map args) {
+  upgrade("Leapfrog Upgrade", "leapfrog/ubuntu14-leapfrog.sh", args.environment_vars)
 }
 
 def addChecksumRule(){
--- a/rpc_jobs/rpc_aio.yml
+++ b/rpc_jobs/rpc_aio.yml
@@ -3,6 +3,9 @@
     # Note: branch is the branch for periodics to build
     #       branches is the branch pattern to match for PR Jobs.
     series:
+      - kilo:
+          branch: kilo
+          branches: "kilo.*"
       - liberty:
           branch: liberty-12.2
           branches: "liberty-.*"
@@ -30,7 +33,7 @@
                     default_volume_type: ceph
             cinder_service_backup_driver: cinder.backup.drivers.ceph
             tempest_service_available_swift: false
-      - upgrade:
+      - majorupgrade:
           STAGES: >-
             Allocate Resources,
             Connect Slave,
@@ -43,10 +46,20 @@
             Prepare Kibana Selenium,
             Kibana Tests,
             Holland,
-            Upgrade,
+            Major Upgrade,
             Cleanup,
             Destroy Slave
           UPGRADE_FROM_REF: "liberty-12.2"
+      - leapfrogupgrade:
+          STAGES: >-
+            Allocate Resources,
+            Connect Slave,
+            Prepare Deployment,
+            Deploy RPC w/ Script,
+            Leapfrog Upgrade,
+            Cleanup,
+            Destroy Slave
+          UPGRADE_FROM_REF: "kilo"
       - xenial:
           IMAGE: "Ubuntu 16.04.2 LTS prepared for RPC deployment"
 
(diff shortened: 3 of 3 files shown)


evrardjp:
Lots of work to be done there. Different approaches can be taken, I took the simplest one, why doesn't touch common libs or other jobs.

mattt416:
recheck_lint

rpc_jobs/rpc_leapfrog.yml
```diff
@@ -0,0 +1,194 @@
...
+                environment_vars = [
+                  "DEPLOY_HAPROXY=yes",
+                  "DEPLOY_AIO=no",
+                  "DEPLOY_MAAS=no",
+                  "DEPLOY_TEMPEST=no",
+                  "DEPLOY_SWIFT=${{DEPLOY_SWIFT}}",
+                  "DEPLOY_CEPH=${{DEPLOY_CEPH}}",
+                  "DEPLOY_ELK=${{DEPLOY_ELK}}",
+                  ]
+                # TODO(evrardjp) add other leapfrog vars to alter behavior
```
mattt416: Hey @evrardjp, I'll review this more closely once it's passing the lint job -- see http://groovy-lang.org/syntax.html#_single_line_comment as `#` is not valid for comments.

rpc_jobs/rpc_leapfrog.yml
```diff
@@ -0,0 +1,194 @@
...
+      - pr:
+          CRON: ""
+          TRIGGER_USER_VARS: "maas_use_api: false"
+      - periodic:
+          branches: "do_not_build_on_pr"
+          NUM_TO_KEEP: 10
+    jobs:
+      - 'RPC-Leapfrog_{series}-{image}-{context}-{ztrigger}'
+
+- job-template:
```
mattt416: We need a default for `NUM_TO_KEEP`, ie. https://github.com/rcbops/rpc-gating/blob/master/rpc_jobs/rpc_aio.yml#L103.

evrardjp:
I was talking about the general approach of copy pasting the aio and just adapting what I need vs doing a clean job that does the bare minimum.
Jesse will submit a patch on this.

odyssey4me:
I've implemented a new stage for the existing AIO jobs, and added the kilo series back with (hopefully) complete exceptions to prevent unexpected jobs.

evrardjp:
Decent temp fix, not a long term solution, which is exactly what I am looking for.

pipeline_steps/aio_prepare.groovy
```diff
@@ -2,7 +2,7 @@ def prepare(){
   common.conditionalStage(
     stage_name: "Prepare Deployment",
     stage: {
-      if (env.STAGES.contains("Upgrade")) {
+      if ((env.STAGES.contains("Major Upgrade")) || (env.STAGES.contains("Leap Upgrade"))) {
```
git-harry: Nit: remove the unnecessary brackets:
```yaml
     if (env.STAGES.contains("Major Upgrade") || env.STAGES.contains("Leap Upgrade")) {
```

pipeline_steps/deploy.groovy
```diff
@@ -65,6 +65,26 @@ def upgrade(Map args) {
   ) // conditionalStage
 }
 
+def upgrade_leap(Map args) {
```
git-harry: This creates a lot of code duplication, what about something like:
```groovy
void upgrade(String stage_name, String upgrade_script, List env_vars) {
  common.conditionalStage(
    stage_name: stage_name,
    stage: {
      environment_vars = env_vars + common.get_deploy_script_env()
      withEnv(environment_vars){
        dir("/opt/rpc-openstack/openstack-ansible"){
          sh "git reset --hard"
        }   
        common.prepareRpcGit()
        dir("/opt/rpc-openstack"){
          sh """ 
            scripts/$upgrade_script
          """
        } // dir
      } // withEnv
    } // stage
  ) // conditionalStage
}

void upgrade_major(Map args) {
  upgrade("Major Upgrade", "test-upgrade.sh", args.environment_vars)
}

void upgrade_leapfrog(Map args) {
  upgrade("Leapfrog Upgrade", "leapfrog/ubuntu14-leapfrog.sh", args.environment_vars)
}
```

If the code needs to diverge in the future, we can always look to separate the implementations. For now I think this will be easier to manage.

pipeline_steps/aio_prepare.groovy
git-harry: I'd prefer to see `leapfrog` used throughout these changes instead of `leap`, there's no need the scrimp on characters and I'd rather the name was consistent.

rpc_jobs/rpc_aio.yml
```diff
@@ -43,10 +46,22 @@
...
             Destroy Slave
           UPGRADE_FROM_REF: "liberty-12.2"
+      - leapupgrade:
+          STAGES: >-
+            Allocate Resources,
+            Connect Slave,
+            Prepare Deployment,
+            Deploy RPC w/ Script,
+            Setup MaaS,
+            Verify MaaS,
```
git-harry: Why haven't the tempest tests etc been included in the stages?

rpc_jobs/rpc_aio.yml
```diff
@@ -211,15 +245,18 @@
...
+                if (env.STAGES.contains("Major Upgrade")) {{
+                  deploy.upgrade_major(environment_vars: environment_vars)
                   deploy.addChecksumRule()
                   maas.deploy()
                   maas.verify()
                   tempest.tempest()
                   kibana.kibana(env.RPC_BRANCH)
                   holland.holland()
                 }}
+                if (env.STAGES.contains("Leap Upgrade")) {{
```
git-harry: Given these are designed to be run in separate jobs, lets have the code ensure it.
```yaml
else if (env.STAGES.contains("Leap Upgrade")) {{
```

rpc_jobs/rpc_aio.yml
```diff
@@ -3,6 +3,9 @@
     # Note: branch is the branch for periodics to build
     #       branches is the branch pattern to match for PR Jobs.
     series:
+      - kilo:
```
git-harry: Are you looking to add standalone kilo tests, separate from upgrades? I'm not clear from the commit message.

rpc_jobs/rpc_aio.yml
```diff
@@ -61,19 +76,37 @@
...
+        context: leapupgrade
+      - series: liberty
+        context: leapupgrade
+      - series: mitaka
+        context: leapupgrade
+      - series: newton140
+        context: leapupgrade
+      - series: master
+        context: leapupgrade
+      # Ceph builds are not run for kilo
```
git-harry: This comment doesn't add anything. Why aren't they? I believe there is Ceph support in kilo.

pipeline_steps/aio_prepare.groovy
evrardjp: Agreed.

evrardjp: Agreed

pipeline_steps/deploy.groovy
evrardjp: Well I forgot we had this script. Maybe it's worth re-using the script directly and patching it to start the leapfrog. Let's rollback this.

rpc_jobs/rpc_aio.yml
evrardjp: We'd like to start with a minimum system leapfrog, and increase over time. Verify Maas would probably be removed too. We can then add the rest.

rpc_jobs/rpc_aio.yml
evrardjp: Agreed. At some point we'll maybe even be able to re-use existing upgrade stages.

git-harry: Actually maybe this should have something along the lines of 
```yaml
if (env.STAGES.contains("Leap Upgrade") && env.STAGES.contains("Major Upgrade")) {{
  BOOM!!!!
}}

odyssey4me: @git-harry Perhaps, but the job will fail anyway if that happens - so it seems a little pointless to try and prevent job misconfiguration when jobs should not be configured by hand in the first place.

rpc_jobs/rpc_aio.yml
odyssey4me: Added some explanation in the commit message.

PR #208, opened 2017-05-25
Change AIO job axes for an increased test matrix

odyssey4me:
In order to cater for a more complex test matrix,
the AIO job axes have been updated as follows.

series:
The branch information to inform the git checkout
executed by jenkins, eg: master.

image:
The image to execute the job on, eg: trusty.

scenario:
The scenario to configure when executing the job,
eg: swift, ceph.

action:
The action to execute, eg: deploy, minor upgrade,
major upgrade, leapfrog.

In order to prepare for the master branch
becoming Ocata, the trusty builds are excluded
for the branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,6 @@
 - Use `_` between a job name and the [job template variables](https://docs.openstack.org/infra/jenkins-job-builder/definition.html#job-template)
 - Use standard capitalization rules, template variables can be an exception to this
 - Examples:
-  - `RPC-AIO_{series}-{context}-{ztrigger}`
-  - `RPC-AIO_master-swift-periodic`
+  - `RPC-AIO_{series}-{image}-{action}-{scenario}-{ztrigger}`
+  - `RPC-AIO_master-xenial-deploy-swift-periodic`
   - `Merge-Trigger-JJB`
--- a/rpc_jobs/rpc_aio.yml
+++ b/rpc_jobs/rpc_aio.yml
@@ -23,27 +23,24 @@
       - mitaka:
           branch: mitaka-13.1
           branches: "mitaka-.*"
+          UPGRADE_FROM_REF: "liberty-12.2"
       - newton140:
           branch: newton-14.0
           branches: "newton-14.0.*"
       - newton141:
           branch: newton-14.1
           branches: "newton-14.1.*"
+          UPGRADE_FROM_REF: "kilo"
       - master:
           branch: master
           branches: "master"
-    context:
-      - swift
-      - ceph:
-          DEPLOY_SWIFT: "no"
-          DEPLOY_CEPH: "yes"
-          CONTEXT_USER_VARS: |
-            ceph_stable_release: "hammer"
-            cinder_cinder_conf_overrides:
-                DEFAULT:
-                    default_volume_type: ceph
-            cinder_service_backup_driver: cinder.backup.drivers.ceph
-            tempest_service_available_swift: false
+    image:
+      - trusty:
+          IMAGE: "Ubuntu 14.04.5 LTS prepared for RPC deployment"
+      - xenial:
+          IMAGE: "Ubuntu 16.04.2 LTS prepared for RPC deployment"
+    action:
+      - deploy
       - majorupgrade:
           STAGES: >-
             Allocate Resources,
@@ -60,7 +57,6 @@
             Major Upgrade,
             Cleanup,
             Destroy Slave
-          UPGRADE_FROM_REF: "liberty-12.2"
       - leapfrogupgrade:
           STAGES: >-
             Allocate Resources,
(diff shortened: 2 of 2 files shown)


mattt416:
This will need to be tested before we can merge.

rpc_jobs/rpc_aio.yml
```diff
@@ -30,7 +35,9 @@
                     default_volume_type: ceph
             cinder_service_backup_driver: cinder.backup.drivers.ceph
             tempest_service_available_swift: false
-      - upgrade:
+    action:
+      - deploy
+      - majorupgrade:
```
evrardjp: we could have kept ``upgrade`` to be non-intrusive, and use (minor) update, leapfrog for the next ones... But we are changing the job names anyway, so I don't think it really matters.

rpc_jobs/rpc_aio.yml
```diff
@@ -47,8 +54,6 @@
             Cleanup,
             Destroy Slave
           UPGRADE_FROM_REF: "liberty-12.2"
```
evrardjp: It makes sense to me to have UPGRADE_FROM_REF in the series.
This way you could have different values for different major upgrades between N and N+1

We could then use a similar var, LEAP_UPGRADE_FROM_REF in the series for leaping too.

rpc_jobs/rpc_aio.yml
```diff
@@ -61,27 +66,23 @@
...
+        image: xenial
+      # Ceph builds are only run for trusty
+      - image: xenial
+        scenario: ceph
     jobs:
-      - 'RPC-AIO_{series}-{context}-{ztrigger}'
+      - 'RPC-AIO_{series}-{image}-{scenario}-{action}-{ztrigger}'
 
 - job-template:
     # DEFAULTS
```
evrardjp: If we are changing this, I'd say stages should be part of the action. Every different action will have different stages. 

Or alternatively, no job should have different stages, and we handle this in a gating script that's called.

README.md
```diff
@@ -30,6 +30,6 @@
 - Use `_` between a job name and the [job template variables](https://docs.openstack.org/infra/jenkins-job-builder/definition.html#job-template)
 - Use standard capitalization rules, template variables can be an exception to this
 - Examples:
-  - `RPC-AIO_{series}-{context}-{ztrigger}`
-  - `RPC-AIO_master-swift-periodic`
-  - `Merge-Trigger-JJB`
\ No newline at end of file
+  - `RPC-AIO_{series}-{image}-{scenario}-{action}-{ztrigger}`
```
evrardjp: I'd be happy to either:
- Not make this too complex, and use this formatting for only greenfields (make upgrade and leapfrog a different beast)
- Make this more complete, and extend this kind of scenario for upgrades/leapfrog by adding a "{from_version}", which would increase the from->to matrix, necessary for leapfrogs next month.

README.md
```diff
@@ -30,6 +30,6 @@
...
+  - `RPC-AIO_master-xenial-swift-deploy-periodic`
```
evrardjp: The more complete approach would have for example:
- `RPC-AIO_master-xenial-swift-deploy-from_nothing-periodic`
- `RPC-AIO_master-trusty-swift-upgrade-from_14.1.0-periodic`
- `RPC-AIO_newton14.1-trusty-swift-leapfrog-from_kilo-periodic`
- `RPC-AIO_newton14.1-trusty-swift-leapfrog-from_liberty-periodic`

Adding an axis for the ``upgrade from`` version makes sense to me.

evrardjp: In that case, the upgrade vars could be set directly in the {from_version}

odyssey4me: Well, that only applies to upgrade jobs. What about greenfield jobs - what would their name be?

odyssey4me: It seems simpler to me to then just add more series, for example:

RPC-AIO_kilo2newton141-trusty-swift-leapfrog-periodic
RPC-AIO_liberty2newton141-trusty-swift-leapfrog-periodic

evrardjp: That's what I did in my first commit. The world goes round :)

odyssey4me: It may make sense, then to have a set of upgrade jobs that specifically cater to upgrades and not deployments. This will end up duplicating a lot in the current job structure, which is not great and I'd prefer to avoid it... Once we've cleared all the logic from the gating repo and put it into the rpc-openstack repositories then I think it'll be easier. But in the short term the most maintainable solution appears to be bastardising the series. :/

rpc_jobs/rpc_aio.yml
mattt416: If we need to cater for more major upgrades then moving this into series seems to make sense.

odyssey4me: Yes, agreed - however I'm trying to make minimal changes in this patch. As upgrades are not executed for any branch but mitaka at this stage, these changes can be made in a patch later which enables upgrades for other branches.

rpc_jobs/rpc_aio.yml
odyssey4me: Sure, but in this patch I'm trying to make a minimal change set to deliver exactly the same results. Subsequent patches can change things as necessary.

evrardjp:
I still like it.

PR #210, opened 2017-05-31
UG-613

dankolbrs:
Changes TMPDIR env var to use /root rather than pubcloud tmp
to resolve out of space issues on pip installs

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -21,6 +21,9 @@ def install_ansible(){
     cp -r /usr/lib64/python2.6/site-packages/selinux .venv/lib64/python2.7/site-packages/ ||:
     source .venv/bin/activate
 
+    # UG-613 change TMPDIR to directory with more space
+    export PREV_TMPDIR=\$TMPDIR
+    export TMPDIR="/var/lib/jenkins/tmp"
     # These pip commands cannot be combined into one.
     pip install --index-url https://pypi.python.org/simple pip==9.0.1 setuptools==33.1.1
     pip install --isolated -c rpc-gating/constraints.txt six packaging appdirs
@@ -30,6 +33,7 @@ def install_ansible(){
       -c rpc-gating/constraints.txt \
       -r rpc-gating/requirements.txt
 
+    export TMPDIR=\$PREV_TMPDIR
     mkdir -p rpc-gating/playbooks/roles
     ansible-galaxy install -r rpc-gating/role_requirements.yml -p rpc-gating/playbooks/roles
   """
--- a/rpc_jobs/periodic_cleanup.yml
+++ b/rpc_jobs/periodic_cleanup.yml
@@ -35,6 +35,9 @@
           stage("Cleanup Jenkins Workspaces"){
             sh "scripts/workspace_cleanup.sh"
           }
+          stage("Cleanup Jenkins TMP Dir"){
+            sh "scripts/tmp_cleanup.sh"
+          }
           stage("Cleanup Docker Objects"){
             sh "scripts/docker_cleanup.sh"
           }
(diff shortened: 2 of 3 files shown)


pipeline_steps/common.groovy
```diff
@@ -21,6 +21,9 @@ def install_ansible(){
     cp -r /usr/lib64/python2.6/site-packages/selinux .venv/lib64/python2.7/site-packages/ ||:
     source .venv/bin/activate
 
+    # UG-613 change TMPDIR to directory with more space
+    export PREV_TMPDIR=\$TMPDIR
+    export TMPDIR="/var/tmp"
```
hughsaunders: I'd go with /var/lib/jenkins/tmp as thats the largest partition. Probably needs creating though :/

hughsaunders:
```
[root@rpc-jenkins-n02 tmp]# df -h
Filesystem            Size  Used Avail Use% Mounted on
/dev/mapper/sysvg-root
                       20G  2.6G   16G  15% /
tmpfs                 3.9G     0  3.9G   0% /dev/shm
/dev/vda1             976M   66M  860M   8% /boot
/dev/mapper/sysvg-tmp
                      976M  486M  440M  53% /tmp
/dev/mapper/sysvg-var
                      7.6G  2.1G  5.2G  28% /var
/dev/mapper/sysvg-varlibjenkins
                       29G  2.4G   25G   9% /var/lib/jenkins
/dev/mapper/sysvg-varlog
                      7.6G  125M  7.1G   2% /var/log
/dev/mapper/sysvg-varlogaudit
                      976M   29M  897M   4% /var/log/audit
```

pipeline_steps/common.groovy
dankolbrs: /var/ is already ~4.5G larger than the partition we're moving from, and `/var/tmp/ ` exists and has the right perms. I'd be concerned that a possible later Jenkins update or puppet run may affect customization to the server and surprise us later.

pipeline_steps/common.groovy
```diff
@@ -21,6 +21,9 @@ def install_ansible(){
     cp -r /usr/lib64/python2.6/site-packages/selinux .venv/lib64/python2.7/site-packages/ ||:
     source .venv/bin/activate
 
+    # UG-613 change TMPDIR to directory with more space
+    export PREV_TMPDIR=\$TMPDIR
```
mkam: I don't think TMPDIR set by default on the slave.

```
[root@rpc-jenkins-n02 ~]# echo $TMPDIR

[root@rpc-jenkins-n02 ~]# su - jenkins
-bash: _have: command not found
-bash-4.1$ echo $TMPDIR

-bash-4.1$
```

hughsaunders:
We should also add a cleanup stage like https://github.com/rcbops/rpc-gating/blob/master/rpc_jobs/periodic_cleanup.yml#L35

pipeline_steps/common.groovy
dankolbrs: It's not from what I've seen either. The PREV_TMPDIR was just for an abundance of caution and to limit TMPDIR changes to just the pip installs, as we don't need to direct all tmp files to this tmp directory

pipeline_steps/common.groovy
hughsaunders: /var also contains the docker devicemapper datafiles which are now limited to about 4GB. so That leaves ~3 for tmp and anything else thats in /var. Probably more than is there currently, but not by much.

pipeline_steps/common.groovy
mkam: Okay, makes sense then!

dankolbrs:
Updated directory. Also created on the build node, using `jenkins:jenkins` permissions rather than `777`
```
[root@rpc-jenkins-n02 jenkins]# pwd
/var/lib/jenkins
[root@rpc-jenkins-n02 jenkins]# ls -lah tmp
total 8.0K
drwxr-sr-x.  2 jenkins jenkins 4.0K May 31 16:07 .
drwxrwsr-x. 13 jenkins jenkins 4.0K May 31 16:07 ..
```

PR #212, opened 2017-06-01
UG-618 Add MaaS entity removal to periodic cleanup

hughsaunders:
While entities tied to cloud instances are removed when those instances
are deleted, entities that are manually created are not. For instance
entities created for multi node aio VMs may not be cleaned up if a.
job is aborted.

This commit adds entity cleanup to the public cloud step of the
periodic cleanup job.

This commit also adds:
 * multi region cloud server cleanup
 * logging

### DIFF
--- a/githooks.sh
+++ b/githooks.sh
@@ -2,7 +2,9 @@
 
 if [ ! -f .git/hooks/pre-commit ]
 then
-  ln lint.sh .git/hooks/pre-commit
+  pushd .git/hooks
+  ln -s ../../lint.sh .git/hooks/pre-commit
+  popd
   echo "lint.sh installed as pre-commit hook"
 else
   echo "pre commit hook already exists, abort"
--- a/lint.sh
+++ b/lint.sh
@@ -78,7 +78,7 @@ check_bash(){
 }
 
 check_python(){
-  flake8 --exclude=.lintvenv . \
+  flake8 --exclude=.lintvenv,webhooktranslator . \
     && echo "Python syntax ok" \
     || { echo "Python syntax fail"; rc=1; }
 }
--- a/scripts/periodic_cleanup.py
+++ b/scripts/periodic_cleanup.py
@@ -4,65 +4,284 @@
 # are in error state or older than AGE_LIMIT hours.
 
 
+from __future__ import print_function
+
+
 import datetime
 import dateutil.parser
 from dateutil.tz import tzutc
+from functools import wraps
+from operator import attrgetter
 import os
 import re
 
+
 import pyrax
 import jenkins_node
+from rackspace_monitoring.providers import get_driver
+from rackspace_monitoring.types import Provider
+import libcloud
 
 
-def get_env_vars():
-    env_vars = {}
-
-    env_vars["age_limit"] = int(os.environ.get("INSTANCE_AGE_LIMIT", 48))
-    env_vars["instance_prefix"] = os.environ.get("INSTANCE_PREFIX")
-    if not env_vars["instance_prefix"]:
-        raise ValueError("INSTANCE_PREFIX must not be empty")
-
-    env_vars["username"] = os.environ["PUBCLOUD_USERNAME"]
-    env_vars["api_key"] = os.environ["PUBCLOUD_API_KEY"]
-    env_vars["region_name"] = os.environ["REGION"]
-
-    return env_vars
+INDENT = 0
+INDENT_STR = 2 * " "
 
 
-def cleanup_instances(cs_client, age_limit, instance_prefix):
-    current_time = datetime.datetime.now(tzutc())
-    max_age = datetime.timedelta(hours=age_limit)
-
-    prefixed_servers = (
-        server for server in cs_client.servers.list()
-        if re.match(instance_prefix, server.name)
+def _indp(message, **kwargs):
+    """ Indent Print """
+    print("{indent}{message}".format(
+            indent=INDENT_STR * INDENT,
+            message=message
+        ),
+        **kwargs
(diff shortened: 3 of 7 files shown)


scripts/periodic_cleanup.py
```diff
@@ -7,62 +7,125 @@
...
+    def cleanup_maas_entities(self):
+        """ Remove maas entities that relate to deleted instances """
+        entities = self.maas.list_entities()
+        for entity in entities:
+            # multi node aio entities are vm.hostname eg infra1.omnantp-52-66d1
+            # standard aio entities don't include a vm eg: ramcp-77-5dcd
+            hostname = entity.label.split('.')[-1]
+            # entities with URIs are tied to cloud servers and cannot be
+            # manually deleted
+            if re.match(self.instance_prefix, hostname) and not entity.uri:
```
mkam: Won't this delete entities of all servers that have the prefix and not just servers that are deleted?

hughsaunders: Good catch, It won't delete any entities that are linked to instances, but it will delete indirectly linked entities (eg entities for MNAIO VMs). 

I'll revise this tomorrow.

mattt416: The multi-node VM entities don't start with `om` though, so you won't actually get a match here on those entities.

mattt416:
@hughsaunders I mentioned the pileup of agent tokens during standup today, but then remembered we can actually tell rpc_maas to use a given agent token instead of creating a new one for each bild.  I've created UG-623 for us to make this change, and with that change in place there won't be any cleanup necessary as far as agent tokens are concerned.

hughsaunders:
@mattt416 Can multiple agents use the same agent token concurrently?

scripts/periodic_cleanup.py
hughsaunders: @mattt416 I added om to the regex. Also have tweaked the test to check if the server exists.

hughsaunders:
recheck_cit_lint

hughsaunders:
recheck_lint

scripts/periodic_cleanup.py
```diff
@@ -4,65 +4,152 @@
...
+        entities = self.maas.list_entities()
+        server_names = (s.name for s in self.get_servers_from_all_regions())
+        for entity in entities:
+            # multi node aio entities are vm.hostname eg infra1.omnantp-52-66d1
+            # standard aio entities don't include a vm eg: ramcp-77-5dcd
+            hostname = entity.label.split('.')[-1]
+            # entities with URIs are tied to cloud servers and cannot be
+            # manually deleted
+            if (re.match(self.instance_prefix, hostname) and not entity.uri
+                    and hostname not in server_names):
```
mattt416: What happens if we deploy MaaS on a lab node?  The entities would get torn down here right?

Would it be better to check the agent `last_connected` value?  Not sure how this is exposed through the library, but from the cli:

```
<Agent: id=swift3.omnanxp-32-3cef, last_connected=1494909106476>
```

If we took that into account, we'd perhaps avoid deleting entities that shouldn't be.

One other small issue here ... we check a cached server list.  Completely unlikely, but something could come online after that list is generated, deploy entities, and then have them torn down because the host isn't in server_names.

hughsaunders: Have added an agent connection time check for entities that have agents. Entities don't seem to have a create time, otherwise I'd add a min age check :/

mattt416:
Apologies @hughsaunders -- I initially said we can override `maas_agent_token` to have all entities use the same agent token rather than create one for each entity.  This would save us having to teardown agent tokens, which is not possible via the ansible module and would therefore need to be done in a cleanup script.

However, after thinking about it more I thought that perhaps we'd be better off exercising that agent token creation code in the rpc_maas by not re-using the same token each time.  This would mean we'd need a new method to clean up agent tokens.  :(  Is that something you can add here or shall I we do that in a follow-up commit?

hughsaunders:
Hey @mattt416 added agent token cleanup in latest push. Regarding testing, I have tested locally except for the agent token cleanup (I tested the script with that line commented out).

hughsaunders:
lint failed on 
```
Error response from daemon: Untar error on re-exec cmd: fork/exec /usr/bin/docker: too many open files
```

hughsaunders:
recheck_lint

hughsaunders:
recheck_lint

githooks.sh
```diff
@@ -2,7 +2,9 @@
 
 if [ ! -f .git/hooks/pre-commit ]
 then
-  ln lint.sh .git/hooks/pre-commit
+  pushd .git/hooks
+  ln -s ../../lint.sh .git/hooks/pre-commit
+  popd
```
mattt416: This seems out of context w/ this PR.

lint.sh
```diff
@@ -78,7 +78,7 @@ check_bash(){
 }
 
 check_python(){
-  flake8 --exclude=.lintvenv . \
+  flake8 --exclude=.lintvenv,webhooktranslator . \
```
mattt416: This seems out of context here also

githooks.sh
hughsaunders: This was required to get lint to pass when switching branches. Previously lint.sh was hard linked to pre-commit so the hook did not reflect the current branch's lint.sh.

lint.sh
hughsaunders: This was also required to get lint to pass, as the webhook translator dir contained build dirs that didn't pass flake8. Its not directly related to this patch but included because it will cause problems if you switch branches.

PR #214, opened 2017-06-01
UG-618 Webhook Translator

hughsaunders:
UG-618 Add MaaS-Jira Webhook Translator

This is a small flask app that receives webhooks from MaaS notifications
and creates jira tickets via the jira API.

While the app is small unfortunately there is quite a lot of python
packaging and deployment code around it.

This commit reuses the nginx that is already deployed to the webhook
proxy node. Nginx receives http requests and passes them on to
webhook translator via a fastcgi socket. The webook translator daemon
is managed as a systemd service and started by socket activation.

Other notes:
* Setuptools' console scripts function is used to create
  startup scripts that are installed into the deployment venv.
  This allows the daemon to run within a venv without systemd
  having to know how to activate a venv.
* get_maas_token_and_url.py has been replaced with maasutils.
  This includes the original functionality and adds the ability
  to set the webhook token for a MaaS account. This token is sent
  with MaaS notifications and verfied by webhook translator to ensure
  that requests that didn't originate from MaaS don't create tickets.
* Webhook translator has some unit tests and these are called from
  tox in lint.sh. Perhaps that should become run-tests.sh or similar in
  future.

---

There are two commits to separate out the addition of error
handling to webhooks.groovy. That commit involved re-indenting
most of that file which would make review harder, so its in it's
own commit. 

---

TODO: 
- [x] Remove webhooktranslator/requirements.txt


hughsaunders:
Still to do: Connect the webhook translator playbook to the job that deploys the webhook instance, and supply the creds.

playbooks/webhooktranslator.yml
```diff
@@ -0,0 +1,40 @@
...
+        dest: "{{RPC_GATING_DIR}}"
+
+    - name: install virualenv tool
+      pip:
+        name: "virtualenv"
+
+    - name: Install webhooktranslator
+      pip:
+        virtualenv: "{{WHT_VENV_DIR}}"
+        name: "{{WHT_DIR}}"
```
evrardjp: virtualenv_python: ?

playbooks/webhooktranslator.yml
```diff
@@ -0,0 +1,40 @@
+- name: Deploy Webhook Translator
+  hosts: job_nodes
+  user: root
+  vars:
+    RPC_GATING_REPO: "http://github.com/rcbops/rpc-gating"
+    RPC_GATING_BRANCH: master
+    RPC_GATING_DIR: /opt/rpc-gating
+    WHT_DIR: "{{RPC_GATING_DIR}}/webhooktranslator"
```
evrardjp: you don't like spaces, do you?

playbooks/webhooktranslator.yml
```diff
@@ -0,0 +1,40 @@
...
+    JIRA_PASSWORD: "{{ansible_env.JIRA_PASSWORD}}"
+    JIRA_PROJECT: "{{ansible_env.JIRA_PROJECT}}"
+  tasks:
+    - name: Clone rpc-gating
+      git:
+        repo: "{{RPC_GATING_REPO}}"
+        version: "{{RPC_GATING_BRANCH}}"
+        dest: "{{RPC_GATING_DIR}}"
+
+    - name: install virualenv tool
```
evrardjp: nit: virualenv -> virtualenv

playbooks/webhooktranslator.yml
```diff
@@ -0,0 +1,40 @@
...
+    - name: Template webhooktranslator service definition
+      template:
+        src: webhooktranslator_service.j2
+        dest: /lib/systemd/system/webhooktranslator.service
+
+    - name: Enable and start webhooktranslator
+      systemd:
+        state: started
+        enabled: yes
+        service: webhooktranslator
```
evrardjp: daemon_reload: yes because you have a new file

webhooktranslator/setup.cfg
```diff
@@ -0,0 +1,25 @@
...
+    Environment :: Web Environment
+    Intended Audience :: Developers
+    Intended Audience :: Information Technology
+    License :: OSI Approved :: Apache Software License
+    Operating System :: OS Independent
+    Programming Language :: Python
+
+[files]
+packages =
+    webhooktranslator
```
evrardjp: I can't say for this, someone else should review.

webhooktranslator/webhooktranslator/webhooktranslator.py
```diff
@@ -0,0 +1,89 @@
...
+        summary="MaaS Alert: {alarm}/{entity}:{state}".format(
+            alarm=alarm,
+            entity=entity,
+            state=state
+        ),
+        description="""
+Alarm {alarm} for entity {entity} is in {state} state.
+The check is {check}.
+
+{{code:title=Full Payload from MaaS}}
```
evrardjp: That's surprising.
Could you document that {{ code }} is jira markup?

hughsaunders:
recheck_lint

playbooks/webhooktranslator.yml
hughsaunders: Added unnecessary spaces for style compliance ;-)

playbooks/webhooktranslator.yml
hughsaunders: Done

playbooks/webhooktranslator.yml
hughsaunders: I _think_ this is ok as is.

playbooks/webhooktranslator.yml
hughsaunders: Done, thanks.

webhooktranslator/webhooktranslator/webhooktranslator.py
hughsaunders: Done

hughsaunders:
I've left my last test deploy up in case its useful for review. 
Deploy Job: https://rpc.jenkins.cit.rackspace.net/job/Webhook-Setup/48/
IP: 104.130.24.98

evrardjp:
great!

webhooktranslator/requirements.txt
```diff
@@ -0,0 +1,3 @@
+jira==1.0.10
```
hughsaunders: This file is not accurate or required. Deps have been inlined into setup.py. I'll remove this after the first round of reviews is complete.

scripts/maasutils.py
```diff
@@ -0,0 +1,58 @@
...
+        api_key
+    )
+
+
+@click.command()
+def get_token_url():
+    token = pyrax.identity.token
+
+    monitoring_service = next(
+        b for b in pyrax.identity.service_catalog if b["type"] == "rax:monitor"
```
evrardjp: How do we ensure that the iterable of pyrax.identity.service_catalog is fine?
Should something special be done if we reach the end of the iterator (raise when exception StopIterator?)

lint.sh
```diff
@@ -78,17 +78,25 @@ check_bash(){
 }
 
 check_python(){
-  flake8 --exclude=.lintvenv . \
+  flake8 --exclude=.lintvenv,webhooktranslator . \
```
hughsaunders: Because it may contain build environments that don't pass flake8, also it handles its own flake8 via tox.

lint.sh
```diff
@@ -78,17 +78,25 @@ check_bash(){
...
     && echo "Python syntax ok" \
     || { echo "Python syntax fail"; rc=1; }
 }
 
+check_webhooktranslator(){
+    pushd webhooktranslator
+    tox && echo "Webhook Translator Unit tests pass" \
+      || { echo "Webhook Translator Unit tests fail"; rc=1; }
+    popd
+}
```
hughsaunders: I considered the same thing, but I'm not sure how you define non python environments with tox?

playbooks/webhooks.yml
```diff
@@ -49,37 +63,104 @@
...
+    JIRA_USER: "{{ lookup('env', 'JIRA_USER') }}"
+    JIRA_PASSWORD: "{{ lookup('env', 'JIRA_PASS') }}"
+    JIRA_PROJECT: "{{ lookup('env', 'JIRA_PROJECT') }}"
+
+  pre_tasks:
+    # used by webhooks.groovy to create the ssh tunnel script
     - name: Write ip file
       delegate_to: localhost
       copy:
         dest: "{{WORKSPACE}}/instance_address"
```
hughsaunders: https://github.com/rcbops/rpc-gating/pull/214/commits/ccc6290ba2ba7193ebdb3d309fd5aab06265e3eb#diff-3a426f9e296ebe82a1aa9ffb6b41c671R42

playbooks/webhooks.yml
```diff
@@ -49,37 +63,104 @@
...
+    - name: Clone rpc-gating
+      git:
+        repo: "{{ RPC_GATING_REPO }}"
+        version: "{{ RPC_GATING_BRANCH }}"
+        dest: "{{ RPC_GATING_DIR }}"
+
+    - name: Install webhooktranslator
+      pip:
+        virtualenv: "{{ WHT_VENV_DIR }}"
+        name: "{{ WHT_DIR }}"
```
hughsaunders: Setuptools sets the execute bit when it creates the console scripts as install time.

scripts/maasutils.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python
+import click
+import pyrax
+import requests
+
+
+@click.option("--username", required=True)
+@click.option("--api-key", required=True)
```
hughsaunders: --help will show the options. While required options is an oxymoron, its not entirely frowned upon. Some people think that arguments should be used for all required params. Other people use options for everything because it means the order of params is not important and gives a name to each parameter that is supplied. I'm in the second camp..

scripts/maasutils.py
hughsaunders: If there are no monitoring endpoints the script will bomb, but thats fine, MaaS isn't going to work without an endpoint. Also this code was ported over from an existing script.

playbooks/templates/webhooktranslator_socket.j2
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Webhook Translator Listen Socket
+
+[Socket]
+SocketUser=www-data
```
hughsaunders: This is designed for a very narrow use case, we can add more flexibility later if required.

playbooks/templates/webhooktranslator_service.j2
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Webhook translator service (MaaS to Jira)
+After=network.target
+
+[Service]
+Environment="RACKSPACE_WEBHOOK_TOKEN={{RACKSPACE_WEBHOOK_TOKEN}}" "JIRA_USER={{JIRA_USER}}" "JIRA_PASSWORD={{JIRA_PASSWORD}}" "JIRA_PROJECT={{JIRA_PROJECT}}"
```
hughsaunders: OK, I'll add that as a TODO.

webhooktranslator/tests/test_webhooktranslator.py
```diff
@@ -0,0 +1,57 @@
...
+        resp = self.app.get(path='/maas',
+                            headers={'x-rackspace-webhook-token': self.token},
+                            content_type="application/json",
+                            data=data)
+        self.assertEqual(resp.status_code, 200)
+        _, kwargs = mock_create_jira_issue.call_args
+        assert "MaaS Alert:" in kwargs['summary']
+        assert "WARNING" in kwargs['summary']
+        assert "Full Payload" in kwargs['description']
+        assert "alert" in kwargs['labels']
```
git-harry: Why are you using assert statements here and not unittest assert methods i.e. `assertIn`?

webhooktranslator/setup.py
```diff
@@ -0,0 +1,22 @@
+from setuptools import setup, find_packages
+
+setup(
+    name='webhooktranslator',
+    description='Receive alerts from MaaS and create jira issues',
+    url='http://github.com/rcbops/rpc-gating/tree/master/webhooktranslator',
```
git-harry: Is there a reason this isn't https?

playbooks/webhooks.yml
evrardjp: I'd rather have this as an env var, it's more explicit than a hidden var in the gating. But fine for me right now.

playbooks/webhooks.yml
evrardjp: great!

webhooktranslator/webhooktranslator/webhooktranslator.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+app = Flask(__name__)
+
+
+def rackspace_webhook_token_required(f):
+    @wraps(f)
+    def decorated_function(*args, **kwargs):
+        token_header = "x-rackspace-webhook-token"
+        token = os.environ.get('RACKSPACE_WEBHOOK_TOKEN', '')
```
hughsaunders: Done

webhooktranslator/webhooktranslator/webhooktranslator.py
```diff
@@ -0,0 +1,99 @@
...
+        if request.headers.get(token_header, 'invalidtoken') != token:
```
hughsaunders: Done

webhooktranslator/webhooktranslator/webhooktranslator.py
```diff
@@ -0,0 +1,99 @@
...
+            return abort(403, "x-rackspace-webhook-token validation failed")
```
hughsaunders: Done

webhooktranslator/webhooktranslator/webhooktranslator.py
```diff
@@ -0,0 +1,99 @@
...
+def json_required(f):
+    @wraps(f)
+    def decorated_function(*args, **kwargs):
+        if not request.is_json:
+            abort(400, "JSON Required")
+        return f(*args, **kwargs)
+    return decorated_function
+
+
+@app.route("/maas", methods=['GET', 'POST'])
```
hughsaunders: Not needed, removed.

webhooktranslator/webhooktranslator/webhooktranslator.py
```diff
@@ -0,0 +1,99 @@
...
+@rackspace_webhook_token_required
+@json_required
+def maas():
+    maas_payload = request.get_json()
+    state = maas_payload['details']['state']
```
hughsaunders: Added exception handler for missing JSON keys.

webhooktranslator/webhooktranslator/webhooktranslator.py
```diff
@@ -0,0 +1,99 @@
...
+    entity = maas_payload['entity']['id']
+    check = maas_payload['check']['id']
+    alarm = maas_payload['alarm']['id']
+
+    create_jira_issue(
+        summary="MaaS Alert: {alarm}/{entity}:{state}".format(
```
hughsaunders: Good point. And the IDs are included in the full payload. Done.

webhooktranslator/webhooktranslator/webhooktranslator.py
```diff
@@ -0,0 +1,99 @@
...
+
+{{code:title=Full Payload from MaaS}}
+{maas_payload}
+{{code}}
+        """.format(
+            alarm=alarm,
+            entity=entity,
+            state=state,
+            check=check,
+            maas_payload=maas_payload
```
hughsaunders: Done

webhooktranslator/webhooktranslator/webhooktranslator.py
```diff
@@ -0,0 +1,99 @@
...
+        ),
+        labels=['maas', 'alert', 'automated', 'jenkins']
+    )
+    return "SUCCESS"
+
+
+def create_jira_issue(summary, description, labels):
+    juser = os.environ['JIRA_USER']
```
hughsaunders: I prefer env vars, as a config file is one more thing to deploy, template update etc. I agree that it would be better to read all the env vars in one place so that its more obvious which env vars are required.  Done, implemented as `read_env()`

webhooktranslator/tests/test_webhooktranslator.py
hughsaunders: Done.

webhooktranslator/setup.py
hughsaunders: Done

playbooks/templates/webhooktranslator_service.j2
hughsaunders: Decided not to do this as some of the variables come from environment variables so make sense to be capitalised, and the others are capitalised for consistency.

hughsaunders:
Added responses for first round of reviews in https://github.com/rcbops/rpc-gating/pull/214/commits/0eb104c8c7512a035ada43477736541298bade08 Haven't got a test instance yet due to long build queue :(

hughsaunders:
recheck_lint

hughsaunders:
@git-harry @evrardjp fixes from first round of reviews are added in a [separate commit](https://github.com/rcbops/rpc-gating/pull/214/commits/0eb104c8c7512a035ada43477736541298bade08) to make re-reviewing easier. Will squash before merge. 

Also ran a new test build:
 - console: https://rpc.jenkins.cit.rackspace.net/job/Webhook-Setup/49/console 
 - ip: 162.242.209.237 
 - sample issue: https://rpc-openstack.atlassian.net/browse/UG-631

scripts/maasutils.py
```diff
@@ -0,0 +1,58 @@
...
+@click.group()
+def cli(api_key, username):
+    pyrax.set_setting("identity_type", "rackspace")
+    pyrax.set_credentials(
+        username,
+        api_key
+    )
```
hughsaunders: The monitoring api doesn't split by regions as the others do. You can view and modify entities that link to instances from any region, regardless of the region specified on connection. As this pyrax connection is only used for maas, specifying the region is superfluous and potentially misleading.

PR #218, opened 2017-06-06
Remove rpc_release from test_holland

odyssey4me:
The value for rpc_release is set via group_vars for
all versions of RPC-O. The playbook should inherit
it from there instead of trying to override it.

This has been working so far due to the rpc_release
override set in user_*.yml files, but as we try to
cut down on those problems like this are likely to
arise.

### DIFF
--- a/playbooks/test_holland.yml
+++ b/playbooks/test_holland.yml
@@ -3,8 +3,12 @@
 - hosts: galera
   user: root
   vars:
-    rpc_release: "{{ lookup('pipe', 'cd /opt/rpc-openstack && git describe --tags --abbrev=0') }}"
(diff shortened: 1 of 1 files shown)


odyssey4me:
Ah @mattt416 good catch - I forgot that there was previously a module used to derive and set the value in each playbook. I'll adjust this patch to only set it if it isn't already set. Meanwhile, I've submitted patches to the older branches to make them behave more consistently:

https://github.com/rcbops/rpc-openstack/pull/2304
https://github.com/rcbops/rpc-openstack/pull/2305
https://github.com/rcbops/rpc-openstack/pull/2306

PR #219, opened 2017-06-06
UG-621 Stage python artifacts on all infra VMs

mkam:
The stage-python-artifacts.yml playbook stages the python artifacts
on localhost by default. For multi-node AIO builds, these artifacts
need to be on all of the infra VMs, which host the repo server containers.


mkam:
recheck_lint

mkam:
Closing in favor of changing the default in stage-python-artifacts.yml.

PR #221, opened 2017-06-07
Remove ANSIBLE_CACHE_PLUGIN from Holland testing

mkam:
This variable caused the rpc_release variable to
be undefined when executing the holland tests
for mitaka.


mkam:
Tested that removing it did not cause the [jsonfile error](https://github.com/rcbops/u-suk-dev/issues/1578#issuecomment-297603765) to return, which is why the variable was added in the first place.

https://rpc.jenkins.cit.rackspace.net/job/mkam-OnMetal-Multi-Node-AIO/44/

PR #223, opened 2017-06-08
Fix execution of tempest on AIO

mattt416:
AIO builds are currently failing with a "No such file or directory"
error when running tempest on the AIOs.  This appears to be a result of
running tempest unwrapped with quotes around the tempest command, which
bash does not like:

```
root@ranxp-16-8323:~# 'cd /tmp && ls -al'
-bash: cd /tmp && ls -al: No such file or directory
root@ranxp-16-8323:~#
```

This commit simply creates a temporary variable depending on whether we
are using a wrapper (and need to quote) or not.

### DIFF
--- a/pipeline_steps/tempest.groovy
+++ b/pipeline_steps/tempest.groovy
@@ -12,9 +12,14 @@ def tempest_install(vm=null){
 }
 
 def tempest_run(wrapper="") {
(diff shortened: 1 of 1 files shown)


mattt416:
Testing this here: https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_mitaka-ceph-periodic/83/console

PR #224, opened 2017-06-08
Removing the hard coded cinder backup option

cloudnull:
This option should be set within the RPC-O project should we need to
have the service disabled. Hard coding it here makes it next to
impossible to test changes that may use this feature within the gate.

### DIFF
--- a/playbooks/vars/aio.yml
+++ b/playbooks/vars/aio.yml
@@ -30,9 +30,6 @@ gating_overrides:
   raw_multi_journal: false
   journal_size: 1024
   osd_directory: true
-  # UG-615 disabling for now due to upstream bug on service restarts
-  # https://github.com/rcbops/rpc-openstack/issues/2258
-  cinder_service_backup_program_enabled: false
   tempest_test_sets: "scenario defcore"
   tempest_run_tempest_opts:
     - "--serial"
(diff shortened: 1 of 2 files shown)


hughsaunders:
Presumably we need to put changes into RPC-O before merging this for example to remove https://github.com/rcbops/rpc-openstack/blob/mitaka-13.1/scripts/deploy.sh#L96 and https://github.com/rcbops/rpc-openstack/blob/liberty-12.2/scripts/deploy.sh#L94 ?

odyssey4me:
@hughsaunders I'd love to know why cinder backup is suddenly being disabled?

odyssey4me:
mitaka: https://github.com/rcbops/rpc-openstack/pull/2315
liberty: https://github.com/rcbops/rpc-openstack/pull/2316

cloudnull:
@hughsaunders it looks like this was already disabled in our variable files "https://github.com/rcbops/rpc-openstack/blob/mitaka-13.1/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml#L150" but reset in the aio script. While I'm curious why we didn't just remove the offending line in the aio script I will put up some reviews in RPC-O to get it fixed.

odyssey4me:
Both the mitaka and liberty branches have their fixes merged, so this should be fine to remove now.

PR #226, opened 2017-06-08
UG-629 Temporary fix for multi-node AIO tempest vars

mkam:
Add quotations and an empty string to tempest vars
for multi-node AIO builds because they are not
surrounded by quotes in the run_tempest.yml playbook.


mkam:
Testing here: https://rpc.jenkins.cit.rackspace.net/view/Multi-Node%20AIO/job/OnMetal-Multi-Node-AIO_newton140-trusty-periodic/5/

PR #228, opened 2017-06-09
UG-619 Add job to deploy long running cloud slave

hughsaunders:
TODO: 
- [x] docker config
- [x] MaaS notification & plan setup
- [x] remove jenkins root privs (granted as standard as part of single use slave creation). 
- [x] security/lockdown - could reuse the work done for the webhook proxy node. Would need to split the security bits from the proxy configuration. 
- [x] check exclusive mode and increase executors

### DIFF
--- a/playbooks/setup_jenkins_slave.yml
+++ b/playbooks/setup_jenkins_slave.yml
@@ -1,8 +1,30 @@
 ---
 - hosts: job_nodes
   user: root
+  vars:
+    # Env lookup returns an empty string when an env var doesn't exist. |default
+    # will normally only replace on unset, not "". If |default(var,true) is
+    # used, then default will replace whenever var evaluates to false, which
+    # isn't helpful for booleans as the user may have specified false. ""|bool
+    # returns false, but these bools should default to true.
+    # A jinja if is used to set a bool default if env lookup returned "".
+    allow_jenkins_sudo: "{% set v=lookup('env', 'allow_jenkins_sudo') %}{% if v=='' %}True{% else %}{{v|bool}}{% endif %}"
+    jenkins_node_exclusive: "{% set v=lookup('env', 'jenkins_node_exclusive') %}{% if v=='' %}True{% else %}{{v|bool}}{% endif %}"
+    # These variables are not booleans so can use the standard default filter
+    jenkins_node_labels: "{{ lookup('env', 'JENKINS_NODE_LABELS'|default('single_use_slave')) }}"
+    jenkins_node_executors: "{{ lookup('env', 'JENKINS_NODE_EXECUTORS'|default('2', true)) }}"
+  roles:
+    - role: willshersystems.sshd
+      sshd:
+        ListenAddress:
+          - "0.0.0.0"
+        PasswordAuthentication: "no"
+        X11Forwarding: "no"
+        PrintLastLog: "no"
+        GatewayPorts: "no"
+        ClientAliveInterval: 15
+        Compression: "yes"
   tasks:
-
     - name: Install apt packages
       apt:
         pkg: "{{ item }}"
@@ -41,6 +63,7 @@
         state: present
         regexp: 'jenkins'
         line: 'jenkins ALL=(ALL) NOPASSWD: ALL'
+      when: "{{ allow_jenkins_sudo }}"
 
     - name: Copy constraints file over to cloud server
       copy:
@@ -106,7 +129,9 @@
           --name {{inventory_hostname}} \
           --ip {{ansible_host}} \
           --creds "SSH Creds for Jenkins instances in public cloud." \
-          --labels single_use_slave
+          --labels "{{ jenkins_node_labels }}" \
+          --executors "{{ jenkins_node_executors }}" \
+          {% if jenkins_node_exclusive %}--exclusive{% endif %}
       register: jenkins_response
       until: jenkins_response|success
       retries: 5
--- a/rpc_jobs/long_running_slave.yml
+++ b/rpc_jobs/long_running_slave.yml
@@ -1,0 +1,96 @@
+- job:
+    name: Long-Running-Slave
+    project-type: workflow
+    concurrent: true
+    properties:
+      - build-discarder:
+          num-to-keep: 30
+    parameters:
+      # See params.yml
+      - rpc_gating_params
+      - instance_params:
+          IMAGE: "Ubuntu 16.04 LTS (Xenial Xerus) (PVHVM)"
+          REGION: "IAD"
+          FLAVOR: general1-8
+      - jenkins_node_params:
+          JENKINS_NODE_LABELS: "pubcloud_multiuse"
+          JENKINS_NODE_EXECUTORS: "16"
+          jenkins_node_exclusive: false
+          allow_jenkins_sudo: false
+      - rpc_maas_params:
+          RPC_MAAS_BRANCH: master
+      - string:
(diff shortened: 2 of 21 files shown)


playbooks/setup_jenkins_slave.yml
```diff
@@ -1,8 +1,23 @@
 ---
 - hosts: job_nodes
   user: root
+  vars:
+    allow_jenkins_sudo: "{{ lookup('env', 'ALLOW_JENKINS_SUDO')|default(true, true)|bool }}"
+    jenkins_node_labels: "{{ lookup('env', 'JENKINS_NODE_LABELS'|default('single_use_slave')) }}"
+    jenkins_node_executors: "{{ lookup('env', 'JENKINS_NODE_EXECUTORS'|default('2')) }}"
+    jenkins_node_exclusive: "{{ lookup('env', 'JENKINS_NODE_EXECLUSIVE'|default('2')) }}"
```
mkam: Typo in EXCLUSIVE.

rpc_jobs/long_running_slave.yml
```diff
@@ -0,0 +1,102 @@
...
+          default: "10"
+          description: |
+            Number of executors to start on each slave
+      - string:
+          name: ALLOW_JENKINS_SUDO
+          default: False
+          description: |
+            Grant the jenkins slave passwordless sudo. This is useful for
+            single use slaves, but must not be set for long running
+            slaves as it will allow jobs to pollute the build environment.
```
mkam: The parameter for exclusive mode is missing.

rpc_jobs/long_running_slave.yml
```diff
@@ -0,0 +1,102 @@
...
+      - rpc_gating_params
+      - single_use_slave_params:
+          IMAGE: "Ubuntu 16.04 LTS (Xenial Xerus) (PVHVM)"
+          REGION: "IAD"
+          FLAVOR: general1-8
+      - rpc_maas_params:
+          RPC_MAAS_BRANCH: master
+      - string:
+          name: JENKINS_NODE_EXECUTORS
+          default: "10"
```
dankolbrs: Is the default 10 executors going to be used when building the long running slave? The flavor being used has quite a bit more storage and CPUs than the current CIT build node and 10 seems fairly low.

hughsaunders: I'll increase it to 15, but the I don't want to match the CIT slave as that is under resourced for the number of executors it has. Also I'm planning to build multiple cloud slaves for redundancy and parallelism.

rpc_jobs/long_running_slave.yml
hughsaunders: Good spot, done.

playbooks/setup_jenkins_slave.yml
hughsaunders: Oops, thanks, done.

hughsaunders:
Notes from first run https://rpc.jenkins.cit.rackspace.net/job/Long-Running-Slave/1/console
- [ ] Jenkins user was given passwordless sudo access despite the parameter being false
- [ ] missing cred: MAAS_WEBHOOK_NOTIFICATION_PLAN_ID

hughsaunders:
Now ready for review. 

Successful Run:
- Jenkins Job: https://rpc.jenkins.cit.rackspace.net/job/Long-Running-Slave/3/console
- Jenkins Slave: https://rpc.jenkins.cit.rackspace.net/computer/lrs-3-1e03/
- Maas Checks: https://intelligence.rackspace.com/cloud/entities/en3XaAt9Vy

rpc_jobs/long_running_slave.yml
```diff
@@ -0,0 +1,96 @@
...
+          dir("rpc-gating"){
+            git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
+            common = load 'pipeline_steps/common.groovy'
+            pubCloudSlave = load 'pipeline_steps/pubcloud.groovy'
+            maas = load 'pipeline_steps/maas.groovy'
+          }
+          dir("rpc-maas"){
+            git branch: env.RPC_MAAS_BRANCH, url: env.RPC_MAAS_REPO
+          }
+          instance_name = common.gen_instance_name()
```
mattt416: any value in doing something like `instance_name = "LONG-RUNNING-SLAVE-${BUILD_NUMBER}"`, just so we can see quickly what the instance is without the more cryptical generic name derived from the job name?

rpc_jobs/rpc_aio.yml
```diff
@@ -193,10 +193,11 @@
            {TRIGGER_USER_VARS}
          UPGRADE_FROM_REF: "{UPGRADE_FROM_REF}"
       - rpc_gating_params
-      - single_use_slave_params:
+      - instance_params:
          IMAGE: "{IMAGE}"
          FLAVOR: "{FLAVOR}"
          REGION: "{REGION}"
+      - jenkins_node_params
```
mkam: Do we need to have the jenkins_node_params in each job? I don't think there are many use cases where we'd want to customize these parameters beyond the defaults other than in the long running slave job.

rpc_jobs/long_running_slave.yml
hughsaunders: They would be the only instances with the lrs prefix, so should be easy to spot. However as this has two votes, I've changed the gen_instance_name function to allow specifying a custom prefix, and have specified prefix="long-running-slave" for this job.

rpc_jobs/rpc_aio.yml
hughsaunders: I was having trouble combining default and bool filters in [ansible](https://github.com/rcbops/rpc-gating/pull/228/files#diff-5be9faccf8f48d89f54f9248ca4de6e3R5), so I decided to default the values in all jobs so that I didn't have to worry about it in ansible. However this does add one more thing that all new jobs must do, so I'll look again at finding a way that doesn't rely on those variables being set.

hughsaunders: Done with a [jinja if](https://github.com/rcbops/rpc-gating/pull/228/files#diff-5be9faccf8f48d89f54f9248ca4de6e3R11).

rpc_jobs/long_running_slave.yml
mattt416: I was actually thinking afterwards that unlike the webhook proxy node, if someone accidentally tore one of these instances down it wouldn't be the end of the world.  However, making it stand out and completely obvious is probably worth it in the long-run, as inevitably these instances will get accidentally deleted.

hughsaunders:
Addressed issues raised by @mattt416 and @mkam. 

Next test:

- Deploy job: https://rpc.jenkins.cit.rackspace.net/computer/long-running-slave-6-5135/
- Slave: https://rpc.jenkins.cit.rackspace.net/computer/long-running-slave-6-5135/


I updated the JJB job to be ubuntu compatible and converted to pipeline. However that job has to stay on an internal slave so it can access the Jenkins API. 

Ready for final review.

hughsaunders:
When this merges, the jjb job itself will fail to update because its changing type, and that isn't allowed. So the current jjb job will need to be deleted and jjb run manually from the cit slave: 

```
cd /var/lib/jenkins/workspace/Jenkins-Job-Builder
. .venv/bin/activate
git fetch && git reset --hard origin/master && jenkins-jobs --conf ../../jenkins_jobs.ini update -r rpc_jobs
```

PR #231, opened 2017-06-12
UG-607 Ensure pull requests link to Jira issues

git-harry:
Add a line of the form Issue: Jira key to the
description of pull requests in tracked repositories.
This change adds a job to update pull requests. Developers only add an
issue key reference to commit messages. This job runs a script to update
the pull request description body with a link to the issue to simplify
accessing it.

Issue: [UG-607](https://rpc-openstack.atlassian.net/browse/UG-607)

### DIFF
--- a/pipeline_steps/github.groovy
+++ b/pipeline_steps/github.groovy
@@ -25,4 +25,44 @@ def create_issue(
   }
 }
 
+/**
+ * Add issue link to pull request description
+ *
+ * Pull request commit messages include the issue key for an issue on Jira.
+ * Update the description of the current GitHub pull request with a link to
+ * the Jira issue.
+ */
+void add_issue_url_to_pr(){
+  List org_repo = env.ghprbGhRepository.split("/")
+  String org = org_repo[0]
+  String repo = org_repo[1]
+
+  Integer pull_request_number = env.ghprbPullId as Integer
+
+  dir(repo) {
+    git branch: env.ghprbSourceBranch, url: env.ghprbAuthorRepoGitUrl
+  }
+  String issue_key = common.get_jira_issue_key(repo)
+
+  withCredentials([
+    string(
+      credentialsId: 'rpc-jenkins-svc-github-pat',
+      variable: 'pat'
+    )
+  ]){
+    sh """#!/bin/bash -xe
+      cd $env.WORKSPACE
+      . .venv/bin/activate
+      python rpc-gating/scripts/ghutils.py add_issue_url_to_pr\
+        --org '$org'\
+        --repo '$repo'\
+        --pat '$pat'\
+        --pull-request-number '$pull_request_number'\
+        --issue-key '$issue_key'
+    """
+  }
+  return null
+}
+
+
 return this;
--- a/pipeline_steps/github.groovy
+++ b/pipeline_steps/github.groovy
@@ -14,12 +14,13 @@ def create_issue(
     sh """#!/bin/bash -xe
       cd ${env.WORKSPACE}
       . .venv/bin/activate
-      python rpc-gating/scripts/ghutils.py create_issue\
-        --tag '$tag'\
-        --link '$link'\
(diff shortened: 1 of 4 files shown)


scripts/ghutils.py
```diff
@@ -37,12 +37,71 @@ def create_issue(tag, link, org, repo, pat, label):
     )
 
 
+@click.command()
+@click.option(
```
hughsaunders: Options that are common between commands can be applied to the group instead of individual commands. However note that they must be specified before the command. eg: `script.py --shared-opt1 --shared-opt2=foo create_issue --create-issue-opt=bah`. This means that scripts that already use ghutils.py will need to be updated, but probably worth it to avoid repeating all the params.

rpc_jobs/pull_request_issue_link.yml
```diff
@@ -0,0 +1,31 @@
+- job:
```
hughsaunders: I think this needs to be a job-template so we can apply this to multiple repos. To start with rpco and rpc-gating would be useful.

hughsaunders:
recheck_lint

pipeline_steps/github.groovy
```diff
@@ -14,15 +14,49 @@ def create_issue(
...
+}
+
+/**
+ * Add issue link to pull request description
+ *
+ * Pull request commit messages include the issue key for an issue on Jira.
+ * Update the description of the current GitHub pull request with a link to
+ * the Jira issue.
+ */
+void add_issue_url_to_pr(){
```
hughsaunders: def void?

git-harry: `def` says the function returns an object `void` says it returns nothing. `def` does not mean the same thing as it does in Python.

pipeline_steps/github.groovy
```diff
@@ -14,15 +14,52 @@ def create_issue(
...
+  List org_repo = env.ghprbGhRepository.split("/")
+  String org = org_repo[0]
+  String repo = org_repo[1]
+  Integer pull_request_number = env.ghprbPullId.toInteger()
+  String issue_key = common.get_jira_issue_key(repo_path=repo)
```
hughsaunders: For this to work with repos that aren't rpc-gating, repo will need to be cloned.

hughsaunders: `repo_path=repo` is not a named parameter its an assignment, which should be ok because assignments return the assigned value in groovy, but not in CPS groovy 😭 😠. The only way of doing named parameters is by accepting a single Map, and then passing a single map.

rpc_jobs/pull_request_issue_link.yml
```diff
@@ -0,0 +1,53 @@
...
+          name: REPO
+          default: "{repo}"
+    dsl: |
+      node() {{
+        deleteDir()
+        if (REPO == "rpc_gating") {{
+          String branch = env.ghprbSourceBranch
+          String url = env.ghprbAuthorRepoGitUrl
+        }} else {{
+          String branch = env.RPC_GATING_BRANCH
```
hughsaunders: Because this variable is typed it goes out of scope after this block.  Would need to be defined outside or not typed.

hughsaunders:
recheck_lint

hughsaunders:
Heh of course lint job failed, because this branch hasn't been rebased to include the fix.

git-harry:
recheck_issue_link

git-harry:
recheck_issue_link

git-harry:
recheck_issue_link

hughsaunders:
recheck_issue_link

git-harry:
recheck_issue_link

git-harry:
recheck_issue_link

git-harry:
recheck_issue_link

git-harry:
recheck_issue_link

git-harry:
recheck_issue_link

PR #233, opened 2017-06-12
Rename LEM job 

alextricity25:
The LEM job has been renamed to "HW" to denote "Hardware".


mattt416:
We could also rename the job, and pair the project down to a single job simply as a reference.  What do you think @alextricity25 ?

alextricity25:
It sounds like you guys prefer that the code stays. I'll go ahead and rename it. Any preferences as to what the name should be?

mattt416:
s/lem/hardware/g ?

PR #234, opened 2017-06-12
RI-20 Add jobs for ocata-15.0 branch

alextricity25:
This commit adds jobs for the ocata-15.0 branch in rpc-openstack.
It also adds exclusions to prevent these jobs from running any permutation that
includes the trusty image and ocata series.

### DIFF
--- a/rpc_jobs/multi_node_aio.yml
+++ b/rpc_jobs/multi_node_aio.yml
@@ -13,6 +13,9 @@
           branch: newton-14.1
           USER_VARS: |
             tempest_test_sets: 'all'
+      - ocata150:
+          branch: ocata-15.0
+          USER_VARS: ""
       - master:
           branch: master
           USER_VARS: |
@@ -30,6 +33,11 @@
     exclude:
       - series: mitaka
         context: xenial
+      # Ocata onwards does not support trusty, so these jobs are excluded
+      - series: ocata150
+        context: trusty
+      - series: master
+        context: trusty
     jobs:
       - 'OnMetal-Multi-Node-AIO_{series}-{context}-{trigger}'
       - 'OnMetal-Multi-Node-AIO-Merge-Trigger_{series}'
--- a/rpc_jobs/rpc_aio.yml
+++ b/rpc_jobs/rpc_aio.yml
@@ -31,6 +31,9 @@
           branch: newton-14.1
           branches: "newton-14.1.*"
           UPGRADE_FROM_REF: "kilo"
+      - ocata150:
+          branch: ocata-15.0
+          branches: "ocata-15.0.*"
       - master:
           branch: master
           branches: "master"
@@ -100,6 +103,8 @@
         action: majorupgrade
       - series: master
         action: majorupgrade
+      - series: ocata
+        action: majorupgrade
       # Xenial builds are run for newton and above
       # as it is not supported distro before newton.
       - series: kilo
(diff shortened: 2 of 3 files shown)


mattt416:
The patch itself looks OK @alextricity25, but my only concern is that we keep adding branches without removing any.  From a Jenkins perspective this isn't really a problem that I'm aware of (short of seeing a bunch of additional jobs in the UI), but obviously from an rpc-o perspective it gets a bit confusing how things should be backported and trying to get a single code-base (rpc-gating) to gate 6 different branches.

alextricity25:
@mattt416 
Part of the reason things have become so complicated is because of a combination of 1) multiple work streams 2) Testing EOL branches because that's what most of our customers are using.

Below is a list of branches that we need to gate, with a small description. Hopefully it'll provide some clarity for both of us as to why we need to gate so many branches.

- master
- ocata-15.0 - Ocata isn't intended to have a release if we still have plans to release Pike(stalled worked ATM), but rather it's here as a "fallback" release in case we don't end up making it to Pike in time. It only needs to be minimally gated
- newton-14.1 - This is a stepping stone release for artifacting work, and a target release for the leap frog upgrade work and therefore needs to be gated
- newton-14.0 - This is the latest stable release of RPCO. It was signed off by our QE team and are making patch releases on a per request basis. It needs to be gated
- mitaka-13.1 - I presume customers are still upgrading to from liberty to mitaka, but I can't be sure. Patch releases are made at a per request basis. From my understanding, this  still needs to be gated as we have a good number of customers on mitaka.
- liberty-12.2 - I believe we also have a good number of customers on liberty. This branch is also used for upgrade jobs to upgrade from liberty to mitaka, so it needs to be gated

> Can we assume that the rpc-o team will deal w/ the fallout if we go the latter route?

What do you mean by "fallout"?

odyssey4me:
I'd like to understand what the point of this is, other than creating more work in terms of managing backports as we find things that need fixing.

Currently we're not testing upgrades between any branches other than liberty->mitaka and kilo->newton. We've put no effort into testing anything else, so us gating beyond ocata adds no value that I can see. Can someone who sees some sort of value here clue me in to help me understand?

It would make more sense to me to bed down (improve test coverage) Ocata, work on moving the magic testing from rpc-gating to rpc-openstack and to get an upgrade path from Newton to Ocata tested. Once Pike hits M3 we could get going for Pike in RPC-O master and therefore do the branch creation then. That should give us time to get a greenfield and upgrade implementation gating and to fix things.

Right now, to me, it feels like we're just adding more work to an already overextended team.

mattt416:
@alextricity25 you're also missing kilo, which support have said they want gating on.  So I actually mis-counted meaning we've got 7 branches to gate on and not 6.  That seems like an awful lot!

N/m on the fallout comment, I was thinking for some reason that this was completely untested jobs being added, but I suppose we did just cut this from master which was being gated fully.

mattt416:
Thinking about this further, we are trying to get to a place where we can troubleshoot each and every periodic failure.  So in actual fact there is a cost involved in adding more and more branches which I didn't originally identify.

odyssey4me:
Another note - there really should be no such things as 'minimally gated'. As we add series we should be adding test coverage. The test coverage should be at least the same as the series before, barring special circumstances (like we're only doing leapfrogs up to newton). If we remove test coverage for any series we are opening ourselves up to regression and avoidable bugs.

alextricity25:
Closing due to lack of interest

PR #241, opened 2017-06-14
Add dashboard link to MaaS Jira tickets

hughsaunders:
This change adds a link to the intelligence dashboard for an alert.
This will allow a user to see historical data related to the alert.

This change also adds a real sample MaaS webhook payload as reality
doesn't quite match the sample given in the docs.

Tests are also udpated to reflect the new sample payload.

### DIFF
--- a/webhooktranslator/tests/test_webhooktranslator.py
+++ b/webhooktranslator/tests/test_webhooktranslator.py
@@ -56,9 +56,13 @@ class WebhooktranslatorTestCase(unittest.TestCase):
         self.assertEqual(resp.status_code, 200)
         _, kwargs = mock_create_jira_issue.call_args
         self.assertIn("MaaS Alert:", kwargs['summary'])
-        self.assertIn("WARNING", kwargs['summary'])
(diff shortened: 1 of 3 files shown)


webhooktranslator/tests/test_webhooktranslator.py
```diff
@@ -56,7 +56,7 @@ def test_json_parsing(self, mock_create_jira_issue):
         self.assertEqual(resp.status_code, 200)
         _, kwargs = mock_create_jira_issue.call_args
         self.assertIn("MaaS Alert:", kwargs['summary'])
-        self.assertIn("WARNING", kwargs['summary'])
+        self.assertIn("CRITICAL", kwargs['summary'])
         self.assertIn("Full Payload", kwargs['description'])
         self.assertIn("alert", kwargs['labels'])
```
dankolbrs: Should this now include an assert for the dashboard link?

hughsaunders: Done.

PR #247, opened 2017-06-16
UG-641 Remove addChecksumRule

odyssey4me:
The requirement to add the checksum in this way
is only necessary for Liberty and earlier. Also,
this should be done in the rpc-o scripts, not
the gating scripts.

This action is already present in both kilo [1] and liberty [2].

[1] https://github.com/rcbops/rpc-openstack/pull/2370
[2] https://github.com/rcbops/rpc-openstack/pull/2369

Issue: [UG-641](https://rpc-openstack.atlassian.net/browse/UG-641)

### DIFF
--- a/pipeline_steps/deploy.groovy
+++ b/pipeline_steps/deploy.groovy
@@ -77,16 +77,4 @@ def upgrade_leapfrog(Map args) {
   upgrade("Leapfrog Upgrade", "leapfrog/ubuntu14-leapfrog.sh", args.environment_vars)
 }
 
-def addChecksumRule(){
-  sh """#!/bin/bash
-    cd /opt/rpc-openstack/rpcd/playbooks
-    ansible neutron_agent \
-      -m command \
-      -a '/sbin/iptables -t mangle -A POSTROUTING -p tcp --sport 80 -j CHECKSUM --checksum-fill'
-    ansible neutron_agent \
-      -m shell \
-      -a 'DEBIAN_FRONTEND=noninteractive apt-get install iptables-persistent'
-  """
-}
-
 return this;
--- a/rpc_jobs/rpc_aio.yml
+++ b/rpc_jobs/rpc_aio.yml
@@ -312,7 +312,6 @@
                   ]
                 aio_prepare.prepare()
                 deploy.deploy_sh(environment_vars: environment_vars)
-                deploy.addChecksumRule()
                 maas.deploy()
                 maas.verify()
                 tempest.tempest()
@@ -325,7 +324,6 @@
                   deploy.upgrade_minor(environment_vars: environment_vars)
                 }} else if (env.STAGES.contains("Major Upgrade")) {{
                   deploy.upgrade_major(environment_vars: environment_vars)
-                  deploy.addChecksumRule()
                   maas.deploy()
                   maas.verify()
                   tempest.tempest()


odyssey4me:
@mattt416 I don't understand what you mean. What does that have to do with his function which re-executes something already executed in RPC-O (for kilo/liberty) and completely unnecessary in Mitaka and above.

mattt416:
@odyssey4me have a look under which conditions rpc-o drops that bit of code.

mattt416:
Because of the way we run deploy.sh (`DEPLOY_OA=no` and `DEPLOY_AIO=yes` the first time, then `DEPLOY_OA=yes` and `DEPLOY_AIO=no` the second time), it means the iptables rules that are in `deploy.sh` in `kilo` and `liberty-12.2` branches does not get executed when we deploy via rpc-gating.  This is a non-issue in mitaka-13.1 onwards as the os_neutron role handles this for us when the os-neutron-install.yml playbook is run.

Essentially, we need to:

1. Figure out a way to run these rules in `liberty-12.2` and `kilo` with the way the gate is calling `deploy.sh`
2. Validate if a `liberty-12.2` -> `mitaka-13.1` upgrade requires us to manually drop the rules again, as we're currently doing.

The implementation of 1. could simply involve creating a script in rpc-openstack that gets called by jenkins, however it's probably best to see what changes will take place to deploy.sh first as we move gating logic out of rpc-gating and into rpc-openstack.

mattt416:
Testing 2. [here](https://rpc.jenkins.cit.rackspace.net/job/mattt-RPC-AIO_mitaka-trusty-majorupgrade-swift-periodic/3/).

odyssey4me:
We can use the role patcher facility in kilo and liberty to modify the os_neutron role to do the same thing as done in mitaka onwards. We've done this to cover other things that block deployments which need solving in the EOL branches of OSA.

mattt416:
Sounds good @odyssey4me -- I'll have a look at doing that today.

https://rpc.jenkins.cit.rackspace.net/job/mattt-RPC-AIO_mitaka-trusty-majorupgrade-swift-periodic/3/ failed on the 2nd tempest run, it looks like the iptables rules are getting lost somewhere.  As a work-around we could have rpc-openstack's test-upgrade.sh write `neutron_metadata_checksum_fix: true` so that the upgrade of neutron re-applies the rule, but I'll have a look at patching kilo/liberty first and seeing if we shove that into scripts/deploy.sh or similar instead.

odyssey4me:
Both https://github.com/rcbops/rpc-openstack/pull/2370 and https://github.com/rcbops/rpc-openstack/pull/2369 have merged. Are we good to go to remove this now?

odyssey4me:
I've rebased the patch and added the respective issue into the commit to appease the issue-link.

mattt416:
This should be good to go @odyssey4me, but running one last test [here](https://rpc.jenkins.cit.rackspace.net/job/mattt-RPC-AIO_mitaka-trusty-majorupgrade-swift-periodic/7/) to verify.

mattt416:
Sadly, PR https://github.com/rcbops/rpc-openstack/pull/2369 has broken liberty->mitaka upgrades.  What appears to happen is that we create some new files in /opt/rpc-openstack/openstack-ansible/playbooks/roles/os_neutron in the liberty branch when we apply the patch via the patcher role.  Via rpc-gating, we `git reset --hard` and then switch branches.  The problem is that in mitaka the os_neutron role moves out into /etc/ansible/roles, leaving /opt/rpc-openstack/openstack-ansible/playbooks/roles/os_neutron/files behind.  When setup-openstack.yml is run, it then pukes because it sees the empty-ish, non-functional role in /opt/rpc-openstack/openstack-ansible/playbooks/roles.

I've created a workaround and am doing a full upgrade test against it before I create PRs in rpc-openstack.

PR #248, opened 2017-06-16
Allow concurrent artifact PR tests

odyssey4me:
Only the Apt and Pipeline jobs need to not
be concurrent. All others can be done in
parallel. This should help speed up feedback
for PR's.

The pipeline jobs should not be concurrently
run because any one job will produce artifacts.

### DIFF
--- a/rpc_jobs/rpc_artifact_build.yml
+++ b/rpc_jobs/rpc_artifact_build.yml
@@ -16,10 +16,12 @@
           IMAGE: "Ubuntu 16.04.2 LTS prepared for RPC deployment"
     context:
       - Git
-      - Apt
+      - Apt:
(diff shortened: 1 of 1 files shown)


rpc_jobs/rpc_artifact_build.yml
```diff
@@ -74,7 +76,7 @@
     # TEMPLATE
     name: 'RPC-Artifact-Build_{series}-{image}-{context}-{ztrigger}'
     project-type: workflow
-    concurrent: false
+    concurrent: true
```
hughsaunders: This will mean every job is concurrent. Did you mean `concurrent: {concurrent}` ? You'd also need to add `concurrent: true` to template defaults.

PR #250, opened 2017-06-16
add additional context for irr gate jobs

cloudnull:
The context field is now being used to pass in an OpenStack release name
into the gate job. This will be used to ensure the role are able to gate
on all major releases presently supported by RPC.

### DIFF
--- a/rpc_jobs/irr_role_test.yml
+++ b/rpc_jobs/irr_role_test.yml
@@ -2,7 +2,7 @@
     name: 'RPC-IRR-Jobs'
     repo:
       # IRR REPOS ARE ADDED HERE
-      - rpc_maas:
+      - rpc-maas:
           irr_repo_url: https://github.com/rcbops/rpc-maas
       - rpc-role-logstash:
           irr_repo_url: https://github.com/rcbops/rpc-role-logstash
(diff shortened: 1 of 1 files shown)


rpc_jobs/irr_role_test.yml
```diff
@@ -15,8 +15,37 @@
           IMAGE: "Ubuntu 16.04 LTS (Xenial Xerus) (PVHVM)"
       - trusty:
           IMAGE: "Ubuntu 14.04 LTS (Trusty Tahr) (PVHVM)"
+    # NOTE(cloudnull): Context are used to set the OpenStack series used in the gate.
```
hughsaunders: Please change its name to reflect its function.

rpc_jobs/irr_role_test.yml
```diff
@@ -15,8 +15,37 @@
...
     context:
-      - base
+      - master
+      - ocata
+      - newton
+      - mitaka
+      - liberty
+      - kilo
```
hughsaunders: This will spin up ~10 15GB instances per PR. That seems a lot. Do any of these instances need to 15gb? Could we reduce the instance size for these jobs?

cloudnull: we could make them 8GB ?

rpc_jobs/irr_role_test.yml
cloudnull: "context" is the correct name as it defines the context in which the gate job will test the playbooks.

PR #251, opened 2017-06-16
UG-573 Add tests for name formatting

dankolbrs:
Adds lint tests for name formatting enforcement of file and job
names

### DIFF
--- a/lint.sh
+++ b/lint.sh
@@ -77,6 +77,20 @@ check_bash(){
   done < <(find ${fargs[@]} -iname \*.sh)
 }
 
+check_naming_standards() {
+  # Limits lint to only gating unique files
+  # Excludes webhooktranslator as it has additional packaging
+  # and it's own tox
+  # Excludes NonCPS.groovy as this filename is required, but
+  # not matching rpc-gating conventions
+  dirs_to_lint="pipeline_steps,rpc_jobs,scripts"
+  exclude_files="NonCPS.groovy"
+  python scripts/lint_naming_conventions.py \
+    --dirs ${dirs_to_lint} --exclude ${exclude_files} \
+    && echo "Naming conventions: OK" \
+    || { echo "Naming conventions: FAIL"; rc=1; }
+}
+
 check_python(){
   flake8 --exclude=.lintvenv,webhooktranslator . \
     && echo "Python syntax ok" \
@@ -96,6 +110,7 @@ check_groovy
 check_ansible
 check_bash
 check_python
+check_naming_standards
 check_webhooktranslator
 
 if [[ $rc == 0 ]]
--- a/scripts/lint_naming_conventions.py
+++ b/scripts/lint_naming_conventions.py
@@ -1,0 +1,109 @@
+#!/usr/bin/env python
+
+# Script used to review jenkins jobs and report job names
+# not following conventions
+# https://github.com/rcbops/rpc-gating#naming-conventions
+
+import argparse
+import os
+import re
+import sys
+import yaml
+
+
+def parse_args():
+    description = "Parses given directories for rpc-gating naming conventions"
+    parser = argparse.ArgumentParser(description=description)
+    parser.add_argument("--exclude-files",
+                        required=False,
+                        default=[],
+                        help="Comma seperated list of files to exclude")
+    parser.add_argument("--dirs",
+                        required=True,
+                        help="Comma seperated dirs to parse")
+    return parser.parse_args()
+
+
+def parse_jjb_file(in_dir, in_file):
+    _rc = 0
+    filename = os.path.join(in_dir, in_file)
+    file_yaml = ''
+    # blanket catch errors reading/parsing yaml
+    try:
+        file_yaml = yaml.safe_load(open(filename).read())
+    except:
+        out = "Lint error: yaml parsing error parsing file %s\n" % (filename)
+        sys.stderr.write(out)
+        return 1
+
+    # Catch NoneType from blank yaml
+    if not file_yaml:
+        out = "Lint error: NoneType returned for yaml file %s\n" % (filename)
+        sys.stderr.write(out)
+        return 1
+
+    for item in file_yaml:
+        if "project" in item:
+            # check project name
+            if parse_job_name(item["project"]["name"], filename):
+                _rc = 1
+            # check project jobs names
+            for job in item["project"]["jobs"]:
+                if parse_job_name(job, filename):
+                    _rc = 1
+
+        if "job-template" in item:
+            # check job-template name
+            if parse_job_name(item["job-template"]["name"], filename):
+                _rc = 1
+
+        if "job" in item:
+            # check individual job
+            if parse_job_name(item["job"]["name"], filename):
+                _rc = 1
+    return _rc
+
+
+def parse_job_name(job_name, file_name):
+    regex = '^([a-zA-Z0-9]+-?)+([_]([{][a-zA-Z0-9]+[}]-?)+)?$'
+    match = re.match(regex, job_name)
+    if not match:
+        out = "Lint error: Job name \"%s\" in \"%s\"" % (job_name, file_name)
+        out += "\n\tdoes not conform to rpc-gating naming conventions.\n"
+        sys.stderr.write(out)
+        return 1
+    return 0
+
+
+def parse_file_name(in_dir, in_file):
+    regex = '^([a-z0-9]+_?)+(\.{1}[a-z]+)?$'
+    filename = os.path.join(in_dir, in_file)
+    match = re.match(regex, in_file)
+    if not match:
+        out = "Lint error: filename \"%s\"\n\t does not " % filename
+        out += "conform to rpc-gating conventions\n"
+        sys.stderr.write(out)
+        return 1
+    return 0
+
+
+if __name__ == "__main__":
+    args = parse_args()
+    rc = 0
+    for _dir in args.dirs.split(','):
+        for root, dirs, files in os.walk(_dir):
+            for _file in files:
+                if _file in args.exclude_files:
+                    pass
+                else:
+                    if (_file.endswith(".yml")) or (_file.endswith(".yaml")):
+                        if parse_jjb_file(_dir, _file):
+                            rc = 1
+                    if parse_file_name(_dir, _file):
+                        rc = 1
+    if rc:
+        out = "RPC-Gating naming conventions:\n\t%s\n" % \
+            "https://github.com/rcbops/rpc-gating#naming-conventions"
+        sys.stderr.write(out)
+
+    sys.exit(rc)
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,3 +10,4 @@ six
 packaging
 appdirs
 rackspace_monitoring
+PyYAML


dankolbrs:
Output from intentionally breaking conventions:
```
Name lint error on jira-utils.py
File naming fail: scripts/jira-utils.py
Lint error on Multi_Node-AIO-Jobs in rpc_jobs/multi_fail.yml
Job naming fail: rpc_jobs/multi_fail.yml

----------------------
Syntax check fail :(
----------------------
```

scripts/lint_naming_conventions.py
```diff
@@ -0,0 +1,64 @@
...
+def parse_job_name(job_name, file_name):
+    regex = '^([a-zA-Z0-9]+-?)+([_]([{][a-zA-Z0-9]+[}]-?)+)?$'
+    match = re.match(regex, job_name)
+    if not match:
+        sys.stderr.write("Lint error on %s in %s\n" % (job_name, file_name))
+        sys.exit(1)
+
+
+def parse_file_name(file_name):
+    regex = '^([a-zA-Z0-9]+_?)+(\.{1}[a-z]+)?$'
```
mkam: Can you remove the `A-Z`? File names should be all lowercase.

dankolbrs: Yep! Will fix.

This is going to result in a fail with https://github.com/rcbops/rpc-gating/blob/master/pipeline_steps/NonCPS.groovy until I get a PR up and merged. It looks like it should be safe to lowercase the filename.

lint.sh
```diff
@@ -77,6 +77,26 @@ check_bash(){
...
 }
 
+check_naming_standards() {
+  # Limits lint to only gating unique files
+  # Excludes webhooktranslator as it has additional packaging
+  # and it's own tox
+  dirs_to_lint="pipeline_steps rpc_jobs scripts"
+  for _dir in ${dirs_to_lint}; do
+    for _file in $(find ${_dir} -type f -not -path '.*/.*'); do
+      python scripts/lint_naming_conventions.py \
```
hughsaunders: I don't really like the new python interpreter for every file approach. I'd prefer to either pass the list of dirs into python, or do the regex match in bash. http://wiki.bash-hackers.org/syntax/ccmd/conditional_expression#regular_expression_matching

lint.sh
```diff
@@ -77,6 +77,26 @@ check_bash(){
...
+        --filename ${_file} \
+        || { echo "File naming fail: ${_file}"; rc=1; }
+    done
+  done
+
+  for _file in $(find rpc_jobs/ -type f); do
+    python scripts/lint_naming_conventions.py \
```
hughsaunders: I don't have a problem with this as bash is not suitable for yaml parsing. Otoh if you did do the tree walk in python it could check job names on all yaml files based on extension.

dankolbrs:
Looking further into it, I'm not sure if renaming `NonCPS.groovy` is going to work or not. Giving it a shot and running a test.

scripts/lint_naming_conventions.py
```diff
@@ -0,0 +1,64 @@
...
+        if "job" in item:
+            # check individual job
+            parse_job_name(item["job"]["name"], filename)
+
+
+def parse_job_name(job_name, file_name):
+    regex = '^([a-zA-Z0-9]+-?)+([_]([{][a-zA-Z0-9]+[}]-?)+)?$'
+    match = re.match(regex, job_name)
+    if not match:
+        sys.stderr.write("Lint error on %s in %s\n" % (job_name, file_name))
```
hughsaunders: It would be helpful to know that its the job name that failed a check, rather than the job itself. Maybe:
```
"Lint error: Job name {job_name} from {file_name} doesn't confirm to rpc-gating naming standards".format(job_name=job_name, file_name=file_name)
```

dankolbrs:
Running in https://rpc.jenkins.cit.rackspace.net/job/danko-rpc-aio/4/

scripts/lint_naming_conventions.py
```diff
@@ -0,0 +1,64 @@
...
+        sys.exit(1)
```
hughsaunders: I'd prefer to find all the lint fails on the first run. Rather than having to run multiple times to discover them all.

pipeline_steps/non_cps.groovy
```diff
@@ -0,0 +1,9 @@
+// Dummy class so that lint checks dont fail due to @NonCPS not being available
```
hughsaunders: This file must be NonCPS.groovy

lint.sh
dankolbrs: Passing the list of dirs to python does sound like a better idea. I started with approaching it in bash, but I thought the regex got a little too sloppy and would be easier to follow in python.

pipeline_steps/non_cps.groovy
dankolbrs: Dang, thats what it was looking like to me. I'll add an exclude.

scripts/lint_naming_conventions.py
```diff
@@ -0,0 +1,93 @@
...
+        for root, dirs, files in os.walk(_dir):
+            for _file in files:
+                if args.exclude_files and _file in args.exclude_files:
+                    pass
+                else:
+                    if args.parse_jjb and _file.endswith(".yml"):
+                        if parse_jjb_file(_dir, _file):
+                            rc = 1
+                    else:
+                        if parse_file_name(_dir, _file):
```
hughsaunders: If you moved this out of the inner else, you'd only need to walk the directory tree once.

dankolbrs: Agreed, but I wanted to keep the `--parse-jjb` flag to act as a unique run only for job names on yml files, and not do it all at once. This way the directories being linted and exclude files could be changed if needed if another directory needs to be added for jobs. I figure the little bit of flexibility that offers is worth the second run of the script. Additionally this separates the lint to clearly be either job names, or file names.

In its current use it doesn't need that, and I can't imagine there's a great likelihood of this being required in the near future, so I'd be fine changing it if ifs not worth that seperation

scripts/lint_naming_conventions.py
```diff
@@ -0,0 +1,93 @@
...
+def parse_args():
+    description = "Parses JJB files to ensure naming conventions followed"
+    parser = argparse.ArgumentParser(description=description)
+    parser.add_argument("--exclude-files",
+                        required=False,
+                        help="Comma seperated list of files to exclude")
+    parser.add_argument("--dirs",
+                        required=True,
+                        help="Comma seperated dirs to parse")
+    parser.add_argument("--parse-jjb", required=False,
```
mattt416: When would you not run this?

scripts/lint_naming_conventions.py
```diff
@@ -0,0 +1,93 @@
...
+
+import argparse
+import os
+import re
+import sys
+import yaml
+
+
+def parse_args():
+    description = "Parses JJB files to ensure naming conventions followed"
```
mattt416: Total nit, but this description is a bit misleading

scripts/lint_naming_conventions.py
```diff
@@ -0,0 +1,93 @@
...
+    return 0
+
+
+if __name__ == "__main__":
+    args = parse_args()
+    rc = 0
+    for _dir in args.dirs.split(','):
+        for root, dirs, files in os.walk(_dir):
+            for _file in files:
+                if args.exclude_files and _file in args.exclude_files:
```
mattt416: Could exclude_files default to [] so that we don't have to check for it?

scripts/lint_naming_conventions.py
```diff
@@ -0,0 +1,93 @@
...
+                _rc = 1
+    return _rc
+
+
+def parse_job_name(job_name, file_name):
+    regex = '^([a-zA-Z0-9]+-?)+([_]([{][a-zA-Z0-9]+[}]-?)+)?$'
+    match = re.match(regex, job_name)
+    if not match:
+        out = "Lint error: Job name \"%s\" in \"%s\"" % (job_name, file_name)
+        out += "\n\tdoes not conform to rpc-gating naming conventions.\n"
```
mattt416: Any value in including the naming conventions URL when a failure occurs?

scripts/lint_naming_conventions.py
dankolbrs: I hit on this a bit in a reply to @hughsaunders , but my intention was to separate this linting between filenames and JJB's. Rather than have this parse all directories for filenames **and** any job/job-template/project names following standards, to only do either/or per run. I also think this makes it a bit more obvious in `lint.sh` that one run of it is file naming conventions, the other is job naming conventions. 

But, the way that this is called in lint.sh, it could really go either way and I don't have a specific example of when this may include a different directory of JJBs to be linted but not filenames.

scripts/lint_naming_conventions.py
dankolbrs: Totally legit. I'll fix.

scripts/lint_naming_conventions.py
dankolbrs: This felt more clear to me than setting default parameters, but good call, an empty list isn't really a magic variable. Will fix.

scripts/lint_naming_conventions.py
dankolbrs: I considered finding a good place to throw that out, and could be done at the end if there is a failure. But I thought otherwise because
* other lints don't provide links to the exact conditions being checked
* the naming standards are in the short readme of the rpc-gating repo, not buried in documentation or requiring a doc build process
* `lint.sh` output is already 227 lines

But, if adding ~2 lines of code saves someone 5 minutes or more in finding or asking for help in finding the naming conventions, seems worth it. I'll add it.

scripts/lint_naming_conventions.py
mattt416: OK, makes sense.  May be worth noting that if `--parse-jjb` is supplied then the other check isn't run?

scripts/lint_naming_conventions.py
dankolbrs:
```
Lint error: filename "rpc_jobs/multi-fail.yml"
	 does not conform to rpc-gating conventions
RPC-Gating naming conventions:
	https://github.com/rcbops/rpc-gating#naming-conventions
File naming conventions: FAIL
Job naming conventions: OK
```

scripts/lint_naming_conventions.py
```diff
@@ -0,0 +1,93 @@
...
+if __name__ == "__main__":
+    args = parse_args()
+    rc = 0
+    for _dir in args.dirs.split(','):
+        for root, dirs, files in os.walk(_dir):
+            for _file in files:
+                if args.exclude_files and _file in args.exclude_files:
+                    pass
+                else:
+                    if args.parse_jjb and _file.endswith(".yml"):
```
mattt416: If I specify `--parse-jjb` but the file doesn't end in `.yml`, then it runs `parse_file_name`.  Expected?  Also, `.yaml` is probably a legitimate extension for those files also?

scripts/lint_naming_conventions.py
```diff
@@ -0,0 +1,99 @@
...
+                        action='store_true')
+    return parser.parse_args()
+
+
+def parse_jjb_file(in_dir, in_file):
+    _rc = 0
+    filename = os.path.join(in_dir, in_file)
+    file_yaml = yaml.safe_load(open(filename).read())
+    for item in file_yaml:
+        if "project" in item:
```
mattt416: Note that the `yaml.safe_load` will throw an exception if it can't load the YAML (syntax error etc.)  Do you want to account for that?  Also, if the file is empty, the for loop bombs as file_yaml = `None`.  The former is probably worth handling, the latter probably much less likely.

scripts/lint_naming_conventions.py
dankolbrs: # Side effect, but seems acceptable. Definite hole I overlooked in keeping this separate, and would be more concise to abandon my plan and just have it all together. 
# `.yaml` would be valid, but rpc-o/osa/rpc-gating appear standardized around `.yml` even though it is not a naming convention, and could completely bypass the linting altogether. Worth adding.

scripts/lint_naming_conventions.py
dankolbrs: Both seem worth adding.

PR #253, opened 2017-06-20
UG-651 Remove nova_cross_az_attach override

mattt416:
This var is correctly set in rpc-openstack now, there should no longer
be a need to override this in rpc-gating.


mattt416:
Once this merges, we should be able to uncomment the override in master of rpc-openstack.  It needs to remain commented in newton-14.0 and newton-14.1 for now.

mattt416:
recheck_lint

PR #256, opened 2017-06-23
[WIP] UG-653 Execute test_holland.yml from rpc-openstack

mattt416:
This commit uses rpc-openstack's in-tree version of test_holland.yml,
rather than copying the version into rpc-openstack from rpc-gating.

NOTE: https://github.com/rcbops/rpc-openstack/pull/2367/ needs to first
      merge and get backported to all branches before this PR can
      merge.


mattt416:
Closing -- ideally this test should live in the rpc support role itself, but doesn't yet.

PR #258, opened 2017-06-26
UG-658 Clean up all slaves

hughsaunders:
This commit addresses a problem where the Jenkins section of the
periodic cleanup job fails because its running on a slave that doesn't
have access to the Jenkins API.

When fixing this I realised that most of the cleanups are node
specific, so we actually need to run them against every node.
This job introduces a parallel stage to clean up various objects on
all slaves, then a second stage to run public cloud and Jenkins node
cleanups on a single internal slave.

Issue: [UG-658](https://rpc-openstack.atlassian.net/browse/UG-658)

### DIFF
--- a/rpc_jobs/periodic_cleanup.yml
+++ b/rpc_jobs/periodic_cleanup.yml
@@ -25,29 +25,58 @@
       - build-discarder:
           days-to-keep: 3
     dsl: |
-      node(){
-        deleteDir()
-        dir("rpc-gating") {
-          stage("Prepare"){
-            git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
-            common = load 'pipeline_steps/common.groovy'
+      // Get list of jenkins slaves
+      @NonCPS
+      def getLongRunningNodes() {
+        return jenkins.model.Jenkins.instance.nodes.grep {
+          // Only return nodes whose name starts with one
+          // of these expressions. Single use slaves won't match
+          // so are filtered out.
+          node -> node.name =~ /^(long-|master|rpc-jenkins-n)/
+        }.collect {
+          // node objects are not serializable so return a list
+          // of names instead :(
+          node -> node.name
+        }
+      }
+
+      // run node cleanups on all nodes
+      def nodes = getLongRunningNodes()
+      def parallel_steps = [:]
+      for (int i=0; i<nodes.size(); i++){
+        nodeName = nodes[i]
+        parallel_steps['node_'+nodeName] = {
+          node(nodeName){
+            stage("Cleanup "+nodeName){
+              deleteDir()
+              dir("rpc-gating") {
+                git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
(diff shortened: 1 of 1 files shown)


hughsaunders:
[Test job](https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/Periodic-Cleanup-testing/detail/Periodic-Cleanup-testing/3/pipeline)
Blue ocean screenshot:

rpc_jobs/periodic_cleanup.yml
```diff
@@ -25,29 +25,53 @@
...
-          stage("Cleanup Jenkins TMP Dir"){
-            sh "scripts/tmp_cleanup.sh"
-          }
-          stage("Cleanup Docker Objects"){
-            sh "scripts/docker_cleanup.sh"
+      // Get list of jenkins slaves
+      @NonCPS
+      def getNodes() {
+        return jenkins.model.Jenkins.instance.nodes.grep {
+          node -> node.name =~ /^(long-|master|rpc-jenkins-n)/
```
git-harry: It's not clear, at least to me, what these patterns match. For example `rpc-jenkins-n` doesn't appear in the codebase.

rpc_jobs/periodic_cleanup.yml
```diff
@@ -25,29 +25,53 @@
...
+        }.collect {
+          node -> node.name
+        }
+      }
+
+      // run node cleanups on all nodes
+      def nodes = getNodes()
+      def branches = [:]
+      for (int i=0; i<nodes.size(); i++){
```
git-harry: I believe Groovy allows the following:

```
for ( nodeName in nodes ) {
```
This strikes me as easier to read.

rpc_jobs/periodic_cleanup.yml
```diff
@@ -25,29 +25,53 @@
...
+        return jenkins.model.Jenkins.instance.nodes.grep {
+          node -> node.name =~ /^(long-|master|rpc-jenkins-n)/
+        }.collect {
+          node -> node.name
+        }
+      }
+
+      // run node cleanups on all nodes
+      def nodes = getNodes()
+      def branches = [:]
```
git-harry: Use of the term `branches` is probably not ideal given that generally in the repo it refers to Git branches.

rpc_jobs/periodic_cleanup.yml
hughsaunders: - **long-** the start of long-running-slave, matches all slaves created by the long running slave job
- **master** matches the master node 😉 
- **rpc-jenkins-n** matches CIT/TES slaves configured via puppet.

rpc_jobs/periodic_cleanup.yml
hughsaunders: Done.

rpc_jobs/periodic_cleanup.yml
hughsaunders: This is a workaround for an unserialisable iterator. (Blame CPS).

PR #262, opened 2017-06-28
AR-106 Do not pull apt data from rpc-repo

odyssey4me:
For the apt artifact creation, rpc-repo is
primarily a publishing point and a place to
store a backup of the database. We should
therefore not be pulling the data from it
every time we run a job, but instead just
execute the job on the long-running node
which has the current database.

Issue: [AR-106](https://rpc-openstack.atlassian.net/browse/AR-106)

### DIFF
--- a/rpc_jobs/rpc_artifact_build.yml
+++ b/rpc_jobs/rpc_artifact_build.yml
@@ -18,6 +18,7 @@
       - Git
       - Apt:
           allow_concurrent: false
+          PULL_FROM_MIRROR: "NO"
       - Python
       - Container
       - Pipeline:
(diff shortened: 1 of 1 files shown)


rpc_jobs/rpc_artifact_build.yml
```diff
@@ -18,6 +18,7 @@
       - Git
       - Apt:
           allow_concurrent: false
+          PULL_FROM_MIRROR: "NO"
```
evrardjp: I don't think it's a good idea.

It means the aptly build will not start by pulling from the mirror. So if we have a non persistent node, it will plainly fail.

For the persistent nodes, it will not fetch anything at the beginning of the process. For a clean node, it will have to redownload everything, and will eventually fail due to OOM and too many things to process. And won't ever be able to recover, because it won't ever resume.

I think it could be better if we split the push of the aptly db/resulting published repository.
In that case, we can have more granularity in what we do.

Please note that if we intend to keep a long running node, maybe we should just have this node as a publishing node, as aptly is intended.

odyssey4me:
Github is dumb and hasn't picked up my rebase. I'll open a new PR.

PR #268, opened 2017-07-03
Include creds for maas testing

cloudnull:
At present the IRR tests do not include API tests becuase the creds used
for maas were passed into the the job. This change allows the job to pass
in environment variables which contain the needed credentials and will
enable us to test additional API interactions from within any IRR test.

Issue: [UG-618](https://rpc-openstack.atlassian.net/browse/UG-618)

### DIFF
--- a/pipeline_steps/irr_role_tests.groovy
+++ b/pipeline_steps/irr_role_tests.groovy
@@ -20,9 +20,12 @@ def run_irr_tests() {
                   ]
                 ]
               ]) // checkout
-              sh """#!/bin/bash
-              bash ./run_tests.sh
-              """
+              List maas_vars = maas.get_maas_token_and_url()
+              withEnv(maas_vars) {
+                sh """#!/bin/bash
+                bash ./run_tests.sh
+                """
+              }
             } // dir
           } // ansiColor
         } catch (e) {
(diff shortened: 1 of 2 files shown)


pipeline_steps/irr_role_tests.groovy
```diff
@@ -20,9 +20,14 @@ def run_irr_tests() {
                   ]
                 ]
               ]) // checkout
-              sh """#!/bin/bash
-              bash ./run_tests.sh
-              """
+              withCredentials(common.get_cloud_creds()){
```
hughsaunders: if you only need cloud creds to get the maas token and url, then this isn't needed as maas.maas_utils already binds common.get_cloud_creds.

pipeline_steps/irr_role_tests.groovy
```diff
@@ -20,9 +20,14 @@ def run_irr_tests() {
...
+                List maas_vars = maas.get_maas_token_and_url()
```
hughsaunders: This will fail as maas isn't [loaded](https://github.com/rcbops/rpc-gating/blob/master/rpc_jobs/irr_role_test.yml#L123).

cloudnull:
the commit message has """Issue: UG-618""" is there more that's needed?

PR #269, opened 2017-07-03
LA-323 Add configurable resource load to leapfrog

git-harry:
The function `upgrade` is modified to enable resources, e.g. compute
instances, to be created prior to a leapfrog upgrade. These resources
will be used elsewhere to monitor the end-user impact of the upgrade.

Three parameters are exposed to control the number of resources that are
created:
- LEAPFROG_TEST_NETWORKS
- LEAPFROG_TEST_SERVERS
- LEAPFROG_TEST_VOLUMES

Depends-on: https://rpc-openstack.atlassian.net/browse/LA-321

Issue: [LA-323](https://rpc-openstack.atlassian.net/browse/LA-323)

### DIFF
--- a/pipeline_steps/deploy.groovy
+++ b/pipeline_steps/deploy.groovy
@@ -45,6 +45,19 @@ def deploy_sh(Map args) {
   ) // conditionalStage
 }
 
+void create_test_resource_load(Integer servers, Integer networks, Integer volumes) {
+  dir("/opt/rpc-openstack/scripts/leapfrog/playbooks/") {
+    sh """
+      openstack-ansible \
+        -i /opt/rpc-openstack/openstack-ansible/playbooks/inventory/dynamic_inventory.py \
+        -e server_count=$servers \
+        -e network_count=$networks \
+        -e volume_count=$volumes \
+        generate-resources.yml
+    """
+  }
+}
+
 def upgrade(String stage_name, String upgrade_script, List env_vars,
             String branch = "auto", String dest = "/opt") {
   common.conditionalStage(
@@ -56,6 +69,13 @@ def upgrade(String stage_name, String upgrade_script, List env_vars,
           sh "git reset --hard"
         }
         common.prepareRpcGit(branch, dest)
+        if ( stage_name == "Leapfrog Upgrade" ) {
+          create_test_resource_load(
+            env.GENERATE_TEST_SERVERS as Integer,
+            env.GENERATE_TEST_NETWORKS as Integer,
+            env.GENERATE_TEST_VOLUMES as Integer,
+          )
+        }
         dir("/opt/rpc-openstack"){
           sh """
             scripts/$upgrade_script
--- a/rpc_jobs/defaults.yml
+++ b/rpc_jobs/defaults.yml
@@ -22,3 +22,7 @@
     JENKINS_NODE_EXECUTORS: "2"
     jenkins_node_exclusive: true
     allow_jenkins_sudo: true
+    # generate_test_params
+    GENERATE_TEST_NETWORKS: "0"
+    GENERATE_TEST_SERVERS: "0"
+    GENERATE_TEST_VOLUMES: "0"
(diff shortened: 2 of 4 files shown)


git-harry:
Test build https://rpc.jenkins.cit.rackspace.net/job/git-harry-RPC-AIO_newton141-trusty-leapfrogupgrade-swift-periodic/18/

pipeline_steps/deploy.groovy
```diff
@@ -55,6 +68,13 @@ def upgrade(String stage_name, String upgrade_script, List env_vars) {
           sh "git reset --hard"
         }
         common.prepareRpcGit()
+        if ( stage_name == "Leapfrog Upgrade" ) {
+          create_test_resource_load(
+            env.LEAPFROG_TEST_SERVERS as Integer,
```
evrardjp: Naming is hard I know, but this load thing could be used for intra-series too, so it's good to rename this env variables.

GENERATE_TEST_{SERVERS,NETWORKS,VOLUMES}

rpc_jobs/defaults.yml
```diff
@@ -21,3 +21,7 @@
     JENKINS_NODE_EXECUTORS: "2"
     jenkins_node_exclusive: true
     allow_jenkins_sudo: true
+    # rpc_leapfrog_params
+    LEAPFROG_TEST_NETWORKS: "0"
```
evrardjp: Same remark here for the L25-27.

I'd rename rpc_leapfrog_params L24 to load_generator_params

rpc_jobs/params.yml
```diff
@@ -148,3 +148,16 @@
       - string:
           name: "RPC_MAAS_BRANCH"
           default: "{RPC_MAAS_BRANCH}"
+
+- parameter:
+    name: rpc_leapfrog_params
```
evrardjp: load_generator_params

rpc_jobs/params.yml
```diff
@@ -148,3 +148,16 @@
...
+    parameters:
+      - string:
+          name: "LEAPFROG_TEST_NETWORKS"
```
evrardjp: GENERATE_TEST_

rpc_jobs/rpc_aio.yml
```diff
@@ -67,9 +67,13 @@
             Connect Slave,
             Prepare Deployment,
             Deploy RPC w/ Script,
+            Install Tempest,
             Leapfrog Upgrade,
             Cleanup,
             Destroy Slave
+          LEAPFROG_TEST_NETWORKS: "6"
```
evrardjp: Please see above

rpc_jobs/rpc_aio.yml
```diff
@@ -255,6 +259,10 @@
               Pause (use to hold instance for investigation before cleanup)
               Cleanup
               Destroy Slave
+      - rpc_leapfrog_params:
```
evrardjp: Please see above.

mancdaz:
@hughsaunders would very much appreciate some eyes if you have the cycles

pipeline_steps/deploy.groovy
git-harry: Could be used for intra-series or will be? When is that likely to happen? I used leapfrog in the variable names because, while the yaml file requires they be specified as parameters for all jobs, they only have meaning for leapfrog upgrades. Using specific variable names should prevent any confusion regarding their use and given the complexity of this repo I think anything that makes the code more readable is a benefit.

If at some point in the future this functionality does get used more generally, there is nothing to prevent the names being changed. Given that we are not committing to this variables being stable, I do not believe this causes an issue.

odyssey4me: @git-harry As soon as leapfrog upgrades use this, intra-series will to. It will be a very fast follow. That said, I don't see that as a reason to hold this back - we can be responsible for changing it to suit more general needs once the leapfrog implementation is done and confirmed to be working.

evrardjp:
@git-harry Could you resolve conflicts, this way we can go forward? It sounds good.

evrardjp:
Let's go forward.

PR #272, opened 2017-07-03
commit on top of old fork

hughsaunders:


hughsaunders:
Created for testing, no longer required.

PR #275, opened 2017-07-04
LA-313 Create influx job

evrardjp:
Job to create an influxdb long running instance.
Uses rpc-maas master branch to deploy using a playbook.

This job is not generic, as the checkouts are quite non-standard.

Issue: [LA-313](https://rpc-openstack.atlassian.net/browse/LA-313)

### DIFF
--- a/pipeline_steps/influx.groovy
+++ b/pipeline_steps/influx.groovy
@@ -1,0 +1,71 @@
+def setup(){
+  instance_name = "INFLUX"
+  pubCloudSlave.getPubCloudSlave(instance_name: instance_name)
+  common.override_inventory()
+  try{
+    common.conditionalStage(
+      stage_name: "Influx",
+      stage:{
+        withCredentials([
+          file(
+            credentialsId: 'id_rsa_cloud10_jenkins_file',
+            variable: 'jenkins_ssh_privkey'
+          ),
+          string(
+            credentialsId: "SSH_IP_ADDRESS_WHITELIST",
+            variable: "SSH_IP_ADDRESS_WHITELIST"
+          ),
+        ]){
+          dir(WORKSPACE){
+            unstash "pubcloud_inventory"
+          }
+          dir('rpc-gating'){
+            git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
+            common.venvPlaybook(
+              playbooks: [
+                "playbooks/slave_security.yml",
+              ],
+              args: [
+                "-i ${env.WORKSPACE}/inventory",
+                "--limit job_nodes",
+                "--private-key=\"${env.JENKINS_SSH_PRIVKEY}\""
+              ],
+              vars: [
+                WORKSPACE: "${env.WORKSPACE}",
+              ]
+            )
+          }
+          dir('rpc-maas'){
+            git branch: env.RPC_MAAS_BRANCH, url: env.RPC_MAAS_REPO
+            sh """
+              export INVENTORY="${env.WORKSPACE}/inventory/hosts"
+              if ! grep log_hosts $INVENTORY; then
+                  echo [log_hosts:children] >> $INVENTORY
+                  echo job_nodes >> $INVENTORY
+              fi
+            """
+            // Run playbooks
+            common.venvPlaybook(
+              playbooks: [
+                "playbooks/maas-tigkstack-influxdb.yml",
+              ],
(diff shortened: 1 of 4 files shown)


pipeline_steps/influx.groovy
```diff
@@ -0,0 +1,67 @@
...
+      stage_name: "Influx",
+      stage:{
+        withCredentials([
+          file(
+            credentialsId: 'id_rsa_cloud10_jenkins_file',
+            variable: 'jenkins_ssh_privkey'
+          ),
+        ]){
+          dir('/tmp/'){
+            unstash "pubcloud_inventory"
```
hughsaunders: This will fail with `ERROR: No such saved stash ‘pubcloud_inventory’`

I don't think you need to worry about stashing with this job as all the tasks that don't run on the influx node will run on the same jenkins node. This is due to the top level node() call in the job definition.

playbooks/slave_open_ports.yml
```diff
@@ -0,0 +1,11 @@
+  - name: Secure Slave
```
hughsaunders: This should have influx in the file name, to show what its opening ports for.

rpc_jobs/long_running_infra.yml
```diff
@@ -0,0 +1,52 @@
+- job:
```
evrardjp: Ok will do.

pipeline_steps/influx.groovy
evrardjp: not sure what you meant here. I thought this line was staging the pubcloud_inventory [1]
I thought the playbooks were started from the jenkins node, therefore I had to give the inventory to connect to the influxdb node, and run my own playbooks.

[1]: https://github.com/rcbops/rpc-gating/blob/master/pipeline_steps/pubcloud.groovy#L33  and is called somewhere in the stack.

playbooks/slave_open_ports.yml
evrardjp: This is something re-usable, I thought it deserved a non-specific name.

pipeline_steps/influx.groovy
evrardjp: Should I change node() to node('CentOS') ?

hughsaunders: Whoops, you are correct, this stash is created in pubcloud.groovy and will exist. Apologies. You still don't need to use stash, but the alternative is copying out of workspace/rpc-gating/playbooks and I don't think thats any neater.

playbooks/slave_open_ports.yml
hughsaunders: I didn't notice that no actual ports are defined. Could we include this task in the slave security playbook but only execute if ports_whitelist is defined?

evrardjp: sure I thought it was good to leave that separately, one is for the basic host setup, the other was for firewalling. But fine for me to merge them.

hughsaunders: I don't have a strong opinion either way, but the slave_security playbook is already defining the base of the firewall, so this wouldn't be out of place there.

evrardjp: I'm fine for merging this with the other, but I thought it was good to keep it independant.

pipeline_steps/influx.groovy
```diff
@@ -0,0 +1,71 @@
...
+          file(
+            credentialsId: 'id_rsa_cloud10_jenkins_file',
+            variable: 'jenkins_ssh_privkey'
+          ),
+          string(
+            credentialsId: "SSH_IP_ADDRESS_WHITELIST",
+            variable: "SSH_IP_ADDRESS_WHITELIST"
+          ),
+        ]){
+          dir(${env.WORKSPACE}){
```
hughsaunders: string interpolation with out quotes won't work.

pipeline_steps/influx.groovy
```diff
@@ -0,0 +1,71 @@
...
+            unstash "pubcloud_inventory"
+          }
+          dir('rpc-gating/playbooks'){
+            git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
```
hughsaunders: This will checkout the rpc-gating repo into rpc-gating/playbooks. So the playbooks will be at rpc-gating/playbooks/playbooks. Same for maas below.

PR #285, opened 2017-07-05
De-dupe STAGES param

mattt416:
This commit de-dupes the STAGES param, allowing us to tweak specific
STAGES depending on action and trigger.

More specifically, with this change in place we can gate [1], which
allows us to remove the deployment of the rpc support role on PR
builds.

Note that `DEPLOY_SUPPORT_ROLE` will only be available once [1]
merges, however setting this in this PR saves us the extra PR once
[1] lands.

[1] https://github.com/rcbops/rpc-openstack/pull/2343

Issue: [UG-646](https://rpc-openstack.atlassian.net/browse/UG-646)

### DIFF
--- a/rpc_jobs/defaults.yml
+++ b/rpc_jobs/defaults.yml
@@ -8,6 +8,7 @@
     DEPLOY_SWIFT: "yes"
     DEPLOY_CEPH: "no"
     DEPLOY_ELK: "yes"
+    DEPLOY_SUPPORT_ROLE: "no"
     # Misc
     CRON: "H H(9-21) * * 1-5"
     CONTEXT_USER_VARS: ""
--- a/rpc_jobs/params.yml
+++ b/rpc_jobs/params.yml
@@ -110,6 +110,10 @@
           name: "DEPLOY_ELK"
           default: "{DEPLOY_ELK}"
           description: "Deploy ELK? yes/no"
+      - string:
+          name: "DEPLOY_SUPPORT_ROLE"
+          default: "{DEPLOY_SUPPORT_ROLE}"
+          description: "Deploy support role? yes/no"
       - text:
           name: "USER_VARS"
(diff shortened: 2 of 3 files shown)


mattt416:
Building https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_kilo-trusty-deploy-swift-periodic/20/ with `KIBANA_SELENIUM_BRANCH=liberty-12.2` and `STAGES="Allocate Resources, Connect Slave, Prepare Deployment, Deploy RPC w/ Script, Prepare Kibana Selenium, Kibana Tests, Cleanup, Destroy Slave"` to see if we can make the kibana stuff work on Kilo.  If so, we can remove the STAGES override for kilo series.

mattt416:
@odyssey4me @evrardjp still a WIP, but can you have a peep for me?

evrardjp:
It looks nice.

rpc_jobs/defaults.yml
```diff
@@ -8,6 +8,7 @@
     DEPLOY_SWIFT: "yes"
     DEPLOY_CEPH: "no"
     DEPLOY_ELK: "yes"
+    DEPLOY_SUPPORT_ROLE: "yes"
```
evrardjp: I'd rather not set it, or set it to the same value as the default in the rpc-o branch. Because this is not branch specific, and we plan to backport the disabling of the support role in all roles, we should probably set that to no.

mattt416: The idea here is to actually test the support role in periodic jobs.  I'm fine not testing it in PRs, but I think it's beneficial to at least test periodically.

evrardjp: I agree. You already switch to yes on L93 of rpc_aio though.

mattt416: Doh, good catch, thanks!

PR #286, opened 2017-07-05
UG-640 Add Jira Issue to linting

dankolbrs:
Adds lint step to verify Jira Issue included in commit. Required to
verify that commit attached to an existing Jira Issue.

Updates get_jira_issue_key function to only view last commit for
Jira issue key which can be included in either subject or body.

Adds git to Dockerfile which is required for new linter function.

Issue: [UG-640](https://rpc-openstack.atlassian.net/browse/UG-640)

Issue: [LA-320](https://rpc-openstack.atlassian.net/browse/LA-320)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -468,7 +468,7 @@ def is_doc_update_pr(git_dir) {
       def rc = sh(
         script: """#!/bin/bash
           set -xeu
-          git status 
+          git status
           git show --stat=400,400 | awk '/\\|/{print \$1}' \
             | egrep -v -e '.*md\$' -e '.*rst\$' -e '^releasenotes/'
         """,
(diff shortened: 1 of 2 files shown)


pipeline_steps/common.groovy
```diff
@@ -492,24 +492,24 @@ def is_doc_update_pr(git_dir) {
...
-    for (def i=0; i<=commit_titles.size(); i++){
-      try{
-        key = (commit_titles[i] =~ key_regex)[0]
-        print ("Found Jira Issue Key: ${key}")
-        return key
-      } catch (e){
-        continue
-      }
+      script: """
+      git remote add upstream https://github.com/${repo_org}/${repo_path}.git 2>&1 > /dev/null
```
hughsaunders: I would put this and the next line in github.add_issue_url_to_pr after the `git` call, then the repo clone, branch checkout and upstream addition are done in the same place. Add_issue_url_to_pr handles the repo setup, and common.get_jira_issue_key assumes the repo is correctly setup and diffs the two refs. It could take an upstream parameter that defaults to `upstream`.

pipeline_steps/common.groovy
```diff
@@ -492,24 +492,24 @@ def is_doc_update_pr(git_dir) {
...
+      git remote update 2>&1 > /dev/null
+      git log --pretty=%B upstream/${ghprbTargetBranch}..${ghprbSourceBranch}""")
+    print("looking for Jira issue keys in the following commits: ${commits}")
+    try{
```
hughsaunders: I was wondering where the for-loop went, but then I realised you're running the regex over the whole output, rather than a line at a time, which is probably faster and less code, bonus 👍

pipeline_steps/common.groovy
dankolbrs: Done.

dankolbrs:
Successfully finding issue in commit message and using out of date origin/master:
https://rpc.jenkins.cit.rackspace.net/job/RPC-Pull-Request-Issue-Link_rpc_gating/160/console

Failing due to not having issue in commit commit and using out of date origin/master:
https://rpc.jenkins.cit.rackspace.net/job/RPC-Pull-Request-Issue-Link_rpc_gating/158/

PR #289, opened 2017-07-06
LA-347 Pass variables to influxdb

evrardjp:
Influxdb deploy playbook needs variables to properly run.
One necessary variable is "influxdb_db_metric_password", which
is already included in rpc maas repo tests.


hughsaunders:
Superseded by #291

PR #290, opened 2017-07-06
LA-347 Skip host key checking

hughsaunders:
As gating playbooks are run from shared long running nodes, IPs will
gradually be added to the known hosts file. These IPs will be recycled
by public cloud and cause intermittent host key failures.


evrardjp:
Even if I don't like disabling security like this, we need this to go forward, short term.

PR #291, opened 2017-07-06
LA-347 Supply influx creds

hughsaunders:
As the influx db instance will be long running, it needs some
actual creds, rather than using publicly accessible defaults.


hughsaunders:
test build https://rpc.jenkins.cit.rackspace.net/job/Influx-Setup/15/console

PR #295, opened 2017-07-06
UG-666 Updates tempest results location

dankolbrs:
Master/Ocata places tempests result in a different path than pre-
Ocata. This updates command to attempt to find the location of
the >=Ocata `tempest_results.xml` file if exists and includes that
in path of results files that are attempted to be copied.

Issue: [UG-666](https://rpc-openstack.atlassian.net/browse/UG-666)

PR #299, opened 2017-07-07
[UG-665] Connect deploy node vm as a Jenkins slave

mkam:
This allows commands to be run directly
on the deploy node instead of having
an extra ssh step.

Issue: [UG-665](https://rpc-openstack.atlassian.net/browse/UG-665)

### DIFF
--- a/playbooks/setup_jenkins_slave.yml
+++ b/playbooks/setup_jenkins_slave.yml
@@ -13,6 +13,7 @@
     # These variables are not booleans so can use the standard default filter
     jenkins_node_labels: "{{ lookup('env', 'JENKINS_NODE_LABELS')|default('single_use_slave', true) }}"
     jenkins_node_executors: "{{ lookup('env', 'JENKINS_NODE_EXECUTORS')|default('2', true) }}"
+    jenkins_ssh_port: "{{ ansible_port | default(22, true) }}"
   roles:
     - role: willshersystems.sshd
       sshd:
@@ -131,6 +132,7 @@
           --creds "SSH Creds for Jenkins instances in public cloud." \
           --labels "{{ jenkins_node_labels }}" \
           --executors "{{ jenkins_node_executors }}" \
+          --port "{{ jenkins_ssh_port }}" \
           {% if jenkins_node_exclusive %}--exclusive{% endif %}
       register: jenkins_response
       until: jenkins_response|success
--- a/rpc_jobs/hw_multi_node_aio.yml
+++ b/rpc_jobs/hw_multi_node_aio.yml
@@ -115,11 +115,12 @@
               kibana = load 'pipeline_steps/kibana.groovy'
             }}
             dir("rpc-gating/playbooks") {{
-              writeFile file: "inventory/hosts", text: "[job_nodes]\n${{INSTANCE_NAME}} ansible_host=${{INSTANCE_IP}}\n"
+              writeFile file: "inventory/hosts", text: "[job_nodes]\n${{env.INSTANCE_NAME}} ansible_host=${{env.INSTANCE_IP}}\n"
               common.install_ansible()
             }}
             ssh_slave.connect()
-            node(INSTANCE_NAME) {{
+            deploy_node = null
+            node(env.INSTANCE_NAME) {{
               common.conditionalStage(
                 stage_name: 'Prepare HW AIO',
                 stage: {{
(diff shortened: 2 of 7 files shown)


mkam:
I didn't remove the ssh to VM code because I thought that should go in a follow up PR in case we still need the functionality or need to do a revert. 

I've had successful builds with master and mitaka, but one thing that isn't working is pulling out the tempest results file because they're in different locations for AIO and multi-node AIO. This doesn't affect whether the build passes or fails from tempest, but we won't have the test results report page. I'll try to update this once https://github.com/rcbops/rpc-gating/pull/295 merges.

rpc_jobs/hw_multi_node_aio.yml
```diff
@@ -129,8 +130,26 @@
...
+            ${{deploy_node}} ansible_port=2222 ansible_host=${{INSTANCE_IP}}
+            """
+            common.drop_inventory_file(inventory_content)
+            dir("rpc-gating/playbooks"){{
+              stash (
+                name: "pubcloud_inventory",
+                include: "inventory/hosts"
+              )
+            }}
+            ssh_slave.connect(2222)
```
hughsaunders: If you wrote both nodes to the same inventory, you'd only need to call ssh_slave.connect once? I like that you're connecting both the host and the VM as nodes, provides flexibility.

rpc_jobs/multi_node_aio.yml
```diff
@@ -186,6 +207,10 @@
         }} finally {{
             pubCloudSlave.delPubCloudSlave(instance_name: instance_name)
             maas.entity_cleanup(instance_name: instance_name)
+            if(deploy_node != null){{
+              instance_name = deploy_node
+              ssh_slave.destroy()
```
hughsaunders: Also destroy should take a name param, so we don't have to set a global to different values between two calls of the function.

rpc_jobs/hw_multi_node_aio.yml
```diff
@@ -139,21 +158,23 @@
...
+              maas.verify()
+              tempest.tempest()
+              kibana.kibana(env.KIBANA_SELENIUM_BRANCH)
+              holland.holland()
             }} // hw node
           }} catch (e){{
               print e
               throw e
           }} finally {{
+              ssh_slave.destroy()
```
hughsaunders: its not obvious that these two calls of ssh_slave.destroy() do different things as there are no different parameters. As noted in another comment, I think destroy should be refactored to take a name argument.

rpc_jobs/hw_multi_node_aio.yml
mkam: The VM creation is still being handled as a step that runs on the host server slave, so when the initial connect happens, the VM doesn't exist.

mkam:
Added the slave_name parameter to the destroy method and went ahead and moved the deploy node connection to its own method.

Run with the latest code: https://rpc.jenkins.cit.rackspace.net/job/mkam-ug-665/30/

rpc_jobs/hw_multi_node_aio.yml
```diff
@@ -139,21 +145,22 @@
...
+              holland.holland()
             }} // hw node
           }} catch (e){{
               print e
               throw e
           }} finally {{
+              ssh_slave.destroy(instance_name)
+              maas.entity_cleanup(instance_name: instance_name)
+              if(deploy_node != null){{
+                ssh_slave.destroy(deploy_node)
```
mattt416: Do we need to check if deploy_node != null?  Would that situation ever exist?

playbooks/setup_jenkins_slave.yml
```diff
@@ -13,6 +13,7 @@
     # These variables are not booleans so can use the standard default filter
     jenkins_node_labels: "{{ lookup('env', 'JENKINS_NODE_LABELS')|default('single_use_slave', true) }}"
     jenkins_node_executors: "{{ lookup('env', 'JENKINS_NODE_EXECUTORS')|default('2', true) }}"
+    jenkins_port: "{{ ansible_port | default(22, true) }}"
```
mattt416: The default kind of gives it away, but we could probably indicate w/ the variable name that this is the ssh port we're connecting into.

rpc_jobs/hw_multi_node_aio.yml
mkam: If the mnaio setup fails, then the deploy node as a slave won't exist, which causes this destroy to fail. It won't change the outcome of the build since it's already failed, but I thought if people saw that failure at a glance, they would think that was the original failure reason.

weezer:
LGTM but someone else needs to approve it.

PR #300, opened 2017-07-07
RE-28 Create jira issues for periodic failures

hughsaunders:
Issue: [RE-28](https://rpc-openstack.atlassian.net/browse/RE-28)

### DIFF
--- a/rpc_jobs/rpc_aio.yml
+++ b/rpc_jobs/rpc_aio.yml
@@ -348,6 +348,9 @@
                 print(e)
                 currentBuild.result = 'FAILURE'
                 throw e
+                if (TRIGGER == 'periodic'){{
+                  common.create_jira_issue(project="RO")
+                }}
               }} finally {{
                 common.archive_artifacts()
                 common.safe_jira_comment("${{currentBuild.result}}: [${{env.BUILD_TAG}}|${{env.BUILD_URL}}]")


mattt416:
Ready to get mah triage on!

mattt416:
It may be worth getting some consensus here before this merges -- ie. is everyone in RE team happy to be investigating these failures?  I suspect most won't be a result of gating-related failures, and should ultimately be investigated by the team in question.  Thoughts?

mkam:
@mattt416 It would be nice if the team that corresponds with the job gets the card, but if the failure is due to issues unrelated to that team, then I always worry people will start getting in the habit of ignoring the jira issues. Having them in RE first before transferring them to the appropriate team could help that. But if we don't think the teams ignoring the gate issues is going to be a problem then we don't have to be a middleman.

odyssey4me:
Personally I think the jobs should go to the disparate teams first, who can then pass it to us if there's an issue with the gating system. The idea is to spread the load, not push it through a bottleneck. If it turns out that gating failures are too common a cause, then we can figure out markers to identify and route those issues appropriately.

hughsaunders:
Which project should receive issues for RPCO periodic failures, RCM?

mattt416:
@hughsaunders I'm not sure -- doesn't sound like RCM based on their mission statement though.

hughsaunders:
Updated to use the [RO project](https://rpc-openstack.atlassian.net/projects/RO/issues).

claco:
I'm with @odyssey4me on this. In an ideal world, part of the act of adding a job to the gate is specifying who to notify when it fails (slack channel?). Maybe part of that can also be on which board to post failures? Require params for a job to be accepted or even run. Should do to the teams first, then back up to RE if it's an infra related issue (capacity, bad base image, networking issue, jenkins falls over, etc)

PR #303, opened 2017-07-11
RE-45 Test mechanisms on rpc-gating changes

hughsaunders:
This is a basic check to ensure that gating chagnes don't break a
sample build from each project. Currently only RPCO is included.

Issue: [RE-45](https://rpc-openstack.atlassian.net/browse/RE-45)

### DIFF
--- a/rpc_jobs/rpc_gating_project_builds.yml
+++ b/rpc_jobs/rpc_gating_project_builds.yml
@@ -1,0 +1,29 @@
+- job:
+    name: 'RPC-Gating-Project-Builds'
+    project-type: workflow
+    concurrent: true
+    triggers:
+      - github-pull-request:
+          org-list:
+            - rcbops
+          github-hooks: true
+          trigger-phrase: '.*recheck_(cit_)?all.*|.*recheck_(cit_)?project_builds.*'
+          only-trigger-phrase: false
+          auth-id: "github_account_rpc_jenkins_svc"
+          status-context: 'CIT/project_builds'
+          cancel-builds-on-update: true
+    properties:
+      - rpc-gating-github
+    dsl: |
+      stage("Instance Build Test"){
+        build(
+          job: "Single-Use-Slave-Example",
+          parameters: [
+            [
+              $class: 'StringParameterValue',
+              name: 'RPC_GATING_BRANCH',
+              value: ghprbSourceBranch
+            ],
+          ]
+        )
+      }
--- a/rpc_jobs/single_use_slave_template.yml
+++ b/rpc_jobs/single_use_slave_template.yml
@@ -7,7 +7,7 @@
       # params you want to override the defaults for.
       - instance_params:
           IMAGE: "Ubuntu 16.04 LTS (Xenial Xerus) (PVHVM)"
-          FLAVOR: "performance2-15"
+          FLAVOR: "performance1-1"
           REGION: "DFW"
       - rpc_gating_params
       - string:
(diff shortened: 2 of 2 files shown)


rpc_jobs/rpc_gating_project_builds.yml
```diff
@@ -0,0 +1,37 @@
...
+    concurrent: true
+    triggers:
+      - github-pull-request:
+          org-list:
+            - rcbops
+          github-hooks: true
+          trigger-phrase: '.*recheck_all.*|.*recheck_project_builds.*'
+          only-trigger-phrase: false
+          auth-id: "github_account_rpc_jenkins_svc"
+          status-context: 'CIT/lint'
```
mkam: This needs to be updated.

hughsaunders: Thanks, Done.

hughsaunders:
recheck_project_builds

hughsaunders:
See the `CIT/project_builds` check for a live test of this PR.

rpc_jobs/rpc_gating_project_builds.yml
```diff
@@ -0,0 +1,29 @@
+- job:
+    name: 'RPC-Gating-Project-Builds'
+    project-type: workflow
+    concurrent: true
+    triggers:
+      - github-pull-request:
+          org-list:
+            - rcbops
+          github-hooks: true
+          trigger-phrase: '.*recheck_all.*|.*recheck_project_builds.*'
```
evrardjp: I see you removed ``_cit_`` is there a reason? I like consistency...

rpc_jobs/rpc_gating_project_builds.yml
```diff
@@ -0,0 +1,29 @@
...
+          only-trigger-phrase: false
+          auth-id: "github_account_rpc_jenkins_svc"
+          status-context: 'CIT/project_builds'
```
evrardjp: shouldn't we do:
```white-list-target-branches: master```?

you said to me:
```If a branch isn’t listed, or isn’t matched by a pattern, then PRs to it will be ignored```.

File: rpc_jobs/single_use_slave_template.yml [hunk 4]
```diff
@@ -22,12 +22,34 @@
...
+            publishHTML(
+              alwaysLinkToLastBuild: true,
+              keepAll: true,
+              reportFiles: "artifacts.html",
+              reportName: "Build Artifact Links",
+              reportDir: "${WORKSPACE}"
+            )
+          }
+        }
+      }
```
evrardjp: Does this auto delete the node now?

Don't you need cleaning artifacts in the future? I agree, here we don't need it, because we just do ">", but it's maybe worth listing somewhere...

rpc_jobs/rpc_gating_project_builds.yml
hughsaunders: _cit_ was only added in rpc-openstack jobs to distinguish between old and new jenkins. rpc-gating has only ever had new jenkins jobs, so there's no need for a differentiator. Now that jenkins.propter is fully retired, _cit_ could be removed from the rpc-openstack jobs as well (or maybe make it optional).

rpc_jobs/rpc_gating_project_builds.yml
hughsaunders: Ahh, unless the whole clause is omitted in which case, PRs for all branches will be tested, which is fine for this repo, as all PRs are against master.

evrardjp: Ahah that was my initial question... English again! 👍  then.

rpc_jobs/single_use_slave_template.yml
hughsaunders: Workspaces are cleaned by the [use_node function](https://github.com/rcbops/rpc-gating/blob/master/pipeline_steps/common.groovy#L591), artifacts will expired out of public cloud in [30 days](https://github.com/rcbops/rpc-gating/blob/master/playbooks/upload_to_cloud_files.yml#L17). [runonpubcloud](https://github.com/rcbops/rpc-gating/blob/master/pipeline_steps/pubcloud.groovy#L118) Handles instance clean up.

rpc_jobs/rpc_gating_project_builds.yml
evrardjp: When I have to recheck, I look at the PR checks in github, and replace / or - with _. It generally works.

Here it won't, because recheck_cit_project_builds won't trigger.
But it's details.

evrardjp:
I don't disagree on the recheck. I am just waiting for an answer for readability.

rpc_jobs/rpc_gating_project_builds.yml
hughsaunders: I’ll add `(_cit_)?`

hughsaunders: Done.

odyssey4me:
I changed the title of the PR because it was confusing me. I've named it the same as the commit subject.

evrardjp:
👍

PR #308, opened 2017-07-11
UG-667 Remove code to run commands on vm

mkam:
Previously, we had to ssh onto the vm
and run commands for multi-node AIO.
Now that we can run directly on the deploy
node, this code is no longer needed.

Issue: [UG-667](https://rpc-openstack.atlassian.net/browse/UG-667)


mkam:
Didn't remove tempest vm code since that is being handled in https://github.com/rcbops/rpc-gating/pull/295.

Test Run:
https://rpc.jenkins.cit.rackspace.net/job/mkam-OnMetal-Multi-Node-AIO_master-xenial-periodic/6/

PR #312, opened 2017-07-12
UG-596 First pass at hardening job

mattt416:
jenkins-job-builder version 2.0.0.0b2 has a `credentials` parameter
type, which would work well for HARDENING_REPO, however as we're still
on 2.0.0.0b1 we cannot make HARDENING_REPO a parameter with a default
from a credential.

Issue: [UG-596](https://rpc-openstack.atlassian.net/browse/UG-596)

### DIFF
--- a/rpc_jobs/hardening.yml
+++ b/rpc_jobs/hardening.yml
@@ -1,0 +1,141 @@
+- project:
+    name: 'RPC-Hardening-Jobs'
+    series:
+      - master:
+          branch: master
+          USER_VARS: |
+            tempest_test_sets: 'scenario'
+    context:
+      - xenial:
+          DEFAULT_IMAGE: "16.04.2"
+    trigger:
+      - periodic:
+          CRON: "0 2 * * 7"
+    jobs:
+      - 'RPC-Hardening_{series}-{context}-{trigger}'
+
+- job-template:
+    name: 'RPC-Hardening_{series}-{context}-{trigger}'
+    project-type: workflow
+    properties:
+      - build-discarder:
+          days-to-keep: 7
+      - rpc-openstack-github
+    concurrent: true
+    triggers:
+      - timed: "{CRON}"
+    parameters:
+      - rpc_gating_params
+      - instance_params:
+          IMAGE: "OnMetal - Ubuntu 14.04 LTS (Trusty Tahr)"
+          FLAVOR: "onmetal-io1"
+          REGION: "IAD"
+      - rpc_repo_params:
+          RPC_BRANCH: "{branch}"
+      - osa_ops_params:
+          DEFAULT_IMAGE: "{DEFAULT_IMAGE}"
+          DATA_DISK_DEVICE: "sdb"
+      - text:
+          name: "USER_VARS"
+          default: "{USER_VARS}"
+          description: "OSA/RPC USER_VARS to inject for this build"
+      - string:
+          name: STAGES
+          default: |
+            Allocate Resources, Connect Slave, Prepare Multi-Node AIO,
+             Prepare RPC Configs, Deploy RPC w/ Script, Install Tempest,
+             Tempest Tests, Cleanup, Destroy Slave
+          description: |
+            Pipeline stages to run CSV. Note that this list does not influence execution order.
+            Options:
+              Allocate Resources
+              Connect Slave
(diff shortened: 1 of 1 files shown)


mattt416:
An initial test is being done here: https://rpc.jenkins.cit.rackspace.net/job/RPC-Hardening_master-xenial-periodic/8/

mattt416:
https://rpc.jenkins.cit.rackspace.net/job/RPC-Hardening_master-xenial-periodic/11/ passed -- this should be ready to go.

rpc_jobs/hardening.yml
```diff
@@ -0,0 +1,133 @@
...
+              Horizon Tests
+              Prepare Kibana Selenium
+              Kibana Tests
+              Holland
+              Pause (use to hold instance for investigation before cleanup)
+              Cleanup
+              Destroy Slave
+
+    dsl: |
+      node("CentOS"){{
```
hughsaunders: Why CentOS? I think CentOS is only needed for connecting the Jenkins slave which is already specified in ssh_slave.groovy.

rpc_jobs/hardening.yml
```diff
@@ -0,0 +1,133 @@
...
+              environment_vars: [
+                "DEPLOY_HAPROXY=yes",
+                "DEPLOY_ELK=yes",
+                "DEPLOY_TEMPEST=no",
+                "DEPLOY_AIO=no",
+                "DEPLOY_MAAS=no"
+                ]
+            ) // deploy_sh
+            unstash 'RPC-Unified-Gating'
+            dir("RPC-Unified-Gating"){{
```
hughsaunders: This should be wrapped in a stage, so that it's logs can be easily found.

rpc_jobs/hardening.yml
```diff
@@ -0,0 +1,133 @@
...
+            ssh_slave = load 'pipeline_steps/ssh_slave.groovy'
+            common = load 'pipeline_steps/common.groovy'
+            multi_node_aio_prepare = load 'pipeline_steps/multi_node_aio_prepare.groovy'
+            deploy = load 'pipeline_steps/deploy.groovy'
+            maas = load 'pipeline_steps/maas.groovy'
+          }} // dir
+
+          withCredentials([
+            string(
+              credentialsId: 'hardening_repo_url',
```
hughsaunders: I'd prefer a parameter rather than a credential. Is it a cred because its an internal url?

rpc_jobs/hardening.yml
hughsaunders: Answered via slack: Its for access to internal repo. However I think the use of the CentOS node could have a much narrower scope.

rpc_jobs/hardening.yml
hughsaunders: Answered via slack: yes.

rpc_jobs/hardening.yml
```diff
@@ -0,0 +1,133 @@
...
+            deploy.deploy_sh(
+              environment_vars: [
+                "DEPLOY_HAPROXY=yes",
+                "DEPLOY_ELK=yes",
+                "DEPLOY_TEMPEST=no",
+                "DEPLOY_AIO=no",
+                "DEPLOY_MAAS=no"
+                ]
+            ) // deploy_sh
+            unstash 'RPC-Unified-Gating'
```
dankolbrs: Why not run other tests while the node is up? While this would result in a longer run time, it could help to ensure the environment built as expected and results from the RPC-Unified-Gating tests were done on a good environment. Even a subset of Tempest such as just smoke tests could help to ensure valid results without the time of a full run.

dankolbrs:
Why not run other tests while the node is up? While this would result in a longer run time, it could help to ensure the environment built as expected and results from the RPC-Unified-Gating tests were done on a good environment. Even a subset of Tempest such as just smoke tests could help to ensure valid results without the time of a full run.

mattt416:
@dankolbrs yeah, I think you're right ... if you're going to deploy for so long we may as well validate that the instance is up.  I've kicked off another test here:

https://rpc.jenkins.cit.rackspace.net/job/RPC-Hardening_master-xenial-periodic/17/

If we find that the build is failing too often because of transient tempest failures then we can update it to ignore tempest failures (but then, is the environment actually in a good state?)

PR #313, opened 2017-07-13
LA-356 Allow SSH Whitelist to be disabled

hughsaunders:
This allows SSH logins from anywhere, which will be useful for debugging
during the early stages of influxdb setup.

Issue: [LA-356](https://rpc-openstack.atlassian.net/browse/LA-356)

### DIFF
--- a/rpc_jobs/influx.yml
+++ b/rpc_jobs/influx.yml
@@ -43,6 +43,12 @@
             hosts
             [log_hosts]
             node ansible_host=YOUR_IP_HERE
+      - bool:
+          name: "USE_SSH_WHITELIST"
+          description: |
+            Enable src restriction on SSH connections. Disabling allows
+            connections from anywhere. When enabled SSH connections will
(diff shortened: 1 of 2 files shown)


rpc_jobs/influx.yml
```diff
@@ -43,6 +43,11 @@
             hosts
             [log_hosts]
             node ansible_host=YOUR_IP_HERE
+      - boolean:
+          name: "USE_SSH_WHITELIST"
+          description: |
+            With this enabled, the fireall will restrict SSH connections to
```
evrardjp: firewall

evrardjp: I'm not sure to understand the description, vs what it does.

hughsaunders:
@evrardjp I don't think a firewall addition is needed, as it's already there: https://github.com/rcbops/rpc-gating/blob/3c27b76703d668b92187db01e782c72866431c33/playbooks/slave_security.yml#L25-L29

PR #315, opened 2017-07-13
Delete kilo leapfrog periodic

mattt416:
The RPC-AIO_kilo-trusty-leapfrogupgrade-swift-periodic job is failing
constantly because we're not setting UPGRADE_FROM_REF appropriately for
this job.  This commit deletes this job since this job is effectively
testing the same thing as
RPC-AIO_newton141-trusty-leapfrogupgrade-swift-periodic.

### DIFF
--- a/rpc_jobs/rpc_aio.yml
+++ b/rpc_jobs/rpc_aio.yml
@@ -148,6 +148,12 @@
         action: leapfrogupgrade
       - series: master
         action: leapfrogupgrade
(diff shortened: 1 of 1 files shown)


mattt416:
Testing this [here](https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_kilo-trusty-leapfrogupgrade-swift-periodic/11/console).  So far it's gotten further than it previously did.

PR #316, opened 2017-07-14
RE-49 Remove get-pip from create_workspace_venv

hughsaunders:
This is to avoid installing the latest setup tools, which doesn't
downgrade cleanly.

A larger refactor is required, but this is a workaround to unblock.

Issue: [RE-49](https://rpc-openstack.atlassian.net/browse/RE-49)


hughsaunders:
Test job: https://rpc.jenkins.cit.rackspace.net/view/AIO/job/RPC-IRR_rpc-maas-master-xenial-master/120/console

pipeline_steps/common.groovy
```diff
@@ -30,13 +30,6 @@ def void create_workspace_venv(){
     # UG-613 change TMPDIR to directory with more space
     export TMPDIR="/var/lib/jenkins/tmp"
 
-    # UG-612 Install pip, setuptools, and wheel on the host
```
odyssey4me: By removing this you have an unknown version of pip on the host which may, or may not, understand the --constraint option. This has been the case before and broke many things, so I think this is likely to make it worse, not better.

hughsaunders: As long as we avoid the centos slaves, the version of pip is known as it is installed in the long running slave deploy process: https://github.com/rcbops/rpc-gating/blob/master/playbooks/setup_jenkins_slave.yml#L101-L118

evrardjp: 👍

odyssey4me: @hughsaunders But https://github.com/rcbops/rpc-gating/blob/master/playbooks/setup_jenkins_slave.yml#L126 is installing an unknown version of setuptools on the slave managing the setup of the new slave.

hughsaunders: Currently centos slaves are not allocated unless explicitly requested. The only time we request centos slaves is when we need to access to the Jenkins REST API (eg for connecting a new slave).

evrardjp: in a venv though

hughsaunders: Also the centos slaves have pip 7.1, the first version that includes constraints.

odyssey4me: OK, in that case I'd suggest that https://github.com/rcbops/rpc-gating/blob/master/playbooks/setup_jenkins_slave.yml#L126 changes to be ``pip install pip setuptools wheel -c rpc-gating/constraints.txt``

hughsaunders: Removed both the pip install lines from the last task in setup_jenkins_slave, as thats covered by create_workspace_venv, which is called before setup_jenkins_slave is run.

PR #320, opened 2017-07-14
RE-49 Always recreate venv

hughsaunders:
Using the old venv may create problems when versions have changed.

Issue: [RE-49](https://rpc-openstack.atlassian.net/browse/RE-49)


hughsaunders:
This is a consequence of  #316

PR #332, opened 2017-07-18
RE-48 Balance instances across regions

hughsaunders:
Make use of the full capacity by building AIOs in all regions.

Issue: [RE-48](https://rpc-openstack.atlassian.net/browse/RE-48)


hughsaunders:
Updated with new approach.

playbooks/allocate_pubcloud.yml
```diff
@@ -2,63 +2,21 @@
...
-
-    - name: Wait for SSH to be available on all hosts
-      wait_for: port=22 host="{{ item.accessIPv4 }}"
-      with_items: "{{ rax.success }}"
-
-    - name: Wait an additional amount of time for any remaning startup tasks to finish
-      pause:
-        minutes: 5
+    - include: tasks/boot_instance_attempt.yml
+      with_items: "{{ region_retries }}"
```
mattt416: Could you add the `when` here, to avoid having to suffix each task in `boot_instance_attempt.yml` with it?

playbooks/allocate_pubcloud.yml
```diff
@@ -2,63 +2,21 @@
...
+    # When booting succeeds, boot_success is set to true so that
+    # all subsequent attempts are skipped.
 
-    - name: Fail if the required number of instances aren't available
-      fail:
-        msg: "At least one public cloud instance failed to start :("
-      when: rax.success|length < count
+    - name: Create region retries list
+      set_fact:
+        region_retries: "{{ region.split(',')|map('trim')|shuffle * 5 }}"
```
evrardjp: region_retries: "{{ 5 * region | shuffle }}" ?

playbooks/tasks/boot_instance_attempt.yml
```diff
@@ -0,0 +1,104 @@
...
+    name: "{{ name }}"
+    flavor: "{{ flavor }}"
+    image: "{{ image }}"
+    count: "{{ count }}"
+    key_name: "{{ keyname }}"
+    region: "{{ item }}"
+    wait: yes
+    wait_timeout: 900
+    auto_increment: no
+  delegate_to: localhost
```
evrardjp: I'd prefer if we had a feature to build from a random region from a list of regions , instead of building this.

playbooks/tasks/boot_instance_attempt.yml
```diff
@@ -0,0 +1,104 @@
...
+      {% for instance in rax.success %}
+      {{instance.name}} ansible_host={{instance.accessIPv4}} ansible_user=root
+      {% endfor %}
+    dest: '{{lookup("env", "WORKSPACE")}}/rpc-gating/playbooks/inventory/hosts'
+  when:
+    - not boot_success
+    - attempt_success
+
+- name: Wait for SSH to be available on all hosts
+  wait_for: port=22 host="{{ item.accessIPv4 }}"
```
evrardjp: not the standard YAML syntax we use to see. 👁  pain.

rpc_jobs/defaults.yml
```diff
@@ -3,7 +3,7 @@
     name: global
     IMAGE: "Ubuntu 14.04 LTS (Trusty Tahr) (PVHVM)"
     FLAVOR: "performance2-15"
-    REGION: "DFW"
+    REGION: "DFW, ORD"
```
evrardjp: if only two regions, why not block and rescue ?

hughsaunders:
Third approach. This version uses with_items on a single task as the loop. It also stores the chosen region in the inventory so that cleanup can remove the instance from the correct region.

hughsaunders:
Before this merges we need to ensure that keys and images are available in all the regions.

rpc_jobs/defaults.yml
mattt416: I think you'd need to make `all_regions` a parameter also because we can only build onmetal v1 in IAD.  For the mnaio jobs, you don't want it falling back to the other regions currently defined in `all_regions`.

odyssey4me:
I think we'll need to rebase and resurrect this to try and improve our capacity for tests as we're getting a lot of failures. For now we have two regions. I'll get to work on automating the image build process so that we can use more regions, but for now I'd say let's go with two.

git-harry:
@hughsaunders what happening here? Do you need any help with this?

git-harry:
Nevermind, I see it's already blocked in Jira.

hughsaunders:
recheck_project_builds

hughsaunders:
recheck_project_builds

hughsaunders:
recheck_project_builds

hughsaunders:
recheck_project_builds

hughsaunders:
recheck_project_builds

hughsaunders:
recheck_project_builds

hughsaunders:
recheck_project_builds

hughsaunders:
recheck_project_builds

hughsaunders:
recheck_project_builds

hughsaunders:
recheck_project_builds

hughsaunders:
recheck_project_builds

hughsaunders:
recheck_project_builds

hughsaunders:
recheck_project_builds

hughsaunders:
recheck_project_builds

hughsaunders:
Originally this patch reduced the wait time for cloud instances from 5 mins to 1 min, but that caused a high change of disconnection a couple of minutes after a job started using the instance. It ires me, but I had to reinstate the 5 minute delay. 

Ready for review

odyssey4me:
@hughsaunders I noticed recently that nodepool specifies an API timeout (1 minute), and a boot timeout (2 minutes). Perhaps we should work something like that in?

hughsaunders:
@odyssey4me as discussed in slack, this problem occurs after the instance has achieved ACTIVE status and SSH is available. So while those timeouts are useful, I don't think they'll help in this case.

playbooks/cleanup_pubcloud.yml
```diff
@@ -1,15 +1,14 @@
 ---
-- hosts: localhost
+- hosts: job_nodes
```
odyssey4me: Doing it this way will require a successful connection to the target host, even though it runs on tasks on there it still has to connect.

I don't know of a local connection will work in this case.

Also, what happens if the host never got created and therefore the connection attempt fails causing the playbook to fail.

hughsaunders: Currently, if an instance fails to build the cleanup will fail so that isn't much of a change. I don't think its entirely unreasonable either. No instance, so it can't be cleaned up. 

We could possibly improve that situation by making cleanup test for an inventory file and deciding that no inventory file means no instances so don't try to cleanup. 

One case we discussed in person is cleaning up instances that have been shutdown as a result of image creation. That case will fail. As discussed that mechnaism will likely change in  the near future when we implement node pool, so it may not be worth fixing. However we will need to ensure that the periodic clenaup job regex will catch those instances.

rpc_jobs/params.yml
```diff
@@ -5,6 +5,14 @@
       - string:
           name: REGION
```
git-harry: I think it would be worth changing this to `REGIONS`, the current names hides that this is a list.

playbooks/allocate_pubcloud.yml
```diff
@@ -4,40 +4,61 @@
...
-      debug:
-        var: rax
+    # Region should be a CSV string.
+    # Primary regions are those specified in the region var,
+    # they are tried first in a random order. After that,
+    # fallback regions are tried in a random order.
+
+    - name: Create primary region list
+      set_fact:
+        regions: "{{ region.split(',')|map('trim')|map('upper')|shuffle * instance_boot_retries }}"
```
git-harry: `instance_boot_retries` seems a bit misleading. If I have `region="first"` and `fallback_regions="second,third"` with `instance_boot_retries=3` that results in the following rax module task getting `["first", "first", "first", "second", "third"]` so in fact there are 4 retries or 5 attempts in total.

git-harry: You should remove `""` from the list otherwise there are a few situations that will cause that to be one of the regions tried e.g. when `region` ends with a `,`.

playbooks/allocate_pubcloud.yml
```diff
@@ -4,40 +4,61 @@
...
+
+    - name: Create fallback region list
+      set_fact:
+        fallback_regions_shuff: "{{ fallback_regions.split(',')|map('trim')|map('upper')|difference(regions)|shuffle }}"
```
git-harry: A bit of a nit, but the use of `"_shuff"` in the name seems unnecessary given both `regions` and `fallback_regions_shuff` represent shuffled lists.

playbooks/allocate_pubcloud.yml
hughsaunders: renamed to primary_region_attempts, with the comment "Each region in the primary regions list will be tried this number of times."

hughsaunders: Done

playbooks/allocate_pubcloud.yml
hughsaunders: `fallback_regions` already exists so I had to call it something else. 
`region --> regions`
`fallback_regions --> fallback_regions_shuf`

hughsaunders:
@git-harry, added ORD to cleanup in https://github.com/rcbops/rpc-gating/pull/435/files other regions can be added as required.

rpc_jobs/params.yml
hughsaunders: Changed to REGIONS, required changes to other jobs.

git-harry:
@hughsaunders that change gives a list of `"DFW IAD ORD"` for the periodic cleanup whereas this patch specifies `"DFW, IAD, ORD, HKG, SYD"` in multiple places.

playbooks/allocate_pubcloud.yml
hughsaunders: Decided to change `REGION --> REGIONS` so now we have:
```
regions --> regions_shuf
fallback_regions --> fallback_regions_shuf
```

playbooks/allocate_pubcloud.yml
hughsaunders: Added select() to reject empty strings.

hughsaunders:
@git-harry added HKG, SYD to cleanup, also fixed inconsistency between periodic_cleanup.py and allocate_pubcloud.yml about whether REGIONS is space or comma delimited. Now either is acceptable by both.

project_builds failed because the single use slave example job needs to be rebuilt to work with this change, will rebuild it then retry.

hughsaunders:
recheck_project_builds

playbooks/roles/maas_entities/tasks/main.yml
```diff
@@ -6,7 +6,7 @@
       api_key: "{{ lookup('env', 'PUBCLOUD_API_KEY') }}"
       state: "{{ state }}"
       label: "{{ item }}"
-      region: "{{ lookup('env', 'REGION') }}"
+      region: "{{ lookup('env', 'REGIONS') }}"
```
git-harry: What happens if `REGIONS` is a list or a the region used was taken from `FALLBACK_REGIONS`?

hughsaunders: Addressed this and save_pubcloud.yml by using rax_region from the generated static inventory.

hughsaunders:
Project builds failed because it requires a rebuild of single-use-slave-example, but it doesn't test any of the things that changed since it last passed, so leaving it.

hughsaunders:
Test jobs rebuilt from RE-48 and and using RE-48 as the rpc_gating branch:
[RPC-AIO_newton-xenial-deploy-swift-periodic](https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_newton-xenial-deploy-swift-periodic/149/) ▶️ 
[OnMetal-Multi-Node-AIO_mitaka-trusty-deploy-none-periodic](https://rpc.jenkins.cit.rackspace.net/view/Multi-Node%20AIO/job/OnMetal-Multi-Node-AIO_mitaka-trusty-deploy-none-periodic/64/) ▶️

hughsaunders:
https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_newton-xenial-deploy-swift-periodic/151/ ✅ 
https://rpc.jenkins.cit.rackspace.net/view/Multi-Node%20AIO/job/OnMetal-Multi-Node-AIO_mitaka-trusty-deploy-none-periodic/67/console ▶️

hughsaunders:
https://rpc.jenkins.cit.rackspace.net/view/Multi-Node%20AIO/job/OnMetal-Multi-Node-AIO_mitaka-trusty-deploy-none-periodic/68 ▶️

hughsaunders:
- [OnMetal-Multi-Node-AIO_mitaka-trusty-deploy-none-periodic](https://rpc.jenkins.cit.rackspace.net/view/Multi-Node%20AIO/job/OnMetal-Multi-Node-AIO_mitaka-trusty-deploy-none-periodic/69/) ▶️

hughsaunders:
- [OnMetal-Multi-Node-AIO_mitaka-trusty-deploy-none-periodic](https://rpc.jenkins.cit.rackspace.net/view/Multi-Node%20AIO/job/OnMetal-Multi-Node-AIO_mitaka-trusty-deploy-none-periodic/70) ❌ 

Failed in the same place but in a different way, would be too coincidental to be unrelated, but retrying. 
```
+ ansible-playbook -i deploy_vm_inventory.3cb7 --limit job_nodes --extra-vars=ansible_port=2222 --private-key=**** -e@vars.setup_jenkins_slave.yml.1b8 setup_jenkins_slave.yml

PLAY [job_nodes] ***************************************************************

TASK [setup] *******************************************************************
fatal: [omnamtdnp-70-f633-infra1-vm]: FAILED! => {"changed": false, "failed": true, "module_stderr": "Shared connection to 104.130.18.22 closed.\r\n", "module_stdout": "/usr/bin/python: can't open file '/root/.ansible/tmp/ansible-tmp-1506694716.87-52234770486655/setup.py': [Errno 2] No such file or directory\r\n", "msg": "MODULE FAILURE"}
	to retry, use: --limit @/var/lib/jenkins/workspace/OnMetal-Multi-Node-AIO_mitaka-trusty-deploy-none-periodic/rpc-gating/playbooks/setup_jenkins_slave.retry

omnamtdnp-70-f633-infra1-vm : ok=0    changed=0    unreachable=0    failed=1  
```

hughsaunders:
- [OnMetal-Multi-Node-AIO_mitaka-trusty-deploy-none-periodic](https://rpc.jenkins.cit.rackspace.net/view/Multi-Node%20AIO/job/OnMetal-Multi-Node-AIO_mitaka-trusty-deploy-none-periodic/71/console) ▶️

hughsaunders:
- [OnMetal-Multi-Node-AIO_mitaka-trusty-deploy-none-periodic](https://rpc.jenkins.cit.rackspace.net/view/Multi-Node%20AIO/job/OnMetal-Multi-Node-AIO_mitaka-trusty-deploy-none-periodic/72/console) ▶️

hughsaunders:
The MNAIO job failed on apt/ssh. That issue is being worked on in #453. The previous build got to tempest which is way past the instance build phase of the job. So this is ready for review.

hughsaunders:
@git-harry updated the commit message

PR #334, opened 2017-07-19
LA-373 Deploy Maas with an env var

evrardjp:
Maas deploy or not should be defined with a single env var.
We shouldn't have steps for maas, the orchestration inside the
project (for example rpc-o leapfrog, or rpc-o deploy) should
handle its deploy.
Gating should only be aware of what to do in a job, not handling
the job's work itself.

Issue: [LA-373](https://rpc-openstack.atlassian.net/browse/LA-373)

### DIFF
--- a/rpc_jobs/params.yml
+++ b/rpc_jobs/params.yml
@@ -114,6 +114,10 @@
           name: "DEPLOY_SUPPORT_ROLE"
           default: "{DEPLOY_SUPPORT_ROLE}"
           description: "Deploy support role? yes/no"
+      - string:
+          name: "DEPLOY_MAAS"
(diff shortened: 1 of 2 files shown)


evrardjp:
Depends on https://github.com/rcbops/rpc-openstack/pull/2399

evrardjp:
For L: https://github.com/rcbops/rpc-openstack/pull/2397 
For M: https://github.com/rcbops/rpc-openstack/pull/2395

evrardjp:
I agree, but if we are short on time, we should merge this. Should, in theory, fix leapfrog metrics gathering.

evrardjp:
I mislinked. Everything should be alright.

PR #336, opened 2017-07-19
LA-373 Env var for maas should not be overriden

evrardjp:
The env var set for maas was overriden into the job, wh
We remove the override, and all the references to maas for the
AIO job.

Issue: [LA-373](https://rpc-openstack.atlassian.net/browse/LA-373)

### DIFF
--- a/rpc_jobs/rpc_aio.yml
+++ b/rpc_jobs/rpc_aio.yml
@@ -252,8 +252,6 @@
               Connect Slave
               Prepare Deployment
               Deploy RPC w/ Script
-              Setup MaaS
-              Verify MaaS
               Install Tempest
               Tempest Tests
               Prepare Kibana Selenium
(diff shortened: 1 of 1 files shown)


rpc_jobs/rpc_aio.yml
```diff
@@ -346,7 +343,6 @@
                 environment_vars = [
                   "DEPLOY_HAPROXY=yes",
                   "DEPLOY_AIO=no",
-                  "DEPLOY_MAAS=no",
```
git-harry: I think this should be:
```
"DEPLOY_MAAS=${{DEPLOY_MAAS}}"
```

hughsaunders: DEPLOY_MAAS is already an environment variable because it is a parameter, so it doesn't need to be in this array. I don't think this array is needed anyway. Why pass env vars around, why not set them directly?

hughsaunders:
@git-harry The rest of this cleanup is part of https://rpc-openstack.atlassian.net/browse/RE-87 which is about having a single entry point for RPC testing. Other cards in that epic may also be related.

rpc_jobs/rpc_aio.yml
git-harry: Learn something new every day.

PR #331, opened 2017-07-17
UG-646 Common node allocation

hughsaunders:
This pull request ensures that MaaS variables are supplied to the rpc deploy process, so that it can configure MaaS for the deployment.

The majority of this change is dedicated to moving node allocation to a common function. This ensures that nodes are allocated in a consistent way, and that the gating utility virtualenv is build exactly once per node.

This PR also prepares rpc-gating for use as a Jenkins shared library, so that we can use implicitly library loading, rather than having to explicitly load every groovy script in every job.

Issue: [UG-646](https://rpc-openstack.atlassian.net/browse/UG-646)

PR #337, opened 2017-07-19
Load maas script to get token and URL

mkam:
If https://github.com/rcbops/rpc-gating/pull/331 merges first, than this PR is not necessary.


mkam:
Prepare deployment stage passed: https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/mkam-RPC-AIO/detail/mkam-RPC-AIO/2/pipeline

PR #340, opened 2017-07-20
UG-668 Create issues in RE project by default

hughsaunders:
MaaS alert JIRA issues are being created in the UG project which is
now legacy. This commit updates the defaults to ensure that new issues
are created in the RE project.

Issue: [UG-668](https://rpc-openstack.atlassian.net/browse/UG-668)


hughsaunders:
Please ping me (slack preferably) when this goes in, as I'll have to update the live webhook translator.

PR #345, opened 2017-07-24
UG-646 add library statement to every job

hughsaunders:
Currently the rpc-gating library is loaded implicitly, jenkins injects
it into the groovy environment before the script is started. The
problem with this approach is that the verison of rpc-gating used
cannot be overridden.

This version uses the RPC_GATING_BRANCH parameter to set the
version of rpc-gating that is loaded. This will return our ability
to test other jobs (such as RPC AIOs) against changes to gating.

This approach also makes it explicit in each job where the library
is coming from. Once this merges, we'll turn off implicit loading
in Jenkins global config.

Issue: [UG-646](https://rpc-openstack.atlassian.net/browse/UG-646)

### DIFF
--- a/rpc_jobs/periodic_cleanup.yml
+++ b/rpc_jobs/periodic_cleanup.yml
@@ -73,7 +73,7 @@
         // common.use_node/shared_slave/internal_slave.
         container.inside {
           stage("Docker Checkout"){
-            git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
+            git branch: env.RPC_GATING_BRANCH, url: "https://github.com/rcbops/rpc-gating"
           }
           stage("Public Cloud Cleanup"){
             withCredentials([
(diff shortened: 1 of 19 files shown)


hughsaunders:
@mattt416 removed RPC_GATING_REPO

rpc_jobs/periodic_cleanup.yml
```diff
@@ -72,7 +73,7 @@
         // common.use_node/shared_slave/internal_slave.
         container.inside {
           stage("Docker Checkout"){
-            git branch: env.RPC_GATING_BRANCH, url: env.RPC_GATING_REPO
+            git branch: env.RPC_GATING_BRANCH, url: common.RPC_GATING_REPO
```
mattt416: Is this here right?

hughsaunders: Nope, fixed. Was experimenting with constants in the common module, but forgot to remove that one.

PR #347, opened 2017-07-24
Change double bracket to single for non-template jobs

mkam:


mkam:
recheck_lint The lint job had the error this fixes in it, so I updated the job manually.

PR #348, opened 2017-07-24
RE-105 Use instance_name argument for deletion

mkam:
The instance_name was being passed to the delete
methods but global variable instance_name was
being used instead.

Issue: [RE-105](https://rpc-openstack.atlassian.net/browse/RE-105)

### DIFF
--- a/pipeline_steps/pubcloud.groovy
+++ b/pipeline_steps/pubcloud.groovy
@@ -55,7 +55,7 @@ def cleanup(Map args){
               "--private-key=\"${env.JENKINS_SSH_PRIVKEY}\"",
             ],
             vars: [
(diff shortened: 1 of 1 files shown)


mkam:
The delete was successful, though the rest of the job failed because it was only a partial deployment:

https://rpc.jenkins.cit.rackspace.net/job/mkam-OnMetal-Multi-Node-AIO_mitaka-trusty-deploy-none-periodic/3/console

```
PLAY [localhost] ***************************************************************

TASK [Get instance info] *******************************************************
ok: [localhost]

TASK [Cleanup public cloud instances] ******************************************
changed: [localhost]

localhost                  : ok=2    changed=1    unreachable=0    failed=0   

```

PR #350, opened 2017-07-26
RE-103 Store rpc-gating venvs

hughsaunders:
RPC-Gating currently builds a venv in the workspace of every node
that is used as part of a job. This is useful as we can rely on the
right version of various tools to be available, however it is network
intensive and slow. This commit reduces the time taken by using a pre
built venv when one is available.

* Adds a job to build a tar containing the rpc-gating venv and ansible
  roles. This tar is pushed to rpc-repo, and named with the SHA of
  rpc-gating it was built with. The most recent ten tars are kept.
  This job is triggered by pushes to rpc-gating, so new versions
  will automatically built as changes are merged.

* Adds functionality to common.install_ansible to check for a venv tar
  if one exists, it is downloaded, expanded and fixed. Otherwise the
  venv is built locally.

  Jobs won't be blocked between a change merging and the venv tar being
  available on the repo servers, as they will fall back to creating the
  venv locally.

Issue: [RE-103](https://rpc-openstack.atlassian.net/browse/RE-103)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -115,7 +115,7 @@ def venvPlaybook(Map args){
         write_json(file: vars_file, obj: args.vars)
         sh """#!/bin/bash -x
           which scl && source /opt/rh/python27/enable
-          . ${env.WORKSPACE}/.venv/bin/activate
+          set +x; . ${env.WORKSPACE}/.venv/bin/activate; set -x
           export ANSIBLE_HOST_KEY_CHECKING=False
           ansible-playbook ${args.args.join(' ')} -e@${vars_file} ${playbook}
         """
@@ -545,7 +545,7 @@ def create_jira_issue(project="RE", tag=env.BUILD_TAG, link=env.BUILD_URL, type=
   ]){
     sh """#!/bin/bash -xe
       cd ${env.WORKSPACE}
-      . .venv/bin/activate
+      set +x; . .venv/bin/activate; set -x
       python rpc-gating/scripts/jirautils.py create_issue\
         --tag '$tag'\
         --link '$link'\
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -1,71 +1,94 @@
 import groovy.json.JsonSlurperClassic
 import groovy.json.JsonOutput
 
-def void create_workspace_venv(){
-  print "create_workspace_venv"
-  sh """#!/bin/bash -xe
+def void install_ansible(){
+  print "install_ansible"
+  sh """#!/bin/bash -e
+    REPO_BASE="https://rpc-repo.rackspace.com/rpcgating/venvs"
     cd ${env.WORKSPACE}
 
-    # Create venv
-    if [[ ! -d ".venv" ]]; then
-      requirements="virtualenv==15.1.0"
-      pip install -U "\${requirements}" \
-        || pip install --isolated -U "\${requirements}"
-      if which scl
-      then
-        # redhat/centos
-        source /opt/rh/python27/enable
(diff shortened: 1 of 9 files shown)


hughsaunders:
recheck_project_builds

hughsaunders:
recheck_project_builds

hughsaunders:
The project_build job is using the workspace venv artefact as added in this PR: https://rpc.jenkins.cit.rackspace.net/job/Single-Use-Slave-Example/98/console

rpc_jobs/build_gating_venv.yml
```diff
@@ -0,0 +1,45 @@
...
+            pushd rpc-gating
+              SHA=\$(git rev-parse HEAD)
+            popd
+            archive="rpcgatingvenv_\${SHA}.tbz"
+            set +u
+            . .venv/bin/activate
+            virtualenv --relocatable .venv
+            deactivate
+            set -u
+            tar cjfp \$archive .venv rpc-gating/playbooks/roles
```
hughsaunders: As thats where the roles stored. Expanding them from the tar will be faster than clone and extract via galaxy.

rpc_jobs/build_gating_venv.yml
```diff
@@ -0,0 +1,45 @@
...
+              || echo "\${REPO_HOST} \$(cat \$REPO_HOST_PUBKEY)" \
+              >> ~/.ssh/known_hosts
+
+            REPO_PATH="/var/www/repo/rpcgating/venvs"
+
+            # Upload generated version
+            scp -i \$REPO_USER_KEY \$archive \$REPO_USER@\$REPO_HOST:\$REPO_PATH
+
+            # Generate index
+            ssh -i \$REPO_USER_KEY \$REPO_USER@\$REPO_HOST "cd \$REPO_PATH; ls -1 > index"
```
mattt416: Do you want to limit rpcgatingvenv_*.tbz to the index?  Otherwise, it puts the index itself in the index.

pipeline_steps/common.groovy
```diff
@@ -1,71 +1,88 @@
...
+    }
+
+    download_venv(){
+      curl -s "\${REPO_BASE}/rpcgatingvenv_\${SHA}.tbz" > venv.tbz
+      tar xjfp venv.tbz
+      # virtualenv --relocatable doesn't update the activate script FGJ.
+      sed -i.bak 's|VIRTUAL_ENV=.*\$|VIRTUAL_ENV=${env.WORKSPACE}/.venv|' .venv/bin/activate
+      # CentOS 6 can take a hike, its glibc isn't new enough for python 2.7.12
+      if which scl; then
+        cp /opt/rh/python27/root/usr/bin/python .venv/bin/python
```
mattt416: I wonder if we're going to run into issues shoving the python binary into a venv that was built with a different version of python?

rpc_jobs/build_gating_venv.yml
evrardjp: Fair enough to store roles as artifacts but it's confusing to mix them together, for me.
I think these are two different kind of artifacts: venvs != ansible roles.
If everybody agrees on it, fine for me.

pipeline_steps/common.groovy
evrardjp: Isn't that what we are doing now already?
Let's try and see what happens? In any case we'll probably have to iterate there.

hughsaunders:
recheck_project_builds

hughsaunders:
@mattt416 I agree, and I'll probably open up the webhook proxy a little so that we can create slaves without using centos nodes, but thats a [separate issue](https://rpc-openstack.atlassian.net/browse/RE-112).

hughsaunders:
Build using stored venv: https://rpc.jenkins.cit.rackspace.net/job/Single-Use-Slave-Example/128/console

PR #354, opened 2017-07-28
Add pins for unpinned roles

hughsaunders:
We were discussing unpinned things in slack, which reminded me that this needed doing.


hughsaunders:
Venv build against the pinned versions: https://rpc.jenkins.cit.rackspace.net/job/Build-Gating-Venv/56/console

PR #355, opened 2017-07-28
RE-133 Revise artifact build job

odyssey4me:
Currently the pipelines, if set to replace
existing artifacts, will double-do the apt/git
artifacts which is unnecessary.

The container artifacts for multiple
distributions cannot be done at the same time
due to a race condition when uploading.

This patch therefore takes away the image
and context axes from the jobs and forces the
jobs to run in series or parallel as
appropriate.


rpc_jobs/rpc_artifact_build.yml
```diff
@@ -131,43 +79,135 @@
...
+
+          // We can run all the artifact build processes in parallel
+          // for PR's because they do not upload anything.
+          if (env.trigger == "pr") {{
+            parallel(
+              "apt": {{
+                // We do not bother ensuring that the long lived
+                // node the the published data set are equal for
+                // PR's because we never publish the resulting
+                // artifacts after PR jobs.
```
hughsaunders: This doesn't quite make sense, also `the the`

odyssey4me:
@mattt416 

This is largely a change to manage a bigger issue which is that whenever newton moves to a new tag, the artifact build process has to be executed manually and very carefully. This ensures that the entire pipeline executes properly, and @hughsaunders and I couldn't really see another way to do the pipeline to produce reliable results.

1. The PR failures for these jobs thus far have yielded no false failures. They have always related to the change in question as far as I know. That said, if it becomes a burden we could probably switch this to operating as two projects instead of one so that PR's can be individually rechecked - unless there's another way.

2. The expected visualisation to use here would be the blueocean view - I don't think the traditional view would work at all due to the parallelism.

3. It's not been tested, but there's nothing going on at the moment which requires these jobs to be working as-is, except current PR's to master/newton. If you'd prefer to implement some sort of test I'm game, although I have no idea how to.

rpc_jobs/rpc_artifact_build.yml
```diff
@@ -131,43 +79,136 @@
...
+            ) // parallel
+          }} else {{
+            // When the job is periodic, the jobs must run in a
+            // particular sequence in order to ensure that each
+            // artifact set is built and uploaded so that it can
+            // be used in the step that follows.
+
+            // The apt artifacts are built on a long-lived slave
+            // with a single executor, so no locking is required.
+            artifact_build.apt()
```
hughsaunders: I think @mattt416 has a point about stages. For PRs, each build is in a labeled block as each parallel section has a label. However in the periodic/serialised build, we've removed the conditional stages and not added any fixed stages. Adding these would be helpful for finding the correct sections of logs, and correctly labeling the blue ocean flow chart. 

All four blocks in the periodic section should be wrapped in a stage. Stage should be the outermost call.

PR #357, opened 2017-07-31
RE-143 Separate Clone and checkout steps

hughsaunders:
A repo can only be cloned at a ref that has been fetched, as there
is no tag for the SHA we are using, it can't be fetched directly.

This commit clones master, then checks out the required SHA as a
second step.

Issue: [RE-143](https://rpc-openstack.atlassian.net/browse/RE-143)


hughsaunders:
@git-harry I'm not sure the problem needs a more general solution, that repo is only used for mnaio, and other checkouts use refs rather than direct SHAs. I don't want to spread the concept of clone ref vs checkout ref any further than we have to.

git-harry:
@hughsaunders yes they use refs in the same way that multi-node used a ref, until it didn't. So the next time we have to pin a SHA the same thing will happen. Something will get merged without anyone realising and things will break.

PR #358, opened 2017-07-31
RE-145 Move mnaio params to params.yml

mattt416:
Currently, jobs that use pipeline_steps/multi_node_aio_prepare.groovy
are failing if they do not have the COMPUTE_NODES and VOLUME_NODES
parameters defined.  Since these parameters are necessary for anything
using pipeline_steps/multi_node_aio_prepare.groovy, we add the params
to rpc_jobs/params.yml instead.

Issue: [RE-145](https://rpc-openstack.atlassian.net/browse/RE-145)

### DIFF
--- a/rpc_jobs/hardening.yml
+++ b/rpc_jobs/hardening.yml
@@ -35,6 +35,8 @@
       - osa_ops_params:
           DEFAULT_IMAGE: "{DEFAULT_IMAGE}"
           DATA_DISK_DEVICE: "sdb"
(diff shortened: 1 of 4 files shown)


mattt416:
Not managed to test this yet as I'm currently unable to build mnaio instances.  :(

mattt416:
Hi @hughsaunders , these params have already been defaulted in defaults.yml.  See https://github.com/rcbops/rpc-gating/blob/master/rpc_jobs/defaults.yml#L29-L30.

--Matt

PR #359, opened 2017-08-01
LA-372 Influx reporting in gating tooling

evrardjp:
Gating could ship this small tool to query influx db based
on env vars. When running a job, we could check for the presence
of the tool, and run it. That would output results to stdout.

On top of that, we could run the tool with parameters to have
artefacts (YAML/subunit) of the data.

Issue: [LA-372](https://rpc-openstack.atlassian.net/browse/LA-372)

### DIFF
--- a/influx-reports/influx-report.sh
+++ b/influx-reports/influx-report.sh
@@ -1,0 +1,6 @@
+PYTHON_PATH="${WORKSPACE}/.venv/bin/python"
+REPORTS_FOLDER="${WORKSPACE}/test_results"
+INFLUX_DOWNTIME_FOLDER="$(dirname "${BASH_SOURCE[0]}")"
+
+mkdir -p "${REPORTS_FOLDER}"
+$PYTHON_PATH $INFLUX_DOWNTIME_FOLDER/influx.py --ymlreport $REPORTS_FOLDER/influx-uptime.yml --subunitreport $REPORTS_FOLDER/influx-uptime.bin
--- a/influx-reports/influx.py
+++ b/influx-reports/influx.py
@@ -1,0 +1,238 @@
+#!/usr/bin/env python3
+""" Get all measurements for an influx job, and
+outputs a report of uptime for maas_ components.
+Inputs:
+    - env.INFLUX_IP (mandatory)
+    - env.INFLUX_PORT (mandatory)
+    - env.BUILD_TAG (mandatory)
+    - ymlreport <ymlreportfile> (optional)
+    - subunitreport <subunitreport> (optional)
+Outputs:
+    - stdout
+    - yaml file with uptime (when ymlreport is used)
+    - subunit binary file with passing/failure criteria
+    (when subunitreport is used)
+"""
+
+import argparse
+from collections import defaultdict
+import datetime
+import logging
+import os
+import math
+import sys
+import yaml
+
+import dateutil.parser
+from influxdb import InfluxDBClient
+from subunit.v2 import StreamResultToBytes
+
+
+class InfluxTimestampParseException(Exception):
+    pass
+
+
+class SubunitContext():
+    """Context manager for writing subunit results."""
+
+    def __init__(self, output_path):
+        self.output_path = output_path
+
+    def __enter__(self):
+        self.output_stream = open(self.output_path, 'wb+')
+        self.output = StreamResultToBytes(self.output_stream)
+        self.output.startTestRun()
+        return self.output
+
+    def __exit__(self, *args, **kwargs):
+        self.output.stopTestRun()
+        self.output_stream.close()
+
+    def status(self, *args, **kwargs):
+        self.output.status(*args, **kwargs)
+
+
+def return_time(client, query, delta_seconds=0):
+    """ From an InfluxDB query, fetch
+    the first point time, and return a
+    python time object. Shift it from
+    a few seconds (delta_seconds) if necessary.
+    """
+    timestamp_query = client.query(query)
+    # Get points is generator, we should just get first
+    # point time (string type).
+    fpt_str = next(timestamp_query.get_points())['time']
+    try:
+        fpt = dateutil.parser.parse(
+            fpt_str) + datetime.timedelta(seconds=delta_seconds)
+    except Exception as alle:
+        raise InfluxTimestampParseException(
+            "Error parsing a timestamp from influx: {}".format(alle))
+    return fpt
+
+
+def calculate_measurement_uptime(client, measurement, job_reference,
+                                 measurement_period_seconds,
+                                 resolution=60):
+    """ For a certain build (job_reference), estimate the amount
+    of seconds a component (column of table) was up.
+    For that, we slice the job by resolution.
+    If I have no data in the slice, I assume downtime. If I have
+    at least one data per slice, I assume up, even if degraded.
+    The count of these positive events multiplied by resolution
+    is an average idea of the uptime.
+    """
+    per_field_up_percent = defaultdict(float)
+
+    query = (
+        "select max(/.*_status/) from {measurement} "
+        "where time < now() and job_reference='{job_reference}' "
+        "group by time({resolution}s) fill(-1)"
+    ).format(measurement=measurement,
+             job_reference=job_reference,
+             resolution=resolution)
+
+    max_amount_of_time_slices = math.ceil(
+        float(measurement_period_seconds) / resolution)
+
+    all_data = client.query(query)
+    for time_slice in all_data.get_points():
+        for element in time_slice:
+            if element != 'time' and time_slice[element] == 1:
+                key = element.replace('max_', '').replace('_status', '')
+                # Mark the field as up
+                per_field_up_percent[key] += (1 / max_amount_of_time_slices)
+                # Mark the measurement as up
+
+    return dict(
+        per_field_uptime_percent=dict(per_field_up_percent)
+    )
+
+
+def build_report_dict(client, measurements, job_reference,
+                      measurement_period_seconds):
+    report = defaultdict(dict)
+    for measurement in measurements:
+        measurement_data = calculate_measurement_uptime(
+            client, measurement, job_reference, measurement_period_seconds)
+        report['runtime_seconds'] = measurement_period_seconds
+        report['per_field_uptime_percent'][measurement] = \
+            dict((measure, round(measure_uptime * 100, 2)) for
+                 measure, measure_uptime in
+                 measurement_data['per_field_uptime_percent'].items())
+    return dict(report)
+
+
+def main(args=None):
+    client = InfluxDBClient(args.influx_ip, args.influx_port,
+                            database='telegraf')
+    """
+    Influx structure:
+        Measurement (eg maas_nova)
+            Point (a timestamp with data)
+                Tags (Indexed K/V, eg job_reference)
+                Fields (Unindexed K/V, eg keystone_user_count)
+    """
+    # First find the first and last timestamp from telegraf.
+    # This way querying maas_* data will always
+    # be accurate (if no data is reported by maas plugins,
+    # this should be a failure metric)
+    find_first_timestamp_query = (
+        "select first(total) from processes "
+        "where job_reference = '{}';".format(args.job_reference)
+    )
+    first_ts = return_time(client, find_first_timestamp_query, delta_seconds=5)
+    find_last_timestamp_query = (
+        "select last(total) from processes "
+        "where job_reference = '{}';".format(args.job_reference))
+    last_ts = return_time(client, find_last_timestamp_query, delta_seconds=-5)
+
+    measurement_period_seconds = (last_ts - first_ts).total_seconds()
+
+    logging.info(
+        ("Metrics were gathered between {first} and {last}\n"
+         "For a total of {seconds} seconds".format(
+             first=first_ts,
+             last=last_ts,
+             seconds=measurement_period_seconds)))
+
+    measurements = ['maas_glance', 'maas_cinder', 'maas_keystone',
+                    'maas_heat', 'maas_neutron', 'maas_nova', 'maas_horizon']
+
+    report = build_report_dict(
+        client, measurements, args.job_reference, measurement_period_seconds)
+
+    yml_report = yaml.safe_dump(report, default_flow_style=False)
+    logging.info(yml_report)
+    if args.ymlreport:
+        with open(args.ymlreport, 'w+') as output_file:
+            output_file.write("---\n")
+            output_file.write(yml_report)
+
+    # Subunit is a unit test result format. Here we output a stream of
+    # test results, one per measurement.
+    # Each measurement is a successfull test if its overall uptime %age
+    # is higher than args.min_uptime
+    if args.subunitreport:
+        with SubunitContext(args.subunitreport) as output:
+            for measurement_name, measurement_data in \
+                    report['per_field_uptime_percent'].items():
+                for measure_name, uptime_percentage in \
+                        measurement_data.items():
+                    status = "fail"
+                    if uptime_percentage > args.min_uptime:
+                        status = "success"
+                    # Record test start
+                    output.status(
+                        test_id=measure_name,
+                        timestamp=first_ts
+                    )
+
+                    # Record end of test
+                    output.status(
+                        test_id=measure_name,
+                        # TODO(hughsaunders): Be more intelligent
+                        # about thresholds
+                        test_status=status,
+                        test_tags=None,
+                        runnable=False,
+                        file_name=measure_name,
+                        file_bytes=" {}% lower than threshold".format(
+                            uptime_percentage).encode('ascii'),
+                        timestamp=last_ts,
+                        eof=True,
+                        mime_type='text/plain; charset=UTF8'
+                    )
+
+
+if __name__ == "__main__":
+    """ Args parser and logic router """
+    logging.getLogger().setLevel(logging.INFO)
+    parser = argparse.ArgumentParser(
+        description='Fetch maas_ metrics, and report downtime data.')
+    parser.add_argument("--ymlreport", help="Yaml report filename")
+    parser.add_argument("--subunitreport", help="Subunit report filename")
+    parser.add_argument(
+        "--min-uptime",
+        help="Minimum uptime required (percentage)",
+        default=1,
+        type=float)
+    arguments = parser.parse_args()
+    try:
+        arguments.influx_ip = os.environ['INFLUX_IP']
+    except KeyError:
+        logging.error("Please set INFLUX_IP")
+        sys.exit(1)
+
+    try:
+        arguments.influx_port = os.environ['INFLUX_PORT']
+    except KeyError:
+        logging.error("Please set INFLUX_PORT")
+        sys.exit(2)
+
+    try:
+        arguments.job_reference = os.environ['BUILD_TAG']
+    except KeyError:
+        logging.error("Please set BUILD_TAG for its usage as job ref")
+        sys.exit(3)
+    main(arguments)
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,12 +3,14 @@ ansible
 appdirs
 click
 github3.py
+influxdb
 jenkins-job-builder
 jenkinsapi
 jira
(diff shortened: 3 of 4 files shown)


evrardjp:
WIP Do not merge. I still need to think about this.

influx-reports/influx.py
```diff
@@ -0,0 +1,140 @@
...
+      logging.error("Please set INFLUX_PORT")
+      sys.exit(2)
+
+    try:
+      job_reference = os.environ['BUILD_TAG']
+    except KeyError:
+      logging.error("Please set BUILD_TAG for its usage as job ref")
+      sys.exit(3)
+
+    client = InfluxDBClient(influx_ip, influx_port, database='telegraf')
```
hughsaunders: No Auth?

influx-reports/influx.py
```diff
@@ -0,0 +1,140 @@
...
+    of seconds a component (column of table) was up.
+    For that, we slice the job by resolution.
+    If I have no data in the slice, I assume downtime. If I have
+    at least one data per slice, I assume up, even if degraded.
+    The count of these positive events multiplied by resolution
+    is an average idea of the uptime.
+    """
+    uptimes = defaultdict(int)
+
+    query = ("select max(/.*_status/) from {} where time < now() and "
```
hughsaunders: I'd prefer multiple formats in the same string to be named

influx-reports/influx.py
```diff
@@ -0,0 +1,140 @@
...
+    a few seconds (delta_seconds) if necessary.
+    """
+    timestamp_query = client.query(query)
+    # Get points is generator, we should just get first
+    # point time (string type).
+    fpt_str = next(timestamp_query.get_points())['time']
+    try:
+        fpt = dateutil.parser.parse(fpt_str) + datetime.timedelta(seconds=delta_seconds)
+    except Exception as e:
+        raise SystemExit(e)
```
hughsaunders: Why escalate the exception to SystemExit, when an uncaught ValueError or TypeError will halt execution?

influx-reports/influx.py
```diff
@@ -0,0 +1,140 @@
...
+                # Input status
+                output.status(
+                    test_id=measure,
+                    timestamp=first_ts
+                )
+
+                # Output the end of the event
+                output.status(
+                    test_id=measure,
+                    # TODO(evrardjp): Define a Threshold to compare with
```
hughsaunders: I realise this is WIP, but the threshold must be included or the subunit stream isn't useful.

influx-reports/influx.py
```diff
@@ -0,0 +1,140 @@
...
+      output_file.write("---\n")
+    if args.subunitreport:
+      subunit_report = True
+      output_stream = open(args.subunitreport, 'wb')
+      output = StreamResultToBytes(output_stream)
+      output.startTestRun()
+
+    for measurement in measurements:
+        measurement_data = return_uptime(client, measurement, job_reference)
+        yml_output = yaml.dump({measurement: dict(measurement_data)}, default_flow_style=False)
```
hughsaunders: I wondered why you wrap a defaultdict in dict, but its because yaml.safe_dump can't represent a defaultdict. Talking of which, you should use safe_dump.

influx-reports/influx.py
```diff
@@ -0,0 +1,140 @@
...
+
+    logging.info("Metrics were gathered between {} and {}".format(first_ts, last_ts))
+
+    measurements = ['maas_glance','maas_cinder','maas_keystone', 'maas_heat', 'maas_neutron',
+                    'maas_nova', 'maas_horizon']
+
+    yml_report = False
+    subunit_report = False
+
+    if args.ymlreport:
```
hughsaunders: I'd wrap this whole section in try/finally to ensure `output.stopTestRun` gets called. You could use context managers and exitstack, seeming as you're in py3.  https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack

influx-reports/influx.py
```diff
@@ -0,0 +1,140 @@
...
+                )
+    if subunit_report:
+        output.stopTestRun()
+        output_stream.close()
+    if yml_report:
+        output_file.close()
+
+if __name__ == "__main__":
+    logging.getLogger().setLevel(logging.INFO)
+    parser = argparse.ArgumentParser(description='Fetch maas_ metrics, and report dowtime data.')
```
hughsaunders: downtime

requirements.txt
```diff
@@ -3,12 +3,14 @@ ansible
...
 click
 github3.py
+influxdb==4.1.1
 jenkins-job-builder
 jenkinsapi
 jira
 packaging
 pyrax
 python-dateutil
+python-subunit
```
hughsaunders: constraint should be added also

influx-reports/influx-report.sh
```diff
@@ -0,0 +1,6 @@
+PYTHON_PATH="${WORKSPACE}/.venv/bin/python"
+REPORTS_FOLDER="~/workspace/test_results"
```
hughsaunders: This should be `${WORKSPACE}/test_results`

influx-reports/influx.py
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env python3
+import argparse
+import datetime
+import dateutil.parser
```
hughsaunders: I don't think this is standard lib, so should be added to requirements and constraints

influx-reports/influx.py
```diff
@@ -0,0 +1,140 @@
...
+        measurement_data = return_uptime(client, measurement, job_reference)
+        yml_output = yaml.dump({measurement: dict(measurement_data)}, default_flow_style=False)
+        logging.info(yml_output)
+        if yml_report:
+            output_file.write(yml_output)
+        if subunit_report:
+            for measure in measurement_data:
+                # Input status
+                output.status(
+                    test_id=measure,
```
hughsaunders: This test name needs to be more specific. `downtime.glance_api` for example?

requirements.txt
```diff
@@ -3,12 +3,14 @@ ansible
 appdirs
 click
 github3.py
+influxdb==4.1.1
```
evrardjp: Good catch

requirements.txt
evrardjp: In the next patchset.

influx-reports/influx.py
evrardjp: Already in requirements.

influx-reports/influx.py
evrardjp: I'm ok if you change it :)

influx-reports/influx.py
evrardjp: not sure what you meant there.

influx-reports/influx.py
```diff
@@ -0,0 +1,140 @@
...
+    uptimes = defaultdict(int)
+
+    query = ("select max(/.*_status/) from {} where time < now() and "
+             "job_reference='{}' group by time({}s) fill(-1)"
+            ).format(table,job_reference,resolution)
+    all_data = client.query(query)
+    for time_slice in all_data.get_points():
+       for element in time_slice:
+           if element != 'time' and time_slice[element] == 1:
+               uptimes[element.replace('max_','')] += (1 * resolution)
```
evrardjp: can replace to resolution.

influx-reports/influx.py
evrardjp: None for now.

influx-reports/influx.py
evrardjp: I could use context managers indeed, but then I have to indent differently depending on what we do... It sounded more complex than just closing.

influx-reports/influx.py
evrardjp: Done.

influx-reports/influx.py
evrardjp: measure's value would be specific. Not sure what you meant there.

influx-reports/influx.py
evrardjp: It isn't useful for now. For this to be useful, we need to:
- add code to find, in a known database of thresholds, the ``measure`` metric threshold.
- Compare this result to the ``measurement_data[measure]`` and output fail/success.

We don't have the code, nor the thresholds yet. That's why TODO.

influx-reports/influx.py
git-harry: I believe @hughsaunders is referring to the replacement fields `{}`, they are currently unnamed. For example `"{} - {named}".format("I have no name", named="I have a name.")`.

evrardjp: oh yeah. Fine for me. @hughsaunders will submit a patch next, so I am fine with that.

evrardjp:
Well, doc of the module is not that bad. However, you should have all the details to run this yourself. Let me ping you that in private.

influx-reports/influx.py
```diff
@@ -0,0 +1,166 @@
...
+        output_stream.close()
+    if yml_report:
+        output_file.close()
+
+
+if __name__ == "__main__":
+    """ Args parser and logic router """
+    logging.getLogger().setLevel(logging.INFO)
+    parser = argparse.ArgumentParser(
+        description='Fetch maas_ metrics, and report downtime data.')
```
git-harry: The file docstring refers to uptime, this description refers to downtime. More generally, the script help text and the YAML output file provide little/no information regarding what the output actually is.

The following is part of a YAML report file I've generated:
```
---
maas_glance:
  glance_api_local_status: 1740
  glance_registry_local_status: 1800
maas_cinder:
  cinder-backup_status: 180
  cinder-scheduler_status: 240
  cinder-volume-lvm_status: 180
  cinder_api_local_status: 1740
```

What do the numbers mean? What are the units? I think it would be useful to include a comment at the top of the YAML file explaining what the data is.

evrardjp: That is fairly valid. It's the seconds of uptime in seconds. Maybe we can rename the keys to glance_api_local_uptime...

git-harry: If these numbers represent uptime in seconds, it would be useful to know what the maximum possible uptime was or the percentage of uptime. Without the knowing the period of time measured it's difficult to know whether or not there is an issue.

evrardjp: Indeed, the maximum possible uptime should already be there. If not, it's just that I forgot it.
For the issue or not should be also handled when we'll have thresholds defined.

influx-reports/influx.py
hughsaunders: Done

influx-reports/influx.py
hughsaunders: Done

influx-reports/influx.py
hughsaunders: Oh whoops, I misread that as a literal.

influx-reports/influx.py
```diff
@@ -0,0 +1,248 @@
...
+                        measurement_period_seconds) * 100
+
+    # total_seconds is the measurement period multiplied by the number of
+    # series. This is used to create an average percent for the measurement.
+    total_seconds = len(per_field_downtime_seconds) \
+        * measurement_period_seconds
+    total_downtime = sum(per_field_downtime_seconds.values())
+    measurement_downtime_seconds = total_downtime \
+        / len(per_field_downtime_seconds)
+    measurement_downtime_percent = \
```
git-harry: The numbers I'm getting when I generate a report look a bit off, for example heat was down for over 7000% of the time! I'm a misunderstanding the numbers?
```---
measurement_downtime_percent:
  maas_cinder: 14.661654135338345
  maas_glance: 44.3609022556391
  maas_heat: 7233.082706766918
  maas_horizon: 39.097744360902254
  maas_keystone: 43.609022556390975
  maas_neutron: 40.100250626566414
  maas_nova: 2739.6616541353387
```

influx-reports/influx.py
```diff
@@ -162,5 +222,27 @@ def main(args=None):
         description='Fetch maas_ metrics, and report downtime data.')
     parser.add_argument("--ymlreport", help="Yaml report filename")
     parser.add_argument("--subunitreport", help="Subunit report filename")
+    parser.add_argument(
+        "--max-downtime",
+        help="Maximum allowable downtime per service (seconds)",
+        default=60*60,
```
evrardjp: 1 hour? Seems high.
Please write: 3600s :p

influx-reports/influx.py
```diff
@@ -53,93 +89,122 @@ def return_uptime(client, table, job_reference, resolution=60):
...
-        logging.info(yml_output)
-        if yml_report:
-            output_file.write(yml_output)
-        if subunit_report:
-            for measure in measurement_data:
-                # Input status
+        with SubunitContext(args.subunitreport) as output:
+            for measurement_name, downtime_seconds in \
+                    report['measurement_downtime_seconds'].items():
+                status = "success"
```
evrardjp: status = fail, and reverse the condition would be safer.

evrardjp:
Hugh's refactor broke subunit, so I have to work on it.

influx-reports/influx.py
```diff
@@ -0,0 +1,252 @@
...
+        "select first(total) from processes "
+        "where job_reference = '{}';".format(args.job_reference)
+    )
+    first_ts = return_time(client, find_first_timestamp_query, delta_seconds=5)
+    find_last_timestamp_query = (
+        "select last(total) from processes "
+        "where job_reference = '{}';".format(args.job_reference))
+    last_ts = return_time(client, find_last_timestamp_query, delta_seconds=-5)
+
+    measurement_period_seconds = (last_ts - first_ts).total_seconds()
```
git-harry: It would be nice to include this in the reports.

evrardjp:
I am rewriting this percentage, and remove the seconds thing.
It doesn't bring anything good.

influx-reports/influx.py
```diff
@@ -0,0 +1,252 @@
...
+def build_report_dict(client, measurements, job_reference,
+                      measurement_period_seconds):
+    report = defaultdict(dict)
+    for measurement in measurements:
+        measurement_data = calculate_measurement_uptime(
+            client, measurement, job_reference, measurement_period_seconds)
+        report['measurement_uptime_seconds'][measurement] = \
+            int(round(measurement_data['measurement_uptime_seconds'], 0))
+        report['measurement_uptime_percent'][measurement] = \
+            round(measurement_data['measurement_uptime_percent'], 5)
```
evrardjp: One zero too much, good catch.

influx-reports/influx.py
```diff
@@ -0,0 +1,252 @@
...
+    at least one data per slice, I assume up, even if degraded.
+    The count of these positive events multiplied by resolution
+    is an average idea of the uptime.
+    """
+    per_field_uptime_seconds = defaultdict(int)
+    per_field_uptime_percent = {}
+
+    query = (
+        "select max(/.*_status/) from {measurement} "
+        "where time < now() and job_reference='{job_reference}' "
```
evrardjp: There should be no filler data, because this would run effectively directly in the job.

influx-reports/influx.py
```diff
@@ -0,0 +1,252 @@
...
+            fpt_str) + datetime.timedelta(seconds=delta_seconds)
+    except Exception as alle:
+        raise InfluxTimestampParseException(
+            "Error parsing a timestamp from influx: {}".format(alle))
+    return fpt
+
+
+def calculate_measurement_uptime(client, measurement, job_reference,
+                                 measurement_period_seconds,
+                                 resolution=60):
```
evrardjp: Points are happening not one per 60 seconds, but one per 30 seconds with jitter.
Having a large enough resolution (2*sampling) should be good enough.

Minutes can be used as unit, but we used seconds everywhere else in our variables.
It makes sense to re-use the same unit.

evrardjp:
@git-harry You can have a look at this version.

influx-reports/influx.py
```diff
@@ -0,0 +1,241 @@
...
+                key = element.replace('max_', '').replace('_status','')
+                # Mark the field as up
+                per_field_up_percent[key] += (1 / max_amount_of_time_slices)
+                # Mark the measurement as up
+                measurement_was_up = True
+        if measurement_was_up:
+            measurement_uptime_percent += (1 / max_amount_of_time_slices)
+
+    return dict(
+        measurement_uptime_percent=measurement_uptime_percent,
```
git-harry: The mean uptime of a services components does not give any useful measure of the availability of the service, I think it should be removed.

influx-reports/influx.py
```diff
@@ -0,0 +1,241 @@
...
+                      measurement_period_seconds):
+    report = defaultdict(dict)
+    for measurement in measurements:
+        measurement_data = calculate_measurement_uptime(
+            client, measurement, job_reference, measurement_period_seconds)
+        report['runtime_seconds'] = measurement_period_seconds
+        report['measurement_uptime_percent'][measurement] = \
+            round(measurement_data['measurement_uptime_percent'] * 100, 4)
+        report['per_field_uptime_percent'][measurement] = \
+            dict((measure, round(measure_uptime * 100,4)) for measure, measure_uptime in measurement_data['per_field_uptime_percent'].items())
```
git-harry: At the very most round to two decimal places, even that is probably of no use but anything more is sub-second accuracy.

influx-reports/influx.py
git-harry: I'm not sure we're talking about the same thing but given the changes to the report, this is now less of an issue.

evrardjp:
@git-harry ready to review.

git-harry:
@evrardjp this is failing the lint check.

influx-reports/influx.py
```diff
@@ -0,0 +1,238 @@
...
+        yml_report = yaml.safe_dump(report, default_flow_style=False)
+        logging.info(yml_report)
+        with open(args.ymlreport, 'w+') as output_file:
+            output_file.write("---\n")
+            output_file.write(yml_report)
+
+    # Subunit is a unit test result format. Here we output a stream of
+    # test results, one per measurement.
+    # Each measurement is a successfull test if its overall uptime %age
+    # is higher than args.min_uptime
```
git-harry: There is no `args.min_uptime`, I assume it was removed in a refactor.

influx-reports/influx-report.sh
```diff
@@ -0,0 +1,6 @@
+PYTHON_PATH="${WORKSPACE}/.venv/bin/python"
+REPORTS_FOLDER="${WORKSPACE}/test_results"
+INFLUX_DOWNTIME_FOLDER="$(dirname "${BASH_SOURCE[0]}")"
```
git-harry: The output data gives figures for uptime, the use of the term downtime in this shell script might be misleading.

influx-reports/influx.py
```diff
@@ -0,0 +1,238 @@
...
+    if args.subunitreport:
+        with SubunitContext(args.subunitreport) as output:
+            for measurement_name, measurement_data in \
+                    report['per_field_uptime_percent'].items():
+                for measure_name, uptime_percentage in \
+                        measurement_data.items():
+                    status = "fail"
+                    if uptime_percentage > args.max_downtime:
```
git-harry: This check is wrong. `uptime_percentage` would need to be less than 1 for the status to show a failure.

evrardjp:
@git-harry Modified.

influx-reports/influx.py
```diff
@@ -0,0 +1,238 @@
...
+    """ Args parser and logic router """
+    logging.getLogger().setLevel(logging.INFO)
+    parser = argparse.ArgumentParser(
+        description='Fetch maas_ metrics, and report downtime data.')
+    parser.add_argument("--ymlreport", help="Yaml report filename")
+    parser.add_argument("--subunitreport", help="Subunit report filename")
+    parser.add_argument(
+        "--min-uptime",
+        help="Minimum uptime required (percentage)",
+        default=1,
```
git-harry: Is this going to be used to fail the build? It seems a bit low if so.

evrardjp: We have about 0% right now, just a little more. I guess it's just as a starter.

evrardjp:
squashed.

evrardjp:
@hughsaunders I'll merge it as is, this way LA-412 can continue forward with just rebasing.

PR #360, opened 2017-08-01
RE-148 Remove newton-14.0 branch from gating

odyssey4me:
The newton-14.0 branch for rpc-openstack has been
removed, so the jenkins jobs for it should also be
removed.

Issue: [RE-148](https://rpc-openstack.atlassian.net/browse/RE-148)


mattt416:
recheck_all

PR #361, opened 2017-08-01
RE-133 runonpubcloud doesn't take an image param

hughsaunders:
Set the IMAGE env var instead.


hughsaunders:
First approach was https://github.com/rcbops/rpc-gating/compare/RE-133_pubcloud_instance_param but that was getting complicated, so decided to stash it and go with this version for now.

PR #362, opened 2017-08-01
RE-151 Add gating for newton branch

mattt416:
Issue: [RE-151](https://rpc-openstack.atlassian.net/browse/RE-151)


mattt416:
@odyssey4me do we need to update rpc_artifact_build.yml here?

odyssey4me:
recheck_cit_all

odyssey4me:
@mattt416 Yes, we will - but I can put in a patch for that.

PR #365, opened 2017-08-01
[WIP] Expand UPGRADE_FROM_REF

mattt416:
The newton-14.1 branch, for example, can do a minor upgrade and a
leapfrog upgrade.  Having a single UPGRADE_FROM_REF makes it difficult
to correctly set the required initial branch.

This commit adds {MAJOR,MINOR,LEAPFROG}_UPGRADE_FROM_REF and sets
UPGRADE_FROM_REF to the relevant value depending on the action.

Issue: newton-14

### DIFF
--- a/pipeline_steps/aio_prepare.groovy
+++ b/pipeline_steps/aio_prepare.groovy
@@ -2,7 +2,9 @@ def prepare(){
   common.conditionalStage(
     stage_name: "Prepare Deployment",
     stage: {
-      if (env.STAGES.contains("Major Upgrade") || env.STAGES.contains("Leapfrog Upgrade")) {
+      if (env.STAGES.contains("Minor Upgrade")
+          || env.STAGES.contains("Major Upgrade")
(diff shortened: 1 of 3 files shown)


mattt416:
Total WIP, please do not merge!

odyssey4me:
@mattt416 I'd prefer to see us move this complexity out of the gating repo and into rpc-openstack. If we stop doing the upgrade branch checkout in the gating repo, but instead pass the axes from the job to the node executing the test, then the node executing the test can figure out what to do using the in-repo code.

For example, we pass the 'action' env var with the value 'leapfrogupgrade', or 'minorupgrade' or 'majorupgrade' to the deploy script. The deploy script then has a set of default start points for each of those and we do not need to set all those values in the gating repo.

odyssey4me:
FYI I think https://github.com/rcbops/rpc-openstack/pull/2466 might fix the minorupgrade test.

PR #368, opened 2017-08-02
RE-170 Remove newton-14.1 jobs

hughsaunders:
This branch is now finalised, so the jobs are no longer required.

Issue: [RE-170](https://rpc-openstack.atlassian.net/browse/RE-170)


hughsaunders:
👎  I need to also remove newton141 from the multinode jobs

hughsaunders:
Updated.

PR #371, opened 2017-08-02
RE-175 Change minorupgrade to run from r14.1.0

mattt416:
Currently, the minorupgrade is upgrading r14.0.1 to newton, which is
actually skipping a minor release.  This commit changes rpc_aio.yml to
upgrade r14.1.0 to newton instead.

Issue: [RE-175](https://rpc-openstack.atlassian.net/browse/RE-175)

### DIFF
--- a/rpc_jobs/rpc_aio.yml
+++ b/rpc_jobs/rpc_aio.yml
@@ -66,7 +66,7 @@
           # NOTE: Once the newton stabilization branch has been created
           #       (newton-14.2 or similar), we'll need to find a way to
           #       set this differently depending on the newton series.
-          UPGRADE_FROM_REF: "r14.0.1"
+          UPGRADE_FROM_REF: "r14.1.0"
     scenario:
       - swift
       - ceph:
@@ -95,7 +95,7 @@
           DEPLOY_SUPPORT_ROLE: "yes"
     exclude:
       # Minor upgrades are only being executed
-      # for r14.0.1->newton for now until
+      # for r14.1.0->newton for now until
       # the minor upgrade testing process is
       # stabilised.
       - series: kilo


mattt416:
A minor upgrade from r14.0.1 to newton doesn't pass, trying r14.1.0 to newton here:

https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_newton-trusty-minorupgrade-swift-periodic/3/console

odyssey4me:
I'd ideally prefer *not* to do this because we'd like to test the earliest possible release as the starting point for the upgrade. That said, if it doesn't work then it also doesn't serve much purpose and we should go with the earliest one that does.

mattt416:
@odyssey4me I know the minorupgrade job was previously working because it wasn't actually upgrading anything (see https://github.com/rcbops/rpc-gating/pull/363).  Once https://github.com/rcbops/rpc-gating/pull/363 merged, I started seeing this error:

```
TASK [pip_install : Get Modern PIP] ********************************************
[0;31mfatal: [rantmsp-1-d464]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: {{ rpco_mirror_base_url }}/os-releases/{{ rpc_release }}/{{ os_distro_version }}/get-pip.py: 'rpco_mirror_base_url' is undefined\n\nThe error appears to have been in '/etc/ansible/roles/pip_install/tasks/install_online.yml': line 16, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Get Modern PIP\n  ^ here\n"}[0m
```
It wasn't clear if r14.0.1 should cleanly upgrade to newton, and I thought it made sense for the newton branch to test a minor upgrade from the last created tag.

odyssey4me:
@mattt416 The cause for that failure was that the new defaults file wasn't copied over. https://github.com/rcbops/rpc-openstack/pull/2466 will fix that, after the backport of course.

PR #373, opened 2017-08-03
RE-133 Ensure variables are typed/scoped

hughsaunders:
In groovy untyped variables are almost global. In order to prevent
conflicts when using parallel, when need to ensure all variables are
typed and therefore scoped.

Also adds a random string to the vars file written by the venvplaybook
function to avoid conflicts.

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -109,9 +109,10 @@ def venvPlaybook(Map args){
       if (!('args' in args)){
         args.args=[]
       }
-      for (def i=0; i<args.playbooks.size(); i++){
-        playbook = args.playbooks[i]
-        vars_file="vars.${playbook.split('/')[-1]}"
+      for (int i=0; i<args.playbooks.size(); i++){
+        String playbook = args.playbooks[i]
+        // randomised vars file path for parallel safety
+        String vars_file="vars.${playbook.split('/')[-1]}.${rand_int_str()}"
         write_json(file: vars_file, obj: args.vars)
         sh """#!/bin/bash -x
           which scl && source /opt/rh/python27/enable
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -125,7 +125,7 @@ def venvPlaybook(Map args){
 } //venvplaybook
 
 def calc_ansible_forks(){
-  def forks = sh (script: """#!/bin/bash
+  String forks = sh (script: """#!/bin/bash
     CPU_NUM=\$(grep -c ^processor /proc/cpuinfo)
     if [ \${CPU_NUM} -lt "10" ]; then
       ANSIBLE_FORKS=\${CPU_NUM}
@@ -143,7 +143,7 @@ to evaluate ${forks} and fail.
 These vars should be set every time deploy.sh or test-upgrade is run
 */
 List get_deploy_script_env(){
-  forks = calc_ansible_forks()
+  String forks = calc_ansible_forks()
   return [
     'ANSIBLE_FORCE_COLOR=true',
     'ANSIBLE_HOST_KEY_CHECKING=False',
@@ -283,19 +283,20 @@ def conditionalStep(Map args){
  *  string: the string to process
  */
 def String acronym(Map args){
-  acronym=""
-  words=args.string.split("[-_ ]")
+  String acronym=""
+  List words=args.string.split("[-_ ]")
   for (def i=0; i<words.size(); i++){
     acronym += words[i][0]
   }
   return acronym
 }
 
-def String rand_int_str(max=0xFFFF, base=16){
+def String rand_int_str(int max=0xFFFF, int base=16){
   return Integer.toString(Math.abs((new Random()).nextInt(max)), base)
 }
 
 def String gen_instance_name(String prefix="AUTO"){
+  String instance_name = ""
   if (env.INSTANCE_NAME == "AUTO"){
     if (prefix == "AUTO"){
       prefix = acronym(string: env.JOB_NAME)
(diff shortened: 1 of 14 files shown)


hughsaunders:
Successfull  parallel instance build: https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/scratchpipeline/detail/scratchpipeline/366/pipeline/16

Job:
```
common.shared_slave(){
      env.INSTANCE_NAME="AUTO"
      env.STAGES="Allocate Resources, Connect Slave, Destroy Slave, Cleanup"
      env.REGION="IAD"
      env.IMAGE="Ubuntu 14.04 LTS (Trusty Tahr) (PVHVM)"
      env.FLAVOR="performance1-1"
      
      parallel ([
          a: {
            pubcloud.runonpubcloud {
                stage("display hostname a"){
                    sh "hostname"
                }
            }
          },
          b: {
            pubcloud.runonpubcloud{ 
                stage("display hostname b"){
                    sh "hostname"
                }
            }
          }
      ])
}
```

hughsaunders:
Test artifact pipeline: https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/RPC-Artifact-Build_newton-periodic/detail/RPC-Artifact-Build_newton-periodic/9/pipeline

hughsaunders:
Test build after splitting periodic container build into multiple stages: https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/RPC-Artifact-Build_newton-periodic/detail/RPC-Artifact-Build_newton-periodic/10/pipeline

rpc_jobs/rpc_artifact_build.yml
```diff
@@ -203,14 +203,14 @@
...
             // image index.
             // ref: RE-115
-            stage('Container'){{
-              lock('artifact_container'){{
-                for (x in image_list[env.series]){{
-                  def image = x
+            lock('artifact_container'){{
+              for (x in image_list[env.series]){{
+                def image = x
+                stage("Container ${{image}}"){{
```
odyssey4me: nice way of handling this :)

playbooks/allocate_pubcloud.yml
```diff
@@ -58,7 +64,7 @@
           {% for instance in rax.success %}
           {{instance.name}} ansible_host={{instance.accessIPv4}} ansible_user=root
           {% endfor %}
-        dest: '{{lookup("env", "WORKSPACE")}}/rpc-gating/playbooks/inventory/hosts'
+        dest: '{{inventory_path}}/hosts'
```
odyssey4me: I'm not clear on why this is necessary. Is the node this executes against not the same as the node doing the executing?

hughsaunders:
test aio build: https://rpc.jenkins.cit.rackspace.net/view/AIO/job/RPC-AIO_mitaka-trusty-deploy-swift-periodic/44/console

playbooks/allocate_pubcloud.yml
hughsaunders: Thats needed because we insert a random string into inventory path for parallel runs, so we need to be able to override inventory_path.

odyssey4me: I missed that the playbook was localhost targeted, and see that now. That still doesn't help me understand though.

I think I'm getting it - you need a var that's overridable here. So it's not really about changes to this playbook, but instead about how the playbook is used.

odyssey4me: Ah yes, now the changes to pipeline_steps/ssh_slave.groovy make more sense.

hughsaunders:
artifact build post split into multiple commits: https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/RPC-Artifact-Build_newton-periodic/detail/RPC-Artifact-Build_newton-periodic/11/pipeline

pipeline_steps/common.groovy
```diff
@@ -283,19 +284,20 @@ def conditionalStep(Map args){
  *  string: the string to process
  */
 def String acronym(Map args){
-  acronym=""
-  words=args.string.split("[-_ ]")
+  String acronym=""
+  List words=args.string.split("[-_ ]")
   for (def i=0; i<words.size(); i++){
```
mattt416: does this def here need to be changed to int?

pipeline_steps/common.groovy
```diff
@@ -461,7 +463,7 @@ def docker_cache_workaround(){
    sh "touch -t 201704100000 *.txt"
 }
 
-def is_doc_update_pr(git_dir) {
+def is_doc_update_pr(String git_dir) {
   if (env.ghprbPullId != null) {
     dir(git_dir) {
       def rc = sh(
```
mattt416: de here needs changing?

pipeline_steps/common.groovy
```diff
@@ -491,7 +493,7 @@ def is_doc_update_pr(git_dir) {
  * This function uses environment variables injected by github pull request
  * builder and so can only be used for PR triggered jobs
  */
-def get_jira_issue_key(repo_path="rpc-openstack"){
+def get_jira_issue_key(String repo_path="rpc-openstack"){
   def key_regex = "[a-zA-Z][a-zA-Z0-9_]+-[1-9][0-9]*"
```
mattt416: def here needs changing?

pipeline_steps/maas.groovy
```diff
@@ -79,7 +79,7 @@ def verify() {
 void add_maas_env_vars(){
   List vars = get_maas_token_and_url()
   for (def i=0; i<vars.size(); i++){
```
mattt416: Don't know which defs are problematic but here's another that can be changed to an int.

hughsaunders:
@mattt416 My explanation wasn't quite correct. Variables declared with `def` have the same scope as typed variables, so its not critical that they are changed. I changed the ones I noticed, for consistency. The problem is variables with no definition beyond their initialisation, eg `foo = 1`

mattt416:
Thanks @hughsaunders -- that makes more sense then.

hughsaunders:
Next test build, previous was thwarted by puppet restarting jenkins: https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/RPC-Artifact-Build_newton-periodic/detail/RPC-Artifact-Build_newton-periodic/12/pipeline

odyssey4me:
Given that the test is now hitting the last stage, everything appears to be working right.

PR #374, opened 2017-08-04
Flip builds to IAD

mattt416:
We've hit a rate limit in DFW and have a critical patch that needs to
gate.  This PR temporarily flips all builds to IAD so we can get this
PR to gate.


mattt416:
Thanks @major !  Looks like builds are taking in DFW again, so closing this for now.

PR #378, opened 2017-08-09
RE-8 Cleanup when job times out

mattt416:
Currently, if a job times out no cleanup is triggered.  This commit moves the
timeout block inside the try/except so that cleanup is triggered if a job
actually times out.

Issue: [RE-8](https://rpc-openstack.atlassian.net/browse/RE-8)

### DIFF
--- a/rpc_jobs/rpc_aio.yml
+++ b/rpc_jobs/rpc_aio.yml
@@ -276,66 +276,66 @@
       env.TARGET = "aio"
       env.ACTION = "{action}"
       env.SERIES = "{series}"
-      timeout(time: 8, unit: 'HOURS'){{
         common.shared_slave() {{
           try {{
-            // prepare vars for leapfrog upgrade test of changes proposed to kilo
-            // this is a special case as we only usually test upgrades on
-            // changes to the destination branch.
-            if (env.SERIES == "kilo"
-                && env.TRIGGER == "pr"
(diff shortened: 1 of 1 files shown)


mattt416:
Tested [here](https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_newton-trusty-deploy-swift-periodic/7/console).  I set the timeout to 1 hour so the build would definitely time out.

rpc_jobs/rpc_aio.yml
```diff
@@ -281,80 +281,80 @@
...
+                    "DEPLOY_SWIFT=${{DEPLOY_SWIFT}}",
+                    "DEPLOY_CEPH=${{DEPLOY_CEPH}}",
+                    "DEPLOY_ELK=${{DEPLOY_ELK}}",
+                    "DEPLOY_SUPPORT_ROLE=${{DEPLOY_SUPPORT_ROLE}}",
+                    ]
+                  aio_prepare.prepare()
+                  deploy.deploy_sh(environment_vars: environment_vars)
                   tempest.tempest()
-                  kibana.kibana(env.KIBANA_SELENIUM_BRANCH)
+                  kibana.kibana(kibana_branch)
```
hughsaunders: did you mean to change the branch?

mattt416: I don't think I did ... I think line 333 is associated w/ line 333 on the other side.

PR #379, opened 2017-08-09
RE-188 Fix stashing of multi-node inventory

git-harry:
01eac2741173f3016e016f0cbcb31d986ff5c25c modified the process of
stashing the inventory. This change resulted in `connect_deploy_node`
creating a stash that neither used the default name nor passed the
correct name to `ssh_slave.connect`.

This change updates `connect_deploy_node` to use the default.

The setting of the default for `args.inventory` is also fixed so that
the value is correctly assigned.

Issue: [RE-188](https://rpc-openstack.atlassian.net/browse/RE-188)

### DIFF
--- a/pipeline_steps/multi_node_aio_prepare.groovy
+++ b/pipeline_steps/multi_node_aio_prepare.groovy
@@ -160,7 +160,7 @@ def connect_deploy_node(name, instance_ip) {
   common.drop_inventory_file(inventory_content)
   dir("rpc-gating/playbooks"){
     stash (
-      name: "pubcloud_inventory",
+      name: "inventory",
       include: "inventory/hosts"
     )
   }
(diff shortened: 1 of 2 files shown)


git-harry:
@odyssey4me @mattt416 you both reviewed the original pull request, hopefully you can confirm whether or not this is sufficient or if it also requires a unique inventory name.

mattt416:
Seeing this in mnaio builds:

```
Caught exception on rpc-jenkins-n02.prod.dfw1.cit.rackspace.net-9082c1d1: java.lang.IllegalArgumentException: Could not instantiate {name=null} for UnstashStep(name: String): java.lang.reflect.InvocationTargetException
```

Merging as I'm assuming this resolves that error.

PR #380, opened 2017-08-10
Cleanup tempest execution

evrardjp:
We need to:
- Avoid running tempest many times for no reason, 1 per branch
  is enough during upgrades.
- Still install tempest if need be (for resource gen)

That shouldn't break stuff, while making our life easier.

### DIFF
--- a/rpc_jobs/rpc_aio.yml
+++ b/rpc_jobs/rpc_aio.yml
@@ -329,21 +329,20 @@
                   ]
                 aio_prepare.prepare()
                 deploy.deploy_sh(environment_vars: environment_vars)
-                tempest.tempest()
+                tempest.tempest_install()
                 kibana.kibana(kibana_branch)
                 holland.holland()
                 if (env.STAGES.contains("Minor Upgrade")) {{
                   deploy.upgrade_minor(environment_vars: environment_vars)
                 }} else if (env.STAGES.contains("Major Upgrade")) {{
                   deploy.upgrade_major(environment_vars: environment_vars)
-                  tempest.tempest()
                   kibana.kibana(env.KIBANA_SELENIUM_BRANCH)
                   holland.holland()
                 }} else if (env.STAGES.contains("Leapfrog Upgrade")) {{
                   deploy.upgrade_leapfrog(environment_vars: environment_vars)
-                  tempest.tempest()
                   kibana.kibana(env.KIBANA_SELENIUM_BRANCH)
                 }}
+                tempest.tempest()
               }} catch (e) {{
                 print(e)
                 currentBuild.result = 'FAILURE'


rpc_jobs/rpc_aio.yml
```diff
@@ -329,21 +329,20 @@
                   ]
                 aio_prepare.prepare()
                 deploy.deploy_sh(environment_vars: environment_vars)
-                tempest.tempest()
+                tempest.tempest_install()
```
git-harry: I'm against this. The purpose is to verify that the base deployment was successful. If tempest is failing all the time we should address that and not ignore it.

evrardjp:
Ok nothing can merge in leapfrog, and we will be bypassing these checks. Why testing kilo in newton branch?
Kilo should deploy just fine. Or we test it in kilo branch.

Let me summarise, in the leapfrog kilo to newton case:
- If there was a temp issue with kilo, we can retry the job
- If there is a long term issue with kilo, we'll have to fix it in kilo, which will still test with this code, on kilo branch
- If there is a long term issue happening after leapfrog this will still be consistently failing
- If there is a temp issue happening after leapfrog, we can still retry/fix the root cause.

On that last case, if the root cause is because we didn't deploy tempest on earlier branch, then maybe there are other things to fix in the leapfrog.

All of the above works if you replace leapfrog with major/minor upgrade and kilo with any previous branch.

evrardjp:
Worst case scenario: We merge things without testing because kilo fails.

odyssey4me:
@git-harry The functional test purpose is served via the resources added, so I think it's fine not to do the tempest tests. If we do them, however, I would suggest a smaller test set - we do not need to do the defcore set, all we need is a canary test.

evrardjp:
@odyssey4me I think that if something broken comes back up thanks to the leapfrog, it's not really hurting the results.

odyssey4me:
@evrardjp Agreed - if the *final* test results are good, then if the initial cloud wasn't totally functional it's immaterial. However it is also possible that we will see weird failures in the leapfrog due to issues which are present in the initial deployment.

mkam:
While it's not ideal to skip the pre-upgrade tests, the purpose of these jobs is to test the leapfrog process, so I personally would rather have the upgrade happen even if the base is broken. The job as is doesn't give us any new information about upgrades if the kilo failures are blocking it. If we skip the tests, we'll at least learn if an upgrade is successful or not with a less than perfectly functioning base.

We could also try to make the results of the first tempest run not block the upgrade so that we have the initial results to compare with the final results in case of failures post-upgrade. That'll be a little tricky to implement, though.

evrardjp:
@mkam @odyssey4me could I get some votes in that case?

rpc_jobs/rpc_aio.yml
hughsaunders: Having discussed this with @evrardjp I can see why the first tempest run could be removed. Firstly some of the base deploys are old versions of kilo that can't be modified and secondly the first deploy should be tested by other jobs.

PR #381, opened 2017-08-10
LA-408 Archive log/config artefacts for MNAIOs

git-harry:
The change updates the `multi_node_aio.yml` so that multi-node builds
now have their log/config artefacts saved. To enable this to work
`archive_artifacts` has been modified to support both AIO and MNAIO
builds.

Issue: [LA-408](https://rpc-openstack.atlassian.net/browse/LA-408)

### DIFF
--- a/playbooks/archive_artifacts.yml
+++ b/playbooks/archive_artifacts.yml
@@ -19,7 +19,7 @@
                {{ inventory_hostname }}:{{ item }}
                {{ artifacts_path }}/{{ ansible_hostname }}
       with_items:
-        - "/openstack/log/{{ ansible_hostname }}-lxc"
+        - "/openstack/log"
         - "/etc"
         - "/var/log"
       delegate_to: "localhost"
@@ -65,7 +65,6 @@
       with_nested:
         - "{{ containers.stdout_lines | default([]) }}"
         -
-          - { pre: "/openstack/log", post: "" }
           - { pre: "/var/lib/lxc", post: "/rootfs/etc" }
           - { pre: "/var/lib/lxc", post: "/delta0/etc" }
       delegate_to: "localhost"
--- a/playbooks/archive_artifacts.yml
+++ b/playbooks/archive_artifacts.yml
@@ -1,0 +1,88 @@
+- hosts: "{{ target_hosts }}"
+  tasks:
+    - name: Create temporary directory to store artifacts
+      file:
+        path: "{{ artifacts_path }}"
+        state: directory
+      delegate_to: "localhost"
+      run_once: yes
+
+    - name: Grab host data
+      command: >
+               rsync
+               --archive
+               --compress
+               --verbose
+               --relative
+               --rsh 'ssh -o StrictHostKeyChecking=no'
+               --ignore-missing-args
+               {{ inventory_hostname }}:{{ item }}
+               {{ artifacts_path }}/{{ ansible_hostname }}
+      with_items:
+        - "/openstack/log/{{ ansible_hostname }}-lxc"
+        - "/etc"
+        - "/var/log"
+      delegate_to: "localhost"
+      tags:
+        - skip_ansible_lint
+
+    - name: Grab leapfrog upgrade .leap marker files
+      command: >
+               rsync
+               --archive
+               --compress
+               --verbose
+               --rsh 'ssh -o StrictHostKeyChecking=no'
+               --include '*.leap'
+               --exclude '*'
+               {{ inventory_hostname }}:/opt/rpc-leapfrog/leap42/
+               {{ artifacts_path }}/{{ ansible_hostname }}/leap_marker_files
+      delegate_to: "localhost"
+      ignore_errors: yes
+      tags:
+        - skip_ansible_lint
+
+    - name: List containers
+      command: "lxc-ls -1"
+      failed_when:
+        - "{{ containers.rc != 0 }}"
+        - "{{ containers.msg != '[Errno 2] No such file or directory' }}"
+      changed_when: false
+      register: containers
+
+    - name: Grab container data
+      command: >
+               rsync
+               --archive
+               --compress
+               --verbose
+               --relative
+               --rsh 'ssh -o StrictHostKeyChecking=no'
+               --ignore-missing-args
+               {{ inventory_hostname }}:{{ item[1].pre }}/{{ item[0] }}{{ item[1].post }}
+               {{ artifacts_path }}/{{ ansible_hostname }}
+      when: "{{ containers.rc == 0 }}"
+      with_nested:
+        - "{{ containers.stdout_lines | default([]) }}"
+        -
+          - { pre: "/openstack/log", post: "" }
+          - { pre: "/var/lib/lxc", post: "/rootfs/etc" }
+          - { pre: "/var/lib/lxc", post: "/delta0/etc" }
+      delegate_to: "localhost"
+      tags:
+        - skip_ansible_lint
+
+    - name: Create archive
+      command: "tar -cjf {{ artifacts_basename }}.tar.bz2 {{ artifacts_basename }}"
+      args:
+        chdir: "{{ artifacts_basedir }}"
+      delegate_to: "localhost"
+      run_once: yes
+      tags:
+        - skip_ansible_lint
+  vars:
+    build_tag: "{{ lookup('env','BUILD_TAG') }}"
+    artifacts_basename: "artifacts_{{ build_tag }}"
+    artifacts_basedir: "{{ lookup('env','WORKSPACE') }}"
+    artifacts_path: "{{ artifacts_basedir }}/{{ artifacts_basename }}"
+    target_hosts: "localhost"
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -325,34 +325,28 @@ def String gen_instance_name(String prefix="AUTO"){
   return instance_name
 }
 
-def archive_artifacts(){
+def archive_artifacts(String build_type = "AIO"){
   try{
-    sh """#!/bin/bash
-    echo "Archiving logs and configs..."
-    d="artifacts_\${BUILD_TAG}"
-    mkdir -p \$d
-
-    # logs and config from the single use slave
-    mkdir -p \$d/\$HOSTNAME/log
-    cp -rp /openstack/log/\$HOSTNAME-* \$d/\$HOSTNAME/log ||:
-    cp -rp /etc/ \$d/\$HOSTNAME/etc
-    cp -rp /var/log/ \$d/\$HOSTNAME/var_log
-
-    # logs and config from the containers
-    while read c
-    do
-      mkdir -p \$d/\$c/log
-      cp -rp /openstack/log/\$c/* \$d/\$c/log 2>/dev/null ||:
-      cp -rp /var/lib/lxc/\$c/rootfs/etc \$d/\$c 2>/dev/null ||:
-      cp -rp /var/lib/lxc/\$c/delta0/etc \$d/\$c 2>/dev/null ||:
-    done < <(lxc-ls)
-
-    # compress to reduce storage space requirements
-    ARTIFACT_SIZE=\$(du -sh \$d | cut -f1)
-    echo "Compressing \$ARTIFACT_SIZE of artifact files..."
-    tar cjf "\$d".tar.bz2 \$d
-    echo "Compression complete."
-    """
+    if ( build_type == "MNAIO" ){
+      args = [
+        "-i /opt/ansible-static-inventory.ini",
+      ]
+      vars = [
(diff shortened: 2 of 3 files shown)


playbooks/archive_artifacts.yml
```diff
@@ -0,0 +1,77 @@
...
+               {{ inventory_hostname }}:/opt/rpc-leapfrog/leap42/
+               {{ artifacts_path }}/{{ ansible_hostname }}/leap_marker_files
+      delegate_to: "localhost"
+      ignore_errors: yes
+
+    - name: List containers
+      command: "lxc-ls -1"
+      failed_when:
+        - "{{ containers.rc != 0 }}"
+        - "{{ containers.msg != '[Errno 2] No such file or directory' }}"
```
evrardjp: Could you explain why this is acceptable?

playbooks/archive_artifacts.yml
```diff
@@ -0,0 +1,77 @@
...
+    - name: Create archive
+      command: "tar -cjf {{ artifacts_basename }}.tar.bz2 {{ artifacts_basename }}"
+      args:
+        chdir: "{{ artifacts_basedir }}"
+      delegate_to: "localhost"
+      run_once: yes
+  vars:
+    artifacts_basename: "artifacts_{{ build_tag }}"
+    artifacts_path: "{{ artifacts_basedir }}/{{ artifacts_basename }}"
+    hosts: "localhost"
```
evrardjp: why would you want to define this, if you give that in the cli? Shouldn't it fail if not define instead?

playbooks/archive_artifacts.yml
```diff
@@ -0,0 +1,77 @@
...
+      delegate_to: "localhost"
+
+    - name: Create archive
+      command: "tar -cjf {{ artifacts_basename }}.tar.bz2 {{ artifacts_basename }}"
+      args:
+        chdir: "{{ artifacts_basedir }}"
+      delegate_to: "localhost"
+      run_once: yes
+  vars:
+    artifacts_basename: "artifacts_{{ build_tag }}"
```
evrardjp: ``artifacts_basedir: "{{ lookup('env','ARTIFACTS_BASEDIR') }}"``. If necessary, ``| default('/workspace/',True)``

``build_tag: "{{ lookup('env','BUILD_TAG') }}"``

playbooks/archive_artifacts.yml
```diff
@@ -0,0 +1,77 @@
+- hosts: "{{ hosts }}"
```
git-harry: It's use doesn't shadow anything, not sure it's really a problem but happy to change.

evrardjp: it will trigger a warning rn with recent ansible and will fail in the future.

playbooks/archive_artifacts.yml
```diff
@@ -0,0 +1,77 @@
...
+  tasks:
+    - name: Create temporary directory to store artifacts
+      file:
+        path: "{{ artifacts_path }}"
```
git-harry: Without creating "{{ artifacts_path }}" rsync will fail, any required subdirectories will be created by rsync. So, this can be done here but it's not necessary.

evrardjp: I thought rsync couldn't create the {{ artifacts_path }}/{{ ansible_hostname }}, just other child directories of "the base dir". I guess I am wrong.

playbooks/archive_artifacts.yml
git-harry: Okay, good to know.

playbooks/archive_artifacts.yml
```diff
@@ -0,0 +1,77 @@
...
+        state: directory
+      delegate_to: "localhost"
+      run_once: yes
+
+    - name: Grab host data
+      command: >
```
git-harry: I had a quick play with it at the time, it didn't work but I may have been doing it wrong. It was easier for me just to use rsync directly.

playbooks/archive_artifacts.yml
git-harry: If "this" refers to ignoring missing directories, it is acceptable because it means `lxc-ls` isn't installed and not all hosts have it e.g. nova-compute and deploy hosts.

playbooks/archive_artifacts.yml
git-harry: Yes, agreed but the lint check failed without it.

playbooks/archive_artifacts.yml
evrardjp: Fine for me, but could you add:
``
tags:
  - skip_ansible_lint
``

for the future, and a todo item to replace that with synchronize module then? 
I am afraid that it should pass the linting right now, so maybe we are already too lenient in our testing.

playbooks/archive_artifacts.yml
evrardjp: Ahah yes, ofc!

playbooks/archive_artifacts.yml
evrardjp: ok.

git-harry:
@evrardjp I've pushed a second commit, let me know if it looks good and I'll squash them down.

evrardjp:
Will do today.

playbooks/archive_artifacts.yml
```diff
@@ -0,0 +1,85 @@
...
+      delegate_to: "localhost"
+      ignore_errors: yes
+      tags:
+        - skip_ansible_lint
+
+    - name: List containers
+      command: "lxc-ls -1"
+      failed_when:
+        - "{{ containers.rc != 0 }}"
+        - "{{ containers.msg != '[Errno 2] No such file or directory' }}"
```
evrardjp: technically you need a changed_when: false to pass linting too.
Sorry I forgot that.

playbooks/archive_artifacts.yml
```diff
@@ -0,0 +1,85 @@
...
+      delegate_to: "localhost"
+      tags:
+        - skip_ansible_lint
+
+    - name: Create archive
+      command: "tar -cjf {{ artifacts_basename }}.tar.bz2 {{ artifacts_basename }}"
+      args:
+        chdir: "{{ artifacts_basedir }}"
+      delegate_to: "localhost"
+      run_once: yes
```
evrardjp: technically to pass linting we need a changed_when.
Or skip_ansible_lint.

git-harry:
recheck_cit_rpc_gating_lint

git-harry:
@evrardjp the lint job doesn't run `ansible-lint`, if it did a number of playbooks would fail. I'll take a look at making your suggested changes but given that the playbook works and these changes won't be enforced, I don't think it's a reason to block the pull request.

odyssey4me:
Is there a reason why this is being done in the rpc-gating repo instead of in the rpc-openstack repo? In my view the artifact gathering is series-specific and best done in-repo per-branch. The rpc-gating repo should just collect the results artifacts from a known location in the workspace.

evrardjp:
It was already done there...

git-harry:
@odyssey4me the purpose of the commit is mainly to provide the same level of artefact-storing support to MNAIO as is currently given to AIO jobs. Refactoring this into another repo brings unnecessary risks and is a separate piece of work.

odyssey4me:
@git-harry Fair enough. If there's a card for the move of all this stuff into RPC-O then I'm OK with it as I know it won't be forgotten.

git-harry:
@odyssey4me it seems the beginnings of this has already been started [1] but has been left in an unhelpful no mans land. I will update the higher level card [2], tracking the more general work to prevent confusion.

[1]  https://rpc-openstack.atlassian.net/browse/UG-645
[2] https://rpc-openstack.atlassian.net/browse/RE-87

playbooks/archive_artifacts.yml
```diff
@@ -0,0 +1,88 @@
...
+               --archive
+               --compress
+               --verbose
+               --relative
+               --rsh 'ssh -o StrictHostKeyChecking=no'
+               --ignore-missing-args
+               {{ inventory_hostname }}:{{ item }}
+               {{ artifacts_path }}/{{ ansible_hostname }}
+      with_items:
+        - "/openstack/log/{{ ansible_hostname }}-lxc"
```
mattt416: Previously we were grabbing `/openstack/log/$HOSTNAME-*`, if we do `/openstack/log/{{ ansible_hostname }}-lxc` here we miss out on a bunch of logs and also the LXC logs should already be included in `/var/log` I believe?

Here's all the logs for services running on the host itself:

```
root@liberty-aio:/openstack/log# ls -dl /openstack/log/$HOSTNAME-*
drwxr-xr-x 2 cinder  cinder  4096 Aug 23 22:15 /openstack/log/liberty-aio-cinder
drwxr-xr-x 2 syslog  adm     4096 Aug 23 21:11 /openstack/log/liberty-aio-haproxy
drwxr-xr-x 2 neutron neutron 4096 Aug 23 22:26 /openstack/log/liberty-aio-neutron
drwxr-xr-x 2 nova    nova    4096 Aug 23 22:22 /openstack/log/liberty-aio-nova
drwxr-xr-x 2 syslog  syslog  4096 Aug 23 22:41 /openstack/log/liberty-aio-swift
root@liberty-aio:/openstack/log#
```

playbooks/archive_artifacts.yml
```diff
@@ -0,0 +1,88 @@
...
+               --verbose
+               --relative
+               --rsh 'ssh -o StrictHostKeyChecking=no'
+               --ignore-missing-args
+               {{ inventory_hostname }}:{{ item[1].pre }}/{{ item[0] }}{{ item[1].post }}
+               {{ artifacts_path }}/{{ ansible_hostname }}
+      when: "{{ containers.rc == 0 }}"
+      with_nested:
+        - "{{ containers.stdout_lines | default([]) }}"
+        -
```
mattt416: Is this empty item intentional?

playbooks/archive_artifacts.yml
```diff
@@ -0,0 +1,88 @@
...
+    - name: Grab leapfrog upgrade .leap marker files
+      command: >
+               rsync
+               --archive
+               --compress
+               --verbose
+               --rsh 'ssh -o StrictHostKeyChecking=no'
+               --include '*.leap'
+               --exclude '*'
+               {{ inventory_hostname }}:/opt/rpc-leapfrog/leap42/
```
evrardjp: We can archive the whole rpc-leapfrog folder.

playbooks/archive_artifacts.yml
evrardjp: Agreed.

git-harry: I believe that should be taken care of by line 68.

playbooks/archive_artifacts.yml
git-harry: rpc-leapfrog is approximately 0.5GB and is mostly just venvs and repos. If there's no concern with the additional data I'm happy to change this given that too little data is much worse than too much when trying to review a failed build.

playbooks/archive_artifacts.yml
git-harry: No, I've missed that.

mattt416: Actually this dash is right, needs to be here.

playbooks/archive_artifacts.yml
mattt416: Line 68 only grabs container data, what if you're on an AIO that has cinder, haproxy, neutron, nova, and swift running outside of containers?  You'd want to grab that data which was done previously using $HOSTNAME-* (this doesn't match containers, since that'd be $HOSTNAME_*).

playbooks/archive_artifacts.yml
git-harry: Ha, oh yes. With the comment in between I didn't spot the additional indentation.

git-harry:
@mattt416 I've tested this change with https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_newton-trusty-deploy-swift-periodic/41/

It created the archive file using the old method and with this patch. When I compared the two, I believe the only difference is the addition of ansible.log. If you want to look at the files, the link on the build page will give you the file produced using the previous method, just modify that link to remove "_old" to get the file created with this patch.

mattt416:
@git-harry it's a bit hard to compare tarballs since the structure has changed quite dramatically (container logs now in `rantdsp-41-e15d/openstack/log/rantdsp-41-e15d_cinder_api_container-3ceef8e0` versus `rantdsp-41-e15d_cinder_api_container-3ceef8e0/log` as an example).  That said, there seem to be _more_ files in the new tarballs so I don't think we've lost anything.

git-harry:
@mattt416 I generated a hash for each file and used that to compare the two. There were a few log files that had changed by the time the second archive file was created but mostly identical.

PR #382, opened 2017-08-10
Do tempest twice

evrardjp:
Tempest has sometimes transient issues, and this can be fixed
by running tempest twice.

We already ran tempest twice on some condition, this formalizes
it for other conditions.

### DIFF
--- a/pipeline_steps/tempest.groovy
+++ b/pipeline_steps/tempest.groovy
@@ -33,17 +33,12 @@ def tempest(){
     stage_name: "Tempest Tests",
     stage: {
       try{
-        def result = tempest_run()
-        def second_result = ""
-        if(result.contains("Race in testr accounting.")){
-          second_result = tempest_run()
+        retry(2){
+          def result = tempest_run()
+          if(result.contains("Race in testr accounting.")) {
+            throw new Exception("Race in testr accounting")
+          }
         }
-        if(second_result.contains("Race in testr accounting.")) {
-          currentBuild.result = 'FAILURE'
-        }
-        } catch (e){
-        print(e)
-        throw(e)
       } finally{
         sh """#!/bin/bash -x
           # Set up variables for copy


evrardjp:
Need expert eyes here.

pipeline_steps/tempest.groovy
```diff
@@ -32,29 +32,27 @@ def tempest(){
...
-          second_result = tempest_run()
-        }
-        if(second_result.contains("Race in testr accounting.")) {
-          currentBuild.result = 'FAILURE'
-        }
+      retry(2){
+        try{
+          def result = tempest_run()
+          if(result.contains("Race in testr accounting.")) {
+            currentBuild.result = 'FAILURE'
```
mattt416: Would you need to throw an exception here so that it tries it again?

evrardjp: I'd say yes.

What about the currentBuild.result? What is used for?
Should we remove that/modify that, so that it only get set when catch clause is reached?

mkam: It was to set it to failure out of caution in case the test results were a false positive. If the `Race in testr accounting` message is in the output, that means the results of one or more workers has been lost (if running in parallel). Tempest won't fail the build even if that error is thrown, so I had to set it manually.

https://bugs.launchpad.net/testrepository/+bug/1538941

I don't believe I've seen it happen since we started using specifically 3 workers. Additionally, the parallel runner is only used in multi-node AIO tests in branches that aren't master since stable/ocata has `--serial` hardcoded. So this is potentially not really a problem anymore. I just kept it in because I'm paranoid and nothing makes me sadder than false positives.

https://github.com/openstack/openstack-ansible-os_tempest/blob/stable/ocata/tasks/tempest_run.yml#L20

mattt416: Let's keep the testr check in place then @mkam.  So if we throw an error in this if block will that be sufficient?

mkam: Yup! That should be fine since it should change the build to failed so you don't have to manually set currentBuild.result.

evrardjp:
I am still not sure how retry works and if the ``finally`` wouldn't prevent a good behavior of the retry function. Anyone before I update this with @mkam 's comments?

pipeline_steps/tempest.groovy
hughsaunders: Would it make more sense to put the retry with the try block so that the finally block only happens once?

PR #384, opened 2017-08-14
RE-193 Add python-selinux when unpacking venv

hughsaunders:
The gating venv doesn't include the selinux module as its created on
ubuntu so this module must be copied in when unpacking the venv.


pipeline_steps/common.groovy
```diff
@@ -66,6 +66,8 @@ def void install_ansible(){
         echo "CentOS node detected, copying in external python interpreter and setting PYTHONPATH in activate script"
         # CentOS 6 can take a hike, its glibc isn't new enough for python 2.7.12
         cp /opt/rh/python27/root/usr/bin/python .venv/bin/python
+        # hack the selinux module into the venv
```
hughsaunders: copied from create_venv above.

PR #386, opened 2017-08-14
RE-196 Add release job

hughsaunders:
Basic job for tagging releases and updating the stablisation branch

Issue: [RE-196](https://rpc-openstack.atlassian.net/browse/RE-196)


hughsaunders:
Sample job: https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/RE-Release/detail/RE-Release/12/

rpc_jobs/release.yml
```diff
@@ -0,0 +1,68 @@
...
+        }
+        common.conditionalStage(
+          stage_name: "Tag",
+          stage: {
+            release.tag(RELEASE_VERSION, rc)
+          })
+        common.conditionalStage(
+          stage_name: "Update RC Branch",
+          stage: {
+            release.update_rc_branch(REPO_URL, MAINLINE, rc)
```
evrardjp: What does this do? I am clueless with the wording.

rpc_jobs/release.yml
```diff
@@ -0,0 +1,68 @@
...
+      common.shared_slave(){
+        rc = "${MAINLINE}-rc"
+        stage("Configure Git"){
+          release.configure_git()
+        }
+        stage("Clone"){
+          release.clone(REPO_URL)
+        }
+        common.conditionalStage(
+          stage_name: "Tag",
```
evrardjp: Tag should probably be not be a conditional stage.

rpc_jobs/release.yml
```diff
@@ -0,0 +1,68 @@
...
+          stage: {
+            release.tag(RELEASE_VERSION, rc)
+          })
+        common.conditionalStage(
+          stage_name: "Update RC Branch",
+          stage: {
+            release.update_rc_branch(REPO_URL, MAINLINE, rc)
+          })
+      }
```
evrardjp: Could this trigger other builds before continuing?

For example, Officially releasing a version x of rpc, should be triggering a change in rpc-o to bump the version number, and build its artifacts for the new version?

@odyssey4me what do you think?

pipeline_steps/release.groovy
```diff
@@ -0,0 +1,49 @@
...
+  print "Resetting ${rc} to head of ${mainline}"
+  sshagent (credentials:['rpc-jenkins-svc-github-ssh-key']){
+    sh """/bin/bash -xe
+      cd repo
+      mkdir -p ~/.ssh
+      ssh-keyscan github.com >> ~/.ssh/known_hosts
+      git fetch --all --tags
+      git checkout -b ${rc} || git checkout ${rc}
+      git reset --hard origin/${mainline}
+      git push -f origin ${rc}:${rc}
```
evrardjp: I'd be generally cautious with the -f.

rpc_jobs/release.yml
hughsaunders: This has to be conditional for creating the rc branch for projects or series that don't have it yet.

rpc_jobs/release.yml
hughsaunders: Updates the rc branch ;-)
It resets the rc branch to head of mainline and pushes that up.

evrardjp: Yes, I understood that part, because it's just exactly what the code does... I'd like to understand the context...

In which case should this happen? What's the reason to trigger it? How do we trigger it?

rpc_jobs/release.yml
hughsaunders: There are two parts to this.. 

1) preparation of the rc branch, which involves sha bumps and changing in any in repo versions references. This will be handled by https://rpc-openstack.atlassian.net/browse/RE-197

2) "A" release of project A should trigger a dependancy bump for all projects that rely on project "A". That doesn't have a card yet, but I'll create one.

pipeline_steps/release.groovy
hughsaunders: This has to have -f, as we can't guarantee a clean fast forward from head of rc to head of mainline. Its equivalent to cutting a new rc branch from head of mainline (which was an earlier iteration of the plan).

rpc_jobs/release.yml
hughsaunders: The release process as I understand it:

1. Tag head of stabilisation branch (this creates a release).  
1. Reset stabalisation branch to head of mainline. 
1. Propose updates to stabalisation branch (sha bumps, version bumps etc) 
1. Wait for stabalisation period then goto 1.  This function is implementing step 2. 

I've realised theres a problem with this logic, the sha bumps need to go into mainline first. However thats a problem to be solved in RE-197, not this PR.

hughsaunders: Commented in [RE-197](https://rpc-openstack.atlassian.net/browse/RE-197?focusedCommentId=167806&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-167806): 

>I think there may need to be two stages of stabalisation prep - pre and post branch cut.
>1. Propose Sha bumps and any other pre-cut changes to mainline (This happens a few days before release day to give time for gating etc).
>1. Tag head of stabilisation branch (this creates a release).
>1. Reset stabalisation branch to head of mainline (this brings in all new changes included SHA bumps proposed in step 1)
>1. Propose in-repo version updates to stabalisation branch, build artifacts etc.
>1. Wait for stabalisation period, then repeat.
>
>1 and 3 will be defined as scripts within each project's repo. Jenkins will handle executing this scripts and creating a PR if the script has dirtied the repo.

rpc_jobs/release.yml
```diff
@@ -0,0 +1,68 @@
...
+    parameters:
+      - rpc_gating_params
+      - string:
+          name: RELEASE_VERSION
+          description: |
+            The version that will be tagged by this job
+      - string:
+          name: NEXT_VERSION
+          description: |
+            The version that will be released after the next stabalisation
```
mattt416: Nit: stabilisation

pipeline_steps/release.groovy
```diff
@@ -0,0 +1,49 @@
...
+
+void tag(String version, String ref){
+  print "Tagging ${version} at ${ref}"
+  sshagent (credentials:['rpc-jenkins-svc-github-ssh-key']){
+    sh """/bin/bash -xe
+      cd repo
+      mkdir -p ~/.ssh
+      ssh-keyscan github.com >> ~/.ssh/known_hosts
+      git fetch --all --tags
+      git tag -a -m \"$version\" \"$version\" origin/$ref
```
mattt416: Can never quite remember -- but does this need to be `${version}` and `${ref}`?

evrardjp: We could omit the -m $version. But I think we need to keep at least the git tag version ref.

rpc_jobs/release.yml
evrardjp: That's definitely clearer.

I don't think the order is correct and the naming conventions are confusing.
We should probably also separate these into different jobs, because some may trigger at different occasions (ad-hoc things).

Let's take a step back. What can happen (examples afterwards for rpc-o):
a) Tag <whatever> in a stabilisation branch with no artefact update (example, doc fix, or minor bug that doesn't require new artefacts, because not everything is an artefact yet).
In this case, the process should probably triggered by the team, and we shouldn't see, as RE, the difference. IMO if the team is managing their repo, we shouldn't care, and we shouldn't build a job for it.
b) Tag <whatever> in a stabilisation branch with artefact update (example, osa change, new version of ELK...). In that case, the team managing their repo would like to have a way to trigger an artifact update that's external to their power. That's where a job comes in, giving a reference/tag. It still doesn't matter for this PR, as this is only caring about RE branching strategies in a global way. Each of the teams have to manage their own work of what's happening in a new tag (and in the meantime, we can help them RPC-O with the setup/automation of this work).
c) Regular RPC-O release triggering, so no ad-hoc work.
That's where this job should trigger, and it should happen simply as this:

1. List all the mainlines we know for all the products (RPC-O, K8s, MaaS)
2. Train is at the station for YYYY-MM (giving an example for a monthly cycle, where MM is next month). Check if our release repo has a review from each of the mainlines into a YYYY-MM file. These PR probably list a tag + sha for their mainline. If a team has no patch, that team missed the train.
3. Wait for a time to determine, which is max(artefact build time).
4. Because all these commits listed in 2. are very similar to the case b) above, I think we can assume that these commits would have build artefacts and updated their products accordingly (that's why we are doing step 3).
5. Then the YYYY-MM document is our manifest of the artefacts that were build for YYYY-MM. When release time (=MM) is reached, release (publishing _publicly_ all the artefacts somewhere)

So I am confused. When I read "Release job", I thought we were thinking of step 5.
Here it seems more like it's kinda b) ?

But let's focus on 5) for this PR, the "manual release job". What should this do?
IMO, building next files for manifest, checking availability of artefacts, publishing them, NOT TAGGING(done in b or 2), cut a new "stabilisation" branch YYYY-MM+1 from tag listed in YYYY-MM or from head of their branch YYYY-MM.
This makes our life simpler: We just read/write files and cut branches.

Am I completely wrong? Separating the tagging of a version from the "Release train" seem easier to me.

pipeline_steps/release.groovy
evrardjp: But that's detail.

odyssey4me: @mattt416 Are you asking due to this being a bash script in a groovy script? I think if we use {} here we may end up triggering a groovy macro?

rpc_jobs/release.yml
```diff
@@ -0,0 +1,100 @@
...
+            period.
+      - string:
+          name: MAINLINE
```
odyssey4me: nit: I think MAINLINE_BRANCH might be better here as it's a little clearer what it relates to.

pipeline_steps/release.groovy
mattt416: @odyssey4me Yep, that's what I'm referring to ... the function below (`update_rc_branch`) uses `${rc}` and `${mainline}`, for example.

rpc_jobs/release.yml
odyssey4me: @evrardjp This is covering https://rpc-openstack.atlassian.net/browse/RE-196 while the next step will be to cover https://rpc-openstack.atlassian.net/browse/RE-197 - I think you might want to take your discussion to those cards.

odyssey4me: @hughsaunders I think that it might be a good idea to describe a summarised version of the process in the actual job using comments to make it easier for anyone editing the code to see what's going on. If we've published a doc of some sort anywhere, a link to the doc is fine too.

pipeline_steps/release.groovy
odyssey4me: Yeah, itd be better if it was consistent. 👍

hughsaunders: Done

rpc_jobs/release.yml
hughsaunders: Done

rpc_jobs/release.yml
hughsaunders: Done

pipeline_steps/release.groovy
```diff
@@ -0,0 +1,24 @@
+void tag(String version, String ref){
+  print "Tagging ${version} at ${ref}"
+  sshagent (credentials:['rpc-jenkins-svc-github-ssh-key']){
+    sh """/bin/bash -xe
+      git fetch --all --tags
+      git tag -a -m \"${version}\" \"${version}\" origin/${ref}
+      git push --tags origin
```
git-harry: I'd prefer to see this only push the tag that has just been created - `git push origin \"${version}\"`. While it shouldn't make any difference, I'd prefer to exclude the possibility of the unexpected.

pipeline_steps/release.groovy
```diff
@@ -0,0 +1,24 @@
...
+}
+
+void update_rc_branch(String repo, String mainline, String rc){
+  print "Resetting ${rc} to head of ${mainline}"
+  sshagent (credentials:['rpc-jenkins-svc-github-ssh-key']){
+    sh """/bin/bash -xe
+      git fetch --all --tags
+      git checkout -b ${rc} || git checkout ${rc}
+      git reset --hard origin/${mainline}
+      git push -f origin ${rc}:${rc}
```
git-harry: This seems like it's going to conflict with GitHub protected branches, or is the intention to stop using that feature?

rpc_jobs/release.yml
```diff
@@ -0,0 +1,101 @@
...
+            The version that will be tagged by this job
+      - string:
+          name: NEXT_VERSION
+          description: |
+            The version that will be released after the next stabilisation
+            period.
+      - string:
+          name: MAINLINE_BRANCH
+          description: The MAINLINE_BRANCH branch to work from
+          default: master
```
git-harry: I'd rather not see a default here, everything else needs to be specified anyway. Lets force the user to provide them all to help prevent mistakes, in an automated world this strikes me as the kind of thing that would be easy to slip through.

pipeline_steps/release.groovy
hughsaunders: Good point, it will conflict if the rc branches are protected, which they probably should be.

odyssey4me: I've made the rc branch protected. Is there any way of allowing the jenkins user to push a tag with protection on? Alternatively, could the protection be removed by the jenkins user, the tag pushed, then protection added back?

pipeline_steps/release.groovy
hughsaunders: Done

rpc_jobs/release.yml
hughsaunders: Done

pipeline_steps/release.groovy
hughsaunders: I've checked rpc-openstack and rpc-maas, and we don't include admins in branch restrictions, so as long as Jenkins is an admin, force push should work. Will do some testing.

git-harry: The only reason admins aren't included is because someone forgot to add that flag, the always were previously.

hughsaunders: Turning off branch protection does loose all settings (which checks are required, who can push etc). So it may be better to toggle include-admins before and after pushing.

hughsaunders:
Added branch protection to ghutils, but haven't wired it in yet.

hughsaunders:
Hey @git-harry this has been revised and I'd appreciate a re-review, thanks!

rpc_jobs/release.yml
```diff
@@ -0,0 +1,97 @@
...
+          common.configure_git()
+        }
+        stage("Clone"){
+          common.clone_with_pr_refs(REPO_URL, MAINLINE_BRANCH)
+        }
+        common.conditionalStage(
+          stage_name: "Tag",
+          stage: {
+            release.tag(RELEASE_VERSION, rc)
+          })
```
git-harry: What about updating https://github.com/rcbops/rpc-openstack/releases? I believe that is still being done as part of the release process.

rpc_jobs/release.yml
```diff
@@ -0,0 +1,97 @@
...
+          description: |
+            Items to run. Options:
+              Tag,
+              Update RC Branch,
+              Mail Notification
+          default: >-
+            Tag,
+            Update RC Branch
+    dsl: |
+      stage("Check Parameters"){
```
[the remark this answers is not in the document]
hughsaunders: Double braces are only required in job templates, not jobs.

rpc_jobs/release.yml
hughsaunders: That will be done, but not by this job. That will be part of the SHA bump / Dependency update job.

rpc_jobs/release.yml
```diff
@@ -0,0 +1,97 @@
...
+        common.conditionalStage(
+          stage_name: "Mail Notification",
+          stage: {
+            // URL Forms:
+            repo_short = common.parse_github_repo_url(env.REPO_URL)['owner_repo']
+            if (env.MAIL_SUBJECT=="auto"){
+              env.MAIL_SUBJECT="${RELEASE_VERSION} of ${repo_short} Released"
+            }
+            if(env.MAIL_BODY=="auto"){
+              env.MAIL_BODY="The RPC team are pleased to announce that" \
```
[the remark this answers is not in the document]
hughsaunders: I use that because I didn't think it would template properly, but as the interpolation is done by groovy rather than jjb it should work. Will try.

pipeline_steps/release.groovy
```diff
@@ -0,0 +1,35 @@
...
+    """
+  }
+  github.set_branch_protection_admin_enforcement(
+    repo_dict['org'],
+    repo_dict['repo'],
+    rc,
+    "True")
+}
```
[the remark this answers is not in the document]
hughsaunders: Done

scripts/ghutils.py
```diff
@@ -92,5 +93,83 @@ def add_issue_url_to_pr(repo, pull_request_number, issue_key):
             raise Exception("There was a failure updating the pull request.")
 
 
+def branch_api_request(
+        repo,
+        branch,
+        method,
+        postfix="/enforce_admins"):
```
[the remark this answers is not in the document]
hughsaunders: Done

scripts/ghutils.py
```diff
@@ -92,5 +93,83 @@ def add_issue_url_to_pr(repo, pull_request_number, issue_key):
...
+    """Make Requests to the github branch protection api
+        Not supported by github3.py yet (6/9/2017)"""
```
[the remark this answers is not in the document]
hughsaunders: Done

scripts/ghutils.py
```diff
@@ -92,5 +93,83 @@ def add_issue_url_to_pr(repo, pull_request_number, issue_key):
...
+def set_admin_enforcement(ctx, branch, admin_enforcement_enabled):
+    repo = ctx.obj
+    if admin_enforcement_enabled == "True":
+        method = "POST"
+    else:
+        method = "DELETE"
+
+    if ctx.invoke(get_branch_protection, branch=branch):
+        print ("Setting branch protection admin"
+               " encforcement to {} ".format(admin_enforcement_enabled))
```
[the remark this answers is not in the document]
hughsaunders: Done

scripts/ghutils.py
```diff
@@ -92,5 +93,83 @@ def add_issue_url_to_pr(repo, pull_request_number, issue_key):
...
+        branch_api_request(repo, branch, method).raise_for_status()
+        ctx.invoke(get_admin_enforcement, branch=branch)
```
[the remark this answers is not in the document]
hughsaunders: Thought it would be useful to show the new state in the log. Removed.

scripts/ghutils.py
```diff
@@ -92,5 +93,83 @@ def add_issue_url_to_pr(repo, pull_request_number, issue_key):
...
+
+
+@cli.command()
+@click.pass_obj
+@click.option(
+    '--branch',
+    required=True,
+    help="Branch to query"
+)
+def get_admin_enforcement(repo, branch):
```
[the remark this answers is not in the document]
hughsaunders: While no longer called by another function, I'll leave this in as it can be used from the CLI.

scripts/ghutils.py
```diff
@@ -92,5 +93,83 @@ def add_issue_url_to_pr(repo, pull_request_number, issue_key):
...
+    required=True,
+    help="Branch to query"
+)
+def get_branch_protection(repo, branch):
+    response = branch_api_request(repo, branch, 'GET', postfix="")
+    if response.status_code == 200:
+        print "Branch Protection Enabled"
+        print json.dumps(response.json(), indent=4)
+        return True
+    else:
```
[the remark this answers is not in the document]
hughsaunders: Done

rpc_jobs/release.yml
```diff
@@ -0,0 +1,97 @@
...
+                            +" ${repo_short} version ${RELEASE_VERSION}" \
+                            +" is now available."
+            }
+            print("Sending notification mail To: ${env.MAIL_TO}, "
+                  +"Subject: ${env.MAIL_SUBJECT}, Body: ${env.MAIL_BODY}")
```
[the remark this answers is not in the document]
hughsaunders: screen shot 2017-09-07 at 09 49 48

Its not hideous, not great. Should be improved in future to include reno and rpc-differ. That also needs to be done for the github release.

rpc_jobs/release.yml
```diff
@@ -0,0 +1,97 @@
...
+          default: >-
+            Tag,
+            Update RC Branch
+    dsl: |
+      stage("Check Parameters"){
+        if (env.REPO_URL == "" || env.MAINLINE_BRANCH == ""){
+          throw new Exception(
+            "REPO_URL and MAINLINE_BRANCH parameters are required.")
+        }
+        if (!(REPO_URL ==~ /git@github.com:[\w-_]*\/[\w-_]*/)){
```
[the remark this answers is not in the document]
hughsaunders: Installed plugin, will try it out.

rpc_jobs/release.yml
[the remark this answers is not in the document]
git-harry: I was referring to the print line above and not the e-mail itself.

hughsaunders:
Test Job: https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/RE-Release/detail/RE-Release/43/pipeline ✅

Release that was created: https://github.com/wherenoworg/testrepo/releases/tag/1.0.8

pipeline_steps/release.groovy
```diff
@@ -0,0 +1,12 @@
+void tag(String version, String ref){
```
hughsaunders: This is no longer used, should be removed.

hughsaunders:
Re-implemented as a python script. Very little jenkins infrastructure, the job has one param `command` which includes the python script and all options. 

Test Job: https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/RE-Release/detail/RE-Release/59/pipeline ✅ 

V3... Ready for review

rpc_jobs/release.yml
```diff
@@ -0,0 +1,90 @@
...
+            # SCRIPT - a script to execute to generate release notes
+            # VERSION - the version to release
+            # PREVIOUS_VERSION - Used as base version for generating release notes
+            # NOTIFICATION_ADDRESS - where to send release notifications
+            python rpc-gating/scripts/release.py \
+                --debug \
+                --org rcbops \
+                --repo REPO \
+                clone \
+                  --ref master-rc \
```
mattt416: Nit, but is the indentation here off?

constraints.txt
```diff
@@ -1,27 +1,35 @@
...
+gitdb2==2.0.2
 github3.py==0.9.6
+GitPython==2.1.5
 idna==2.5
 influxdb==4.1.1
 ip-associations-python-novaclient-ext==0.2
 ipaddress==1.0.18
 iso8601==0.1.11
-jenkins-job-builder==2.0.0.0b1
+jenkins-job-builder==2.0.0.0b2
```
mattt416: Do we need to do further testing here since this is updating?

rpc_jobs/release.yml
hughsaunders: No, because ref is an arg for clone, but org and repo are top level args.

constraints.txt
hughsaunders: Good plan, we should build one of the periodic AIOs with b2 to check.

scripts/ghutils.py
```diff
@@ -92,5 +103,171 @@ def add_issue_url_to_pr(repo, pull_request_number, issue_key):
...
+            version,            # tag name
+            ref,                # tag reference
+            version,            # release name
+            body.read()         # release body
+        )
+        print "Release {} created.".format(version)
+    except github3.models.GitHubError as e:
+        print "Error creating release: {}".format(e)
+        if e.code == 422:
+            print "Failed to create release, tag already exists?"
```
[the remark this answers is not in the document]
hughsaunders: This particular call creates the tag and the release.

scripts/release.py
```diff
@@ -0,0 +1,103 @@
...
+    "--ref",
+    help="Ref that will be released. Release notes scripts compare from "
+         "--prev-version to --ref."
+)
+@click.option(
+    "--prev-version",
+    help="Last released version. Release notes scripts should compare "
+         "prev-version to ref"
+)
+def generate_release_notes(script, rnfile, text, ref, prev_version):
```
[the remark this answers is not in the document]
hughsaunders: Yes mumble mumble not supported by click.

scripts/ghutils.py
```diff
@@ -92,5 +109,210 @@ def add_issue_url_to_pr(repo, pull_request_number, issue_key):
...
+        if e.code == 404:
+            print "Failed to create release, Jenkins lacks repo perms?"
+            raise SystemExit(6)
+        else:
+            raise e
+
+
+@cli.command()
+@click.option("--url")
+@click.option("--ref", help="ref to checkout", default="master")
```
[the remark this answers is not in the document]
hughsaunders: Removed

scripts/release.py
```diff
@@ -0,0 +1,103 @@
...
+    release_notes = ""
```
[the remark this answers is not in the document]
hughsaunders: Removed

scripts/release.py
```diff
@@ -0,0 +1,103 @@
...
+    if script:
+        ctx_obj = click.get_current_context().obj
+        try:
+            ref = ctx_obj.rc_ref
```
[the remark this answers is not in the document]
hughsaunders: Refactored that into a function as the patern is repeated, added comment in the function.

scripts/release.py
```diff
@@ -0,0 +1,103 @@
...
+        except AttributeError:
+            if not ref:
+                raise ValueError("--ref is required when using --script.")
+        logger.debug("Generating release notes from script: {}".format(script))
+        try:
+            sub_env = os.environ.copy()
+            sub_env["PREVIOUS_VERSION"] = prev_version
```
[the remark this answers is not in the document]
hughsaunders: I think the validation should be done in the script that receives the parameter, this module doesn't necessarily know the range of acceptable values. Though we could try and validate git refs.

scripts/release.py
```diff
@@ -0,0 +1,103 @@
...
+            release_notes = out
+        except Exception as e:
+            logger.error(traceback.format_exc())
+            logger.error("Failed to generate release notes: {}".format(e))
+            click.get_current_context().exit(-1)
+    elif rnfile:
+        logger.debug("Reading release notes from file: {}".format(rnfile))
+
+        try:
+            release_notes = open(rnfile, "r").read()
```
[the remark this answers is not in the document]
hughsaunders: Context manager added.

scripts/release.py
```diff
@@ -0,0 +1,103 @@
...
+        except IOError as e:
+            logger.error(traceback.format_exc())
+            logger.error("Failed to read release notes file: {}".format(e))
+            click.get_current_context().exit(-1)
+    elif text:
```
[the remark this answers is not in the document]
hughsaunders: The simplest way to not generate release notes would be to omit the generate_release_notes command.

hughsaunders:
Example release created: https://github.com/wherenoworg/rpc-openstack/releases/tag/ht1.27
Job that created it: https://rpc.jenkins.cit.rackspace.net/job/RE-Release/18/parameters/

hughsaunders:
Another success this time, with the API in line with https://rpc-openstack.atlassian.net/wiki/spaces/RE/pages/19005457/RE+for+Projects

https://rpc.jenkins.cit.rackspace.net/job/RE-Release/26/parameters/ ✅ 
Release: https://github.com/wherenoworg/rpc-openstack/releases/tag/ht1.29

scripts/ghutils.py
```diff
@@ -92,5 +111,219 @@ def add_issue_url_to_pr(repo, pull_request_number, issue_key):
...
+@click.pass_obj
+@click.option(
+    '--version',
+    required=True,
+    help="Symbollic name of Release (eg r14.1.99)"
+)
+@click.option(
+    '--ref',
+    help="Reference to create release from (branch, SHA etc)"
+         " May be omitted if supplied to clone earlier in the chain."
```
evrardjp: Isn't that risky?
Shouldn't we always ensure the reference is set?

scripts/ghutils.py
```diff
@@ -92,5 +111,219 @@ def add_issue_url_to_pr(repo, pull_request_number, issue_key):
...
+)
+@click.option(
+    '--bodyfile',
+    help="File containing release message body"
+    # Can't use type=click.File because the file may not exist on startup
```
evrardjp: Maybe this whole CLI that does all the steps at the same time should then broken into smaller bits, to then rely on the existing click function and for ease of use?

scripts/notifications.py
```diff
@@ -0,0 +1,111 @@
...
+        auth=("api", mailgun_api_key),
+        data={"from": "RPC-Jenkins@rackspace.com",
+              "to": [to],
+              "subject": subject,
+              "text": body
+              }
+    )
+
+
+cli.add_command(mail)
```
evrardjp: see comments above.

job_dsl/release.groovy
```diff
@@ -0,0 +1,27 @@
...
+  }
+  stage("Release"){
+    withCredentials([
+      string(
+        credentialsId: 'rpc-jenkins-svc-github-pat',
+        variable: 'PAT'
+      ),
+      string(
+        credentialsId: 'mailgun_hughsaunders_endpoint',
+        variable: 'MAILGUN_ENDPOINT'
```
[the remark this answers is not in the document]
hughsaunders: Its currently my domain, if account with user/pass in passwordsafe. No domain verified because I didn't want to use my personal domain. That does mean that every recipient must be added to the test recipients list. 

I realise that account isn't optimal, it would be simple to change accounts if someone comes up with a better one.

scripts/notifications.py
```diff
@@ -0,0 +1,111 @@
...
+            raise ValueError("No value found for {var_name} ({e})".format(
+                var_name=var_name, e=e))
+
+    logger.debug("Try context Result: {var_name} --> {result}".format(
+        var_name=var_name, result=var
+    ))
+    return var
+
+
+@click.command()
```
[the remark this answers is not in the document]
hughsaunders: Done

scripts/notifications.py
hughsaunders: Done

scripts/ghutils.py
```diff
@@ -92,5 +111,219 @@ def add_issue_url_to_pr(repo, pull_request_number, issue_key):
...
+    assert mainline_sha == repo.branch(rc).commit.sha
+    logger.debug("rc branch update complete")
+
+
+@cli.command()
+@click.pass_obj
+@click.option(
+    '--version',
+    required=True,
+    help="Symbollic name of Release (eg r14.1.99)"
```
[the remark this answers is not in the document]
hughsaunders: Done

scripts/ghutils.py
hughsaunders: An exception will be thrown if exception if ref is not supplied, either via create_release --ref or to clone earlier.

scripts/release.py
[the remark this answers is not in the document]
git-harry: Sure but it's nice to have a consistent interface.

scripts/ghutils.py
hughsaunders: Yeah, That would only be a change to the job, as the python modules can be used independently in their current state.

scripts/release.py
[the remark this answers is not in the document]
git-harry: You could just put a test in the function, e.g. default them all to `None` and have something like
```python
if len(filter(lambda x: x is not None, (script, rnfile, text))) != 1:
    BOOM!
```

hughsaunders:
Test job after modifications requested by @evrardjp : https://rpc.jenkins.cit.rackspace.net/job/RE-Release/29/parameters/ ✅  
Resulting release: https://github.com/wherenoworg/rpc-openstack/releases/tag/ht1.34

hughsaunders:
Issue link job failed because it used the master version of gating venv, with re-196 branch of scripts, so its missing the git module. That will be fixed once this is merged into master. That will invalidate the current master venv and a new one will be built.

evrardjp:
I'd be happy to see other reviewers on this, because this is an important step, and it already has a +1.

I am ok with it at first sight.

scripts/release.py
```diff
@@ -0,0 +1,152 @@
...
+        logger.debug("Release notes supplied inline: {}".format(text))
+        with open(dst_file, "w") as out:
+            out.write(text)
+    else:
+        logger.error("One of script, text or file must be specified. "
+                     "No release notes generated")
+        click.get_current_context().exit(-1)
+    with open(dst_file, "r") as notes:
+        ctx_obj.release_notes = notes.read()
```
git-harry: I would leave this out. The only place it's used, `create_release`, can read the file itself. This mechanism makes `create_release` harder to read.

scripts/ghutils.py
```diff
@@ -92,5 +111,219 @@ def add_issue_url_to_pr(repo, pull_request_number, issue_key):
...
+        branch_protection_response = branch_api_request(repo, rc, 'GET')
+        if branch_protection_response.status_code == 200:
+            # rc branch exists and protection enabled
+            logger.debug("Branch {} has protection enabled".format(rc))
+            branch_protection_enabled = True
+            # disable branch protection
+            r = branch_api_request(repo, rc, 'DELETE')
+            r.raise_for_status()
+            logger.debug("Branch protection disabled")
+        elif branch_protection_response.status_code == 404:
```
git-harry: Why would branch protection not be enabled? I don't think this should be ignored, it should be required.

scripts/release.py
hughsaunders: Generate release notes exposes all the mechanisms for obtaining release notes, from a script, text or file. It made sense to me to keep them all as options within one command.

git-harry: I'm don't follow what you are saying here. I'm not suggesting removing support for script/file/text. I'm referring to the use of `ctx_obj` which doesn't add any functionality in this case given that `create_release` will read from a file.

rpc_jobs/release.yml
```diff
@@ -0,0 +1,120 @@
...
+    parameters:
+      - rpc_gating_params
+      - text:
+          name: COMMAND
+          default: |
+            # the following parameters must be replaced with your values:
+            # REPO - the repo URL for the project to be released
+            # MAINLINE - the mainline branch, usually master
+            # REF - the rc/stabilisaiton branch, usually master-rc.
+            # VERSION - Symbollic name of the release, not the ref to be released
```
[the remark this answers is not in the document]
hughsaunders: Done

rpc_jobs/release.yml
```diff
@@ -0,0 +1,120 @@
...
+            # Note: NOTIFICATION_ADDRESS must be in authorised recipients
+            #       list for the mailgun account, if using mailgun.
+            python rpc-gating/scripts/release.py \
+                --debug \
+                --org rcbops \
+                --repo REPO \
+                clone \
+                  --ref RC-BRANCH \
+                generate_release_notes \
+                    --script "gating/generate_release_notes/pre" \
```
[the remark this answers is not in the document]
hughsaunders: That confluence page is the definitive documentation at this point. Those three scripts were added to the job to comply with the standard set out on that page.

scripts/ghutils.py
hughsaunders: It is required, but at this point we are trying to disable it, if its already disabled, nothing to do. 
Once the push has happened, branch protection will be enabled regardless of whether it was enabled previously.

scripts/ghutils.py
```diff
@@ -92,5 +111,219 @@ def add_issue_url_to_pr(repo, pull_request_number, issue_key):
...
+    try:
+        origin = repo.remotes.origin
+        origin.set_url(url)
+    except Exception as e:
+        origin = repo.create_remote('origin', url)
+    repo.git.fetch(["-u", "-v", "-f", url] + refspec.split())
+    try:
+        getattr(origin.refs, ref).checkout()
+    except AttributeError as e:
+        print ("Ref {ref} not found in {url}".format(
```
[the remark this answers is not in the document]
hughsaunders: Fixed, and 4 others.

scripts/notifications.py
```diff
@@ -0,0 +1,107 @@
...
+         var_name: Name of the option
+         context_attr: attribute to attempt to read from the context object.
+    """
+    logger.debug("Try context: {var_name}:{var}, {c_attr}".format(
+        var_name=var_name, var=var, c_attr=context_attr
+    ))
+    try:
+        var = getattr(ctx_obj, context_attr)
+    except AttributeError as e:
+        if not var:
```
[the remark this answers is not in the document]
hughsaunders: Replaced with `is not None`.

scripts/release.py
hughsaunders: The functionality added is that all the release notes sources are options to the same command, rather than two on generate_release notes and one on create release. For a user it probably makes sense to have all the options for adding release notes in one place?

scripts/release.py
```diff
@@ -0,0 +1,152 @@
...
+                        logger.error(
+                            "Script {s} failed. (Return Code: {rc})".format(
+                                s=script_path, rc=proc.returncode
+                            ))
+                        click.get_current_context().exit(-1)
+                else:
+                    logger.debug("Script not found: {}".format(script_path))
+                    if script.endswith("/run"):
+                        logger.error(
+                            "Failed to generate relese notes:"
```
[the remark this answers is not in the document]
hughsaunders: Done

scripts/release.py
```diff
@@ -0,0 +1,152 @@
...
+         " in the chain."
+)
+def generate_release_notes(scripts, rnfile, text, ref, prev_version,
+                           clone_dir):
+    ctx_obj = click.get_current_context().obj
+    clone_dir = try_context(ctx_obj, clone_dir, "clone_dir", "clone_dir")
+    # cwd == $WORKSPACE
+    artifact_dir = "{cwd}/artifacts".format(cwd=os.getcwd())
+    dst_file = "{artifact_dir}/release_notes".format(artifact_dir=artifact_dir)
+    os.system("mkdir -p {}".format(artifact_dir))
```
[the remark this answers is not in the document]
hughsaunders: os.makedirs doesn't have -p equivalent functionality:
```Recursive directory creation function. Like mkdir(), but makes all intermediate-level directories needed to contain the leaf directory. Raises an error exception if the leaf directory already exists or cannot be created. The default mode is 0777 (octal).```

Switched os.system for subprocess.check_call

File: scripts/release.py (hunk 40 above)
Comment (git-harry) [reply scripts/release.py]: I think you are misunderstanding me, I'm referring to the following lines and the associated code in other functions:
```python
    with open(dst_file, "r") as notes:
        ctx_obj.release_notes = notes.read()
```
You've coupled `generate_release_notes` with `create_release`. `generate_release_notes` hardcodes the filename where it writes the release notes but also adds it to `ctx_obj`. `create_release` checks `ctx_obj` but also takes a filename `bodyfile`, which is specified in `rpc_jobs/release.yml`, although this is not used.

I think it would be simpler if both `generate_release_notes` and `create_release` simply communicated via the file and so `generate_release_notes` should take the same `--bodyfile` argument instead of hardcoding the path.

scripts/release.py
```diff
@@ -0,0 +1,150 @@
...
+            clone_dir=clone_dir
+        )
+        logger.debug("script work dir: {}".format(script_work_dir))
+        for script in scripts:
+            try:
+                script_path = "{script_work_dir}/{script}".format(
+                    script_work_dir=script_work_dir,
+                    script=script
+                )
+                if os.path.exists(script_path):
```
git-harry: Along the lines of my other comment, the script(s) should take the release notes path as an argument or maybe just write to stdout so that we are specifying a specific location in less places. That allows us to change how this is all done without affecting the users.

scripts/ghutils.py
git-harry: Yeah, I missed that.

rpc_jobs/release.yml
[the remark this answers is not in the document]
git-harry: `scripts/release.py` does not enforce the naming convention, it assumes it, but gives a specific error message when a script called `run` is used. Either we should not allow arbitrarily named scripts to run or error messages should be generalised.

scripts/release.py
hughsaunders: I agree that release.yml should not specify `--bodyfile` for create_release. I think that is leftover from before the context object was used. `create_release` could be used independently of generate_release_notes by using the bodyfile argument you mention. 

The path is hard coded because its part of the specification in RE for projects. Its very simple specify a path relative to the workspace where a script must write release notes to. I initially used stdout, but that made implementation more complicated as you have to be careful about unintended leaks to stdout.

Maybe it would be better to keep all the hard coded paths in the job, rather than in the python? As the pre/run/post scripts are currently.

scripts/release.py
hughsaunders: See https://github.com/rcbops/rpc-gating/pull/386/files#r141568148

I'll add the path as an option in the python implementation then pass in the path from the job.

rpc_jobs/release.yml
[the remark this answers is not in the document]
hughsaunders: I'll add a --required-script parameter so that release.py doesn't have to know which scripts are required.

scripts/release.py
git-harry: I think you've agreed with me while having a completely different opinion.
- the context object is a hinderance not a help
- a specification that requires a script to write to a certain location is less simple than one that requires a script to write to a configurable location. If you can guarantee that location will never change, then hardcode it but I don't see how it's ever possible to make that guarantee.
- Yes, lets have one place where this config is set.

rpc_jobs/release.yml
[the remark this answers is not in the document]
git-harry: Okay, but then I don't see how `release.py` can continue to have tests such as `if script.endswith("/run"):`.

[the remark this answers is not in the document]
hughsaunders: Specifying whether a script is required in parameters would be instead of release.py testing the script path.  I had a look at implementing --required-script, but that would make ordering difficult. 

```
generate_release_notes
  --script pre
  --required-script run
  --script post
```

The scripts and required-scripts would get collected into separate arrays, and the ordering would be lost.

Will try an `optional:` modifier within the script path. Eg `--script optional:/opt/myoptionalscript --script /opt/myrequiredscript` This ordering can be maintained while still communicating if the script is required or not.

hughsaunders:
Successfull test: https://rpc.jenkins.cit.rackspace.net/job/RE-Release/ ✅ 

@git-harry please re-review.

scripts/ghutils.py
```diff
@@ -92,5 +111,220 @@ def add_issue_url_to_pr(repo, pull_request_number, issue_key):
...
+    branch_protection_enabled = False
+
+    # check if branch exists
+    if rc in (b.name for b in repo.iter_branches()):
+        logger.debug("Branch {} exists".format(rc))
+        # rc branch exists
+        branch_protection_response = branch_api_request(repo, rc, 'GET')
+        if branch_protection_response.status_code == 200:
+            # rc branch exists and protection enabled
+            logger.debug("Branch {} has protection enabled".format(rc))
```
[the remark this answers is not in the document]
hughsaunders: Done.

hughsaunders:
https://rpc.jenkins.cit.rackspace.net/job/RE-Release/41/console ▶️

hughsaunders:
Logging as requested by @git-harry  
```
DEBUG:ghutils:Branch RE-196 has protection enabled, config: {u'url': u'https://api.github.com/repos/wherenoworg/rpc-openstack/branches/RE-196/protection', u'enforce_admins': {u'url': u'https://api.github.com/repos/wherenoworg/rpc-openstack/branches/RE-196/protection/enforce_admins', u'enabled': False}}
```

PR #389, opened 2017-08-14
Add gate job for ops fabric chatbot

supermari0:
This is my first attempt at adding a gate job for the
rcbops/ops-fabric-chatbot repo. I want this job to:

- Install any needed dependencies (ops-fabric-chatbot/install-deps.sh)
- Run tests (ops-fabric-chatbot/run-tests.sh)

I would like this job to fail if either of the above commands return
exit code != 0. A timeout of 5 minutes would be good, too, but I haven't
implemented that yet.


supermari0:
recheck

supermari0:
switching this to a different PR from a feature branch in this repo

PR #390, opened 2017-08-14
TURTLES-67 Add gate job for ops fabric chatbot

supermari0:
This is my first attempt at adding a gate job for the
rcbops/ops-fabric-chatbot repo. I want this job to:

- Install any needed dependencies (ops-fabric-chatbot/install-deps.sh)
- Run tests (ops-fabric-chatbot/run-tests.sh)

I would like this job to fail if either of the above commands return
exit code != 0. A timeout of 5 minutes would be good, too, but I haven't
implemented that yet.

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -439,22 +439,46 @@ def prepareRpcGit(String branch = "auto", String dest = "/opt"){
 
     print("Repo: ${env.RPC_REPO} Branch: ${branch}")
 
-    // checkout used instead of git as a custom refspec is required
-    // to checkout pull requests
-    checkout([$class: 'GitSCM',
-      branches: [[name: branch]],
-      doGenerateSubmoduleConfigurations: false,
-      extensions: [[$class: 'CleanCheckout']],
-      submoduleCfg: [],
-      userRemoteConfigs: [
-        [
-          url: env.RPC_REPO,
-          refspec: '+refs/pull/*:refs/remotes/origin/pr/* +refs/heads/*:refs/remotes/origin/*'
-        ]
-      ]
-    ]) // checkout
-    sh "git submodule update --init"
+    clone_with_pr_refs(env.RPC_REPO, branch)
   } // dir
+}
+
+// Clone repo with Refspecs required for PRs.
+// Shouldn't need to supply any params to checkout a PR merged with the base.
+// Uses shell+git to avoid hostname verification failures with
+// the built in git scm step.
+// Use init + fetch instead of clone so that the repo can
+// be cloned into a non-empty directory. Thats added for
+// compatibility with the jenkins git scm step.
+void clone_with_pr_refs(
+  String repo="https://github.com/${env.ghprbGhRepository}.git",
+  String ref="origin/pr/${env.ghprbPullId}/merge",
+  String refspec='+refs/pull/\\*:refs/remotes/origin/pr/\\*'\
+                +' +refs/heads/\\*:refs/remotes/origin/\\*'
+){
+  if(repo == "https://github.com/.git"){
+    throw new Exception(
+      "repo not supplied to common.clone_with_pr_refs or env.ghprbGhRepository"\
+      + " not set."
+    )
+  }
+  if(ref == "origin/pr//merge"){
+    throw new Exception(
+      "ref not supplied to common.clone_with_pr_refs or env.ghprbPullID not "\
+      + "set, attempting to checkout PR for a periodic build?")
+  }
+  print "Cloning Repo: ${repo}@${ref}"
+  sshagent (credentials:['rpc-jenkins-svc-github-ssh-key']){
+    sh """/bin/bash -xe
+      # use init + fetch to avoid the "dir not empty git fail"
+      git init .
+      git remote add origin "${repo}"
+      # Don't quote refspec as it should be separate args to git.
+      git fetch --tags origin ${refspec}
+      git checkout ${ref}
+      git submodule update --init
+    """
+  }
 }
 
 /* Set mtime to a constant value as git doesn't track mtimes but
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -247,6 +247,11 @@ def run_script(Map args) {
  *    - STAGES: String list of stages that should be run
  */
 def conditionalStage(Map args){
+  if (env.STAGES == null){
+  throw new Exception(
+    "ConditionalStage used without STAGES env var existing."\
+    + " Ensure the top level job has a string param called STAGES.")
+  }
   stage(args.stage_name){
     if (env.STAGES.contains(args.stage_name)){
         print "Stage Start: ${args.stage_name}"
@@ -271,6 +276,11 @@ def conditionalStage(Map args){
  * run in a stage.
  */
 def conditionalStep(Map args){
+  if (env.STAGES == null){
+    throw new Exception(
+      "ConditionalStep used without STAGES env var existing."\
+      + " Ensure the top level job has a string param called STAGES.")
+  }
   if (env.STAGES.contains(args.step_name)){
       print "Step Start: ${args.step_name}"
       args.step()
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -481,6 +481,30 @@ void clone_with_pr_refs(
   }
 }
 
+void configure_git(){
+  print "Configuring Git"
+  withCredentials([
+    usernamePassword(
+    credentialsId: "github_account_rpc_jenkins_svc",
+    usernameVariable: "GIT_USER",
+    passwordVariable: "GIT_PASS"
+     )
+  ]){
+    // credentials store created to ensure that non public repos
+    // can be cloned when specified as https:// urls.
+    // Ssh auth is handled in clone_with_pr_refs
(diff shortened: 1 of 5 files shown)


hughsaunders:
Hey @supermari0, thanks for this PR :)
Looks good, I've made a few changes to use the Jenkins DSL/pipeline script rather than builder blocks. Also I've attempted to extract out the git clone with pr refs code from [common.prepareRpcGit](https://github.com/rcbops/rpc-gating/blob/master/pipeline_steps/common.groovy#L424), as we need to test projects beside rpc-o. (Note to self: that can also be used in irr_role_tests.groovy).

It's not finished (hence the wip tag) so feel free to poke it, or I'll keep going tomorrow.

pipeline_steps/common.groovy
```diff
@@ -439,24 +439,50 @@ def prepareRpcGit(String branch = "auto", String dest = "/opt"){
...
+  print "Cloning Repo: ${repo}@${ref}"
+  sshagent (credentials:['rpc-jenkins-svc-github-ssh-key']){
+    sh """/bin/bash -xe
+      mkdir -p ~/.ssh
+      ssh-keyscan github.com >> ~/.ssh/known_hosts
+      # use init + fetch to avoid the "dir not empty git fail"
+      git init .
+      git remote add origin "${repo}"
+      # Don't quote refspec as it should be separate args to git.
+      git fetch --tags origin ${refspec}
```
supermari0: Will this pull via HTTPS?

I noticed that this is using the service SSH key, but it looks like the repo is defined on L454 as using HTTPS. I believe that clone via SSH is needed for private repos.

supermari0:
@hughsaunders Mostly LGTM, one comment about clone with SSH vs. HTTPS

pipeline_steps/common.groovy
hughsaunders: Private repos can be cloned via ssh or http as long as its an authenticated connection. I want to support both types of URL, so I've added the ssh agent for ssh urls, and a git credentials store for https urls.

hughsaunders:
successfull build: https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/ops-fabric-chatbot-gate/detail/ops-fabric-chatbot-gate/10/pipeline

hughsaunders:
@mattt416 As I'm now an author, could you review this? Thanks!

pipeline_steps/common.groovy
```diff
@@ -439,24 +449,72 @@ def prepareRpcGit(String branch = "auto", String dest = "/opt"){
...
+// the built in git scm step.
+// Use init + fetch instead of clone so that the repo can
+// be cloned into a non-empty directory. Thats added for
+// compatibility with the jenkins git scm step.
+void clone_with_pr_refs(
+  String repo="https://github.com/${env.ghprbGhRepository}.git",
+  String ref="origin/pr/${env.ghprbPullId}/merge",
+  String refspec='+refs/pull/\\*:refs/remotes/origin/pr/\\*'\
+                +' +refs/heads/\\*:refs/remotes/origin/\\*'
+){
```
mattt416: Is it cleaner to just pass in the env vars as args and then check if they're empty?

pipeline_steps/common.groovy
```diff
@@ -439,24 +449,72 @@ def prepareRpcGit(String branch = "auto", String dest = "/opt"){
...
-      ]
-    ]) // checkout
-    sh "git submodule update --init"
+    clone_with_pr_refs(env.RPC_REPO, branch)
   } // dir
 }
 
+// Clone repo with Refspecs required for PRs.
+// Should't need to supply any params to checkout a PR merged with the base.
+// Uses shell+git to avoid hostname verification failures with
```
mattt416: Sounds like https://stackoverflow.com/questions/38461705/checkout-jenkins-pipeline-git-scm-with-credentials.  The failure there was because the credentials passed in did not match the URL (ie. username/key for SSH).  I'd prefer to use checkout if at all possible!

rpc_jobs/ops_fabric_chatbot.yml
```diff
@@ -0,0 +1,53 @@
...
+                - rcbops
+            github-hooks: true
+            trigger-phrase: 'recheck'
+            only-trigger-phrase: false
+            auth-id: "github_account_rpc_jenkins_svc"
+            status-context: 'ops-fabric-chatbot-gate'
+    dsl: |
+      library "rpc-gating@${env.RPC_GATING_BRANCH}"
+      common.shared_slave(){
+        pubcloud.runonpubcloud(){
```
mattt416: Did you try running this in a container, to avoid having to spin up a cloud instance?

hughsaunders: No, but @supermari0 mentioned they have a docker registry, and that could be done. Theres no docker file yet though, so maybe that could be a future iteration? Alternatively I could create a bare-bones docker file tomorrow.

pipeline_steps/common.groovy
hughsaunders: I got the credentials working with the checkout step, the problem was host key verification, not user key/creds. The solutions are mostly "add this hostkey to the jenkins user's known hosts" but I don't want jobs making changes outside of their workspace and I don't want to have to do any manual setup of slaves. Using shell+git to clone keeps all the config within the workspace. 

I guess we could have something  in the long-running-slave job, or in the setup-jenkins-slave playbook that added the github host keys to the jenkins user ? But i don't see whats so wrong with shelling out to git, thats what checkout + git plugin does anyway 🤷‍♂️

pipeline_steps/common.groovy
hughsaunders: The neat thing about not passing the env vars is that you don't need to use them, for example pull id doesn't make sense for periodics, so a different structure for repo and ref can be used. See the usage in prepareRpcGit for example, sometimes ref is simply "master".

hughsaunders:
This now depends on https://github.com/rcbops/ops-fabric-chatbot/pull/8

hughsaunders:
Successfull Test run: https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/ops-fabric-chatbot-gate/detail/ops-fabric-chatbot-gate/19/pipeline/46

Lint and Unit test results reported to Jenkins: https://rpc.jenkins.cit.rackspace.net/job/ops-fabric-chatbot-gate/19/testReport/

hughsaunders:
rcbops/ops-fabric-chatbot#8 is merged.

pipeline_steps/common.groovy
supermari0: :+1:

hughsaunders:
Test job after the removal of https creds (beb413dea52fe1c15d72d3fae328374ba2897e3a) : https://rpc.jenkins.cit.rackspace.net/job/ops-fabric-chatbot-gate/

pipeline_steps/common.groovy
hughsaunders: I've removed the https creds functionality due to security concerns raised by @mattt416. Https connections will be unauthenticated, ssh should be used when an authenticated connection is required. The ops-fabric-chatbot repo is not public so an authenticated connection is required for cloning, which is fine as the [default protocol is now ssh](https://github.com/rcbops/rpc-gating/pull/390/files#diff-0f12f66d6e63bf3b031477881aedeaadR466).

pipeline_steps/common.groovy
```diff
@@ -439,24 +449,64 @@ def prepareRpcGit(String branch = "auto", String dest = "/opt"){
...
-        ]
-      ]
-    ]) // checkout
-    sh "git submodule update --init"
+    clone_with_pr_refs(env.RPC_REPO, branch)
   } // dir
 }
 
+// Clone repo with Refspecs required for PRs.
+// Should't need to supply any params to checkout a PR merged with the base.
```
mkam: Typo with shouldn't.

rpc_jobs/ops_fabric_chatbot.yml
```diff
@@ -0,0 +1,51 @@
...
+    concurrent: true
+    properties:
+      - github:
+          url: https://github.com/rcbops/ops-fabric-chatbot/
+      - build-discarder:
+          num-to-keep: 30
+    parameters:
+      - rpc_gating_params
+    triggers:
+        - github-pull-request:
```
mkam: The spacing for the github-pull-request section looks like it's over two spaces more than it needs to be.

rpc_jobs/ops_fabric_chatbot.yml
```diff
@@ -0,0 +1,51 @@
+- job:
+    name: ops-fabric-chatbot-gate
+    description: "lint and unit tests for ops-fabric-chatbot"
+    disabled: false
+    display-name: "ops-fabric-chatbot lint and unit tests"
```
mkam: I'd prefer if the job name followed the naming conventions in the README.

pipeline_steps/common.groovy
hughsaunders: Done

rpc_jobs/ops_fabric_chatbot.yml
hughsaunders: Done

rpc_jobs/ops_fabric_chatbot.yml
hughsaunders: Done

PR #391, opened 2017-08-15
RE-199 Improve test for venv availability

hughsaunders:
Instead of checking the index file, try and download and unpack the
venv tar, if that succeeds unpack it, else create the venv locally.

This is to prevent problems where the tar is listed in the index
but is not fully written/synced, so the job attempts to download it
but either that or unpacking fails.

Issue: [RE-199](https://rpc-openstack.atlassian.net/browse/RE-199)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -55,9 +55,7 @@ def void install_ansible(){
       ansible-galaxy install -r rpc-gating/role_requirements.yml -p rpc-gating/playbooks/roles
     }
 
-    download_venv(){
-      curl -s "\${REPO_BASE}/rpcgatingvenv_\${SHA}.tbz" > venv.tbz
-      tar xjfp venv.tbz
+    unpack_venv(){
       op=\$(cat .venv/original_venv_path) # Original Path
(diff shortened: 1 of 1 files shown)


pipeline_steps/common.groovy
```diff
@@ -56,8 +56,6 @@ def void install_ansible(){
     }
 
     download_venv(){
```
mattt416: This function name is now a little misleading -- any value in cleaning that up?

pipeline_steps/common.groovy
```diff
@@ -78,13 +76,13 @@ def void install_ansible(){
       SHA=\$(git rev-parse HEAD)
     popd
 
-    curl -s "\${REPO_BASE}/index" > index ||:
```
mattt416: Do we still need to create/update the index in `rpc_jobs/build_gating_venv.yml` then?

hughsaunders: I guess not, but no harm in maintaining an index file.

pipeline_steps/common.groovy
hughsaunders: I agree, will rename unpack_venv

PR #392, opened 2017-08-16
RE-201 Use docker cli's filtering instead of awk

hughsaunders:
This is likely to be more reliable than awking the output.
The awk version missed some containers as they didn't contain
"Exited" in the status field even though they were no longer running.

Issue: [RE-201](https://rpc-openstack.atlassian.net/browse/RE-201)

### DIFF
--- a/scripts/docker_cleanup.sh
+++ b/scripts/docker_cleanup.sh
@@ -2,7 +2,8 @@
 
 set -xe
 echo "Remove exited containers"
-docker ps -a |awk '/Exited/{print $1}' |while read cid; do docker rm $cid||:; done
+#https://docs.docker.com/engine/reference/commandline/ps/#examples
+(docker ps -a -q -f "status=exited"; docker ps -a -q -f "status=dead")| while read cid; do docker rm $cid||:; done
 
 echo "Remove old Images"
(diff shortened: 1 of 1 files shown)


scripts/docker_cleanup.sh
```diff
@@ -2,7 +2,8 @@
 
 set -xe
 echo "Remove exited containers"
-docker ps -a |awk '/Exited/{print $1}' |while read cid; do docker rm $cid||:; done
+#https://docs.docker.com/engine/reference/commandline/ps/#examples
+(docker ps -a -q -f "status=exited"; docker ps -a -q -f "status=dead")| while read cid; do docker rm $cid||:; done
```
mattt416: Does `docker ps -a -q -f "status=exited" -f "status=dead"` not work?

hughsaunders: No :(

git-harry: I just tried `(docker ps -a -q -f "status=running" -f "status=created")| while read cid; do echo $cid||:; done`, specifying multiple filters appeared to work fine for me. That said, this is a fairly minor point.

PR #395, opened 2017-08-17
RE-205 Only run tempest_install() on leapfrog

mattt416:
If you now attempt to run the Destroy Slave and Cleanup stages on their
own, the build will attempt to first install tempest.  This may or may
not succeed depending on the state of the AIO.

Since the initial tempest resources are only required for the resource
generation tests run in a leapfrog upgrade, we move this line to within
the leapfrog block.

### DIFF
--- a/rpc_jobs/rpc_aio.yml
+++ b/rpc_jobs/rpc_aio.yml
@@ -329,7 +329,6 @@
                   ]
                 aio_prepare.prepare()
                 deploy.deploy_sh(environment_vars: environment_vars)
-                tempest.tempest_install()
                 kibana.kibana(kibana_branch)
                 holland.holland()
                 if (env.STAGES.contains("Minor Upgrade")) {{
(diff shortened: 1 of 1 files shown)


rpc_jobs/rpc_aio.yml
```diff
@@ -339,6 +338,8 @@
                   kibana.kibana(env.KIBANA_SELENIUM_BRANCH)
                   holland.holland()
                 }} else if (env.STAGES.contains("Leapfrog Upgrade")) {{
+                  // Required for leapfrog resource generation
+                  tempest.tempest_install()
                   deploy.upgrade_leapfrog(environment_vars: environment_vars)
```
git-harry: What about moving it into `deploy.upgrade_leapfrog`, that should ensure it can't get removed by mistake and simplifies the template.

mattt416:
@git-harry think your suggesting of moving it makes sense.  Testing this change [here](https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_kilo-trusty-leapfrogupgrade-swift-periodic/15/console) to make sure tempest can run from deploy.upgrade_leapfrog() without issue.

evrardjp:
I'd agree with the place change, but I'd do it in two different commits, for faster iteration!

mattt416:
That build failed, not even sure we can test this stuff now because of recent changes.  :-/

mattt416:
The test job I spun up is broken completely (and actually, thought this was previously removed/disabled), so created a new test [here](https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_newton-trusty-leapfrogupgrade-swift-periodic/27/console).

evrardjp:
Are we ready to merge?

PR #401, opened 2017-08-24
RE-220 Use fork of openstack-ansible-ops

mattt416:
The current SHA is broken (cobbler26 no longer available), and upstream
has completely rewritten the mnaio build tool which will require a
bunch of RE effort to integrate into rpc-gating.  As a stop gate, we're
going to use a fork of openstack-ansible-ops so we can continue to use
the mnaio jobs until we're able to bump to the head of
openstack-ansible-ops.

Issue: [RE-220](https://rpc-openstack.atlassian.net/browse/RE-220)

### DIFF
--- a/rpc_jobs/params.yml
+++ b/rpc_jobs/params.yml
@@ -66,10 +66,10 @@
     parameters:
       - string:
           name: OSA_OPS_REPO
-          default: https://github.com/openstack/openstack-ansible-ops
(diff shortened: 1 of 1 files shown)


mattt416:
I'm happy using your fork for now -- forking onto rcbops gives it a bit more permanence which I think we should maybe avoid?

PR #404, opened 2017-09-01
RE-225 Re-order & make RPC-O AIO test consistent

odyssey4me:
Re-order the tests to ensure that Tempest runs
before Kibana, then Holland.

The deploy tests all verify that everything is
working for the source branch, so no need to
re-test for upgrades. Also, the MNAIO test
implementation does the more thorough test
regime of before and after.

The tests are now all consistent across all
'actions'. They always run in the same order,
always run for all tests and always use the
same settings. This is simpler to read and
understand what's going on.

Issue: [RE-225](https://rpc-openstack.atlassian.net/browse/RE-225)

### DIFF
--- a/rpc_jobs/rpc_aio.yml
+++ b/rpc_jobs/rpc_aio.yml
@@ -284,16 +284,6 @@
       timeout(time: 8, unit: 'HOURS'){{
         common.shared_slave() {{
           try {{
-            // This sets kibana_branch based on STAGES
-            // Note that we do this before the block below since
-            // env.UPGRADE_FROM_ENV gets overwritten
-            if (env.STAGES.contains("Minor Upgrade")
-                || env.STAGES.contains("Major Upgrade")
-                || env.STAGES.contains("Leapfrog Upgrade")) {{
-              kibana_branch = env.UPGRADE_FROM_REF
-            }} else {{
-              kibana_branch = env.KIBANA_SELENIUM_BRANCH
-            }}
             // prepare vars for leapfrog upgrade test of changes proposed to kilo
             // this is a special case as we only usually test upgrades on
             // changes to the destination branch.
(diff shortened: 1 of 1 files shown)


rpc_jobs/rpc_aio.yml
```diff
@@ -329,19 +329,23 @@
                   ]
                 aio_prepare.prepare()
                 deploy.deploy_sh(environment_vars: environment_vars)
-                kibana.kibana(kibana_branch)
```
hughsaunders: Could you put `tempest.tempest()` here instead of inside each branch of the if?

rpc_jobs/rpc_aio.yml
```diff
@@ -329,19 +329,23 @@
...
                   holland.holland()
                 }} else if (env.STAGES.contains("Leapfrog Upgrade")) {{
                   deploy.upgrade_leapfrog(environment_vars: environment_vars)
+                  tempest.tempest()
                   kibana.kibana(env.KIBANA_SELENIUM_BRANCH)
+                }} else {{
+                  // in a non-upgrade situation, these tests are run
+                  tempest.tempest()
+                  kibana.kibana(kibana_branch)
+                  holland.holland()
```
hughsaunders: Tempest and holland are always run, so they can go after the if, then create a new if block to  only run holland if Upgrade is not in stages.

odyssey4me: Holland is run in the Major upgrade though.

evrardjp: That sounds more complex than it should be...

I like it the way it is. However a few tests are missing:
- Minor upgrade is missing kibana and holland
- Leapfrog is missing holland, but that's ok.

odyssey4me: Leapfrog only ever ran Holland before, not after.

PR #405, opened 2017-09-01
RO-2433 Remove support role/holland tests

odyssey4me:
The support role is being removed from being a part
of RPC-O as it is not an engineering deliverable,
not tested and cannot be reliably included in the
product.

As the support role is also the way that the holland
backup system is deployed, the holland tests are
removed along with it.

Issue: [RO-2433](https://rpc-openstack.atlassian.net/browse/RO-2433)


odyssey4me:
For the record, a card has been registered to make this new testing gap clear: https://rpc-openstack.atlassian.net/browse/RO-2440

major:
I'll be here for a while longer so I'll merge this for now.

PR #406, opened 2017-09-04
TURTLES-67 Add ops fabric analytics service gate

supermari0:
This adds a gate job for the ops-fabric-analytics repo. I copied most of this from the chatbot gate PR.

I am not sure that I have API access to Jenkins, so I have not tested
this yet.

The goal of this gate is to return failed status if any of the tox (py35
or pep8 tests) fail. JUnit output is nice to have, but I would rather
have a working gate more quickly if possible.

The one part I'm not sure of is if tox exists in the container at
the time the tests are run.

### DIFF
--- a/rpc_jobs/ops_fabric_analytics.yml
+++ b/rpc_jobs/ops_fabric_analytics.yml
@@ -1,0 +1,40 @@
+- job:
+    name: Ops-Fabric-Analytics-Gate
+    description: "lint and unit tests for ops-fabric-analytics"
+    disabled: false
+    display-name: "ops-fabric-analytics lint and unit tests"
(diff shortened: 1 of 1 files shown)


supermari0:
recheck_project_builds

supermari0:
For some reason the recheck doesn't seem to be working

supermari0:
from slack:

```
Matt Thompson [1:44 PM] 
@mari6459 if we were modifying an existing job we could do some testing to ensure it still works as expected, but since it's a brand new job we can just merge and see how it goes

Mario Villaplana 
[1:45 PM] 
ok, i'll merge and try it out, thanks
```

trying this out

PR #407, opened 2017-09-06
RE-359 change gate failure ticket type to 'Issue-gate-failure'

mancdaz:
The Aglie Analytics tool is unable to filter on JIRA ticket labels, so
we need to ensrue gate failure tickets are created with a bespoke issue
type (Issue-releng-platform-alert) which we *can* filter on

Issue: [RE-359](https://rpc-openstack.atlassian.net/browse/RE-359)


mancdaz:
OK @hughsaunders, as discussed I made the issue type name more generic, and added it to the maas alert ticket creation function too

PR #409, opened 2017-09-06
RE-279 Build/save image in public cloud

odyssey4me:
In order to allow the capability to save an image
from a prepped instance in public cloud, a job
added which does so.

The nova CLI is used as rpc-gating uses pyrax
and therefore has access to the nova CLI. The
openstack client would be more ideal to use, but
switching from pyrax to it is out of scope for
this task.

Issue: [RE-279](https://rpc-openstack.atlassian.net/browse/RE-279)

### DIFF
--- a/pipeline_steps/artifact_build.groovy
+++ b/pipeline_steps/artifact_build.groovy
@@ -108,4 +108,36 @@ def container(String image) {
   } // pubcloud slave
 }
 
+def cloudimage(Map args) {
+  inventory="inventory.${common.rand_int_str()}"
+  inventory_path="${WORKSPACE}/rpc-gating/playbooks/${inventory}"
+  String instance_name = common.gen_instance_name()
+  try {
+    pubcloud.getPubCloudSlave(
+      image: args.src_image,
+      instance_name: instance_name,
+      inventory: inventory,
+      inventory_path: inventory_path,
+      region: args.region
+    )
+    pubcloud.savePubCloudSlave(
+      image: args.dest_image,
+      instance_name: instance_name,
+      inventory: inventory,
+      inventory_path: inventory_path,
+      region: args.region
+    )
+  } catch (e) {
+    print(e)
+    throw e
+  } finally {
+    pubcloud.delPubCloudSlave(
+      instance_name: instance_name,
+      inventory: inventory,
+      inventory_path: inventory_path,
+      region: args.region
+    )
+  }
+}
+
 return this
--- a/rpc_jobs/cloud_image_build.yml
+++ b/rpc_jobs/cloud_image_build.yml
@@ -1,0 +1,59 @@
+- job:
+    name: Cloud-Image-Build
+    project-type: workflow
+    concurrent: false
+    properties:
+      - build-discarder:
+          num-to-keep: 30
+    parameters:
+      - instance_params
+      - rpc_gating_params
+    dsl: |
+      library "rpc-gating@${RPC_GATING_BRANCH}"
+
+      // These are defined statically instead of parameters
+      // to ensure that the job is always executed the same
+      // way.
+      env.FLAVOR = "performance1-1"
+      env.STAGES = "Allocate Resources, Save Slave, Cleanup"
+      image_list = [
+        [ src: "Ubuntu 16.04 LTS (Xenial Xerus) (PVHVM)",
+          dest: "Ubuntu 16.04 LTS prepared for RPC deployment" ],
+        [ src: "Ubuntu 14.04 LTS (Trusty Tahr) (PVHVM)",
+          dest: "Ubuntu 14.04 LTS prepared for RPC deployment" ]
+      ]
+      region_list = [
+        "DFW", "HKG", "IAD", "ORD", "SYD"
+      ]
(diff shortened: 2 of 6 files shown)


odyssey4me:
The image save is being tested with https://rpc.jenkins.cit.rackspace.net/job/odyssey4me-image-save-test/ - that seems to be running fine so I'm verifying that the images are usable now.

odyssey4me:
This has been tested and has generated images for Trusty and Xenial in DFW, IAD and ORD: https://rpc.jenkins.cit.rackspace.net/job/Cloud-Image-Build/

pipeline_steps/artifact_build.groovy
```diff
@@ -108,4 +108,30 @@ def container(String image) {
   } // pubcloud slave
 }
 
+def cloudimage(Map args) {
+  inventory="inventory.${common.rand_int_str()}"
+  inventory_path="${WORKSPACE}/rpc-gating/playbooks/${inventory}"
```
hughsaunders: This should be passed into getPubCloudSlave

pipeline_steps/artifact_build.groovy
```diff
@@ -108,4 +108,30 @@ def container(String image) {
...
+  String instance_name = common.gen_instance_name()
+  try {
+    pubcloud.getPubCloudSlave(
+      instance_name: instance_name,
+      image: args.src_image,
+      region: args.region
+    )
+    pubcloud.savePubCloudSlave(
```
hughsaunders: I like this high level function, It doesn't seem like much of an extension to do getpubcloudslave, aio deploy, savepubcloudslave. Though obviously rehydration would be required afterwards.

pipeline_steps/artifact_build.groovy
odyssey4me: Done.

pipeline_steps/artifact_build.groovy
odyssey4me: Yep, that's the intent - it'd probably require some sort of jenkins slave cleanup in the cleanup script though... but that's for a follow on.

pipeline_steps/artifact_build.groovy
hughsaunders: Sorry, I should have been more explicit, inventory_path is required as well.

odyssey4me: Ah, thanks for clarifying via IRC. Done.

scripts/rax_create_image.py
```diff
@@ -0,0 +1,100 @@
...
+
+    # Setup argument parsing
+    parser = argparse.ArgumentParser(
+        description='Basic pyrax CLI utilities',
+        epilog='Licensed "Apache 2.0"')
+
+    parser.add_argument(
+        '-c',
+        '--credentialsfile',
+        help='<Required> The path to the pyrax.cfg file',
```
evrardjp: nit: ``<Required>`` is confusing, if you have a default.

scripts/rax_create_image.py
```diff
@@ -0,0 +1,100 @@
...
+    parser.add_argument(
+        '-i',
+        '--imagename',
+        help='<Required> The name to use for the saved image',
+        required=True
+    )
+
+    # Parse arguments
+    args = parser.parse_args()
+
```
evrardjp: nit: I suppose here we assume arguments are always good and well introduced? That is because this cli will only be used in gating, where we own the arguments given into this cli?

rpc_jobs/cloud_image_build.yml
```diff
@@ -0,0 +1,59 @@
...
+    dsl: |
+      library "rpc-gating@${RPC_GATING_BRANCH}"
+
+      // These are defined statically instead of parameters
+      // to ensure that the job is always executed the same
+      // way.
+      env.FLAVOR = "performance1-1"
+      env.STAGES = "Allocate Resources, Save Slave, Cleanup"
+      image_list = [
+        [ src: "Ubuntu 16.04 LTS (Xenial Xerus) (PVHVM)",
```
evrardjp: This means the source image will always be able to get built, but it also means our target is shifting, because everytime we'll build based on that, we'll have a different image id. 

So if we expect reproducibility, this is kinda good but not great.

Because Rackspace doesn't hold the "first release" image of each distro, should we either ask rackspace to have a "first release" image of each distro (and use this image id as source?) OR directly save the original image ID somewhere(to be able to track how this image was built?).
2) Directly do

odyssey4me: We don't expect reproducibility. When this job is activated, we want a new image built at the time based on the current image available.

scripts/rax_create_image.py
odyssey4me: Yes. The hope is that we phase out the use of this script altogether and shift to using the openstack CLI very soon instead - but for now pyrax has to be used.

rpc_jobs/cloud_image_build.yml
evrardjp: Good, that's exactly what I hoped and understood. I like evidence!

PR #411, opened 2017-09-06
add pike job to rpc-maas

cloudnull:

### DIFF
--- a/rpc_jobs/irr_role_test.yml
+++ b/rpc_jobs/irr_role_test.yml
@@ -19,6 +19,7 @@
     context:
       - master:
           irr_flavor: "performance2-15"
+      - pike
       - ocata
       - newton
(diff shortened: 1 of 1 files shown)


rpc_jobs/irr_role_test.yml
```diff
@@ -40,6 +43,8 @@
       - image: trusty
         context: master
       - image: trusty
```
major: Shouldn't this and the one below it be `xenial`?

major: Nevermind, I missed the `exclude:` at the top there. ;)

mattt416:
We REALLY need https://github.com/rcbops/rpc-gating/pull/332 -- but I won't block this PR on that merging.

PR #413, opened 2017-09-12
LA-436 Fix issues relating to testing instances

git-harry:
The running of `generate-resources.yml` is now managed by rpc-openstack,
the code to perform that function here is removed.

To ensure that Telegraf is properly installed during the upgrade phase
of an upgrade, the appropriate credentials are supplied to the upgrade
script.

To test instance networking a number have vars are set because pinging
using the floating IP address does not work with the default
configuration when using vxlan.

Issue: [LA-436](https://rpc-openstack.atlassian.net/browse/LA-436)

### DIFF
--- a/playbooks/vars/all.yml
+++ b/playbooks/vars/all.yml
@@ -27,3 +27,7 @@ default_gating_overrides:
   # too low for the liberty->mitaka upgrade job where more space is used by
   # additional packages, venvs, logs, etc.
   percent_used_critical_threshold: 95
+  # These neutron vars are required to allow instances to be pinged after an
+  # upgrade when using vxlan
+  neutron_l2_population: False
+  neutron_vxlan_group: 224.0.0.1
--- a/playbooks/vars/all.yml
+++ b/playbooks/vars/all.yml
@@ -27,7 +27,3 @@ default_gating_overrides:
   # too low for the liberty->mitaka upgrade job where more space is used by
   # additional packages, venvs, logs, etc.
   percent_used_critical_threshold: 95
-  # These neutron vars are required to allow instances to be pinged after an
-  # upgrade when using vxlan
-  neutron_l2_population: False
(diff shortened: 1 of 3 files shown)


playbooks/vars/all.yml
```diff
@@ -27,3 +27,7 @@ default_gating_overrides:
   # too low for the liberty->mitaka upgrade job where more space is used by
   # additional packages, venvs, logs, etc.
   percent_used_critical_threshold: 95
+  # These neutron vars are required to allow instances to be pinged after an
+  # upgrade when using vxlan
+  neutron_l2_population: False
```
mattt416: The rpc-o default in newton for `neutron_l2_population` is `True`, are we concerned at all that we're not testing that default in any capacity?

Also, does this need to be set for kilo before the upgrade happens?  If so, I believe that config now all lives in repo at https://github.com/rcbops/rpc-openstack/tree/kilo/scripts/config.

git-harry: @mattt416 this is based on the testing @hughsaunders did to get existing floating IP addresses to continue to work after an upgrade. My understanding is no customer is using vxlan and so this configuration change is required specifically to test this part of the upgrade. An alternative option would be to have a script in osa-ops specifically for restarting the appropriate networking components after the upgrade when using vxlan with l2 pop.

Regarding where the configuration lives, I was not aware of this change until yesterday when I picked up RE-97, mnaio's do not currently use the new location. Depending on your first point, I will also set it in the patches to RPCO.

mattt416: @git-harry since we only do the persistent resource checks on an mnaio, how about moving this `neutron_l2_population` override to `playbooks/vars/onmetal.yml`?

git-harry:
@mattt416 how does that look? I can then transfer them over to RPCO as part of enabling the new config functionality for multi-node deployments.

PR #415, opened 2017-09-13
RLM-204 Move Kilo Leapfrog Jobs to Periodic Only

antonym:
Adds Kilo Leapfrog PRs to the exclude list for MultiAIO and AIO:

Issue: [RLM-204](https://rpc-openstack.atlassian.net/browse/RLM-204)

### DIFF
--- a/rpc_jobs/multi_node_aio.yml
+++ b/rpc_jobs/multi_node_aio.yml
@@ -101,7 +101,10 @@
       # Leapfrog upgrades are only run for kilo --> newton141
       # as the upgrade method is not supported for any
       # other target series.
-      # the kilo --> newton141 job is triggered by PRs to either branch.
+      # the kilo --> newton141 job is only ran as a periodic job until
+      # it's more reliable and can be tested as a PR
+      - action: leapfrogupgrade
+        trigger: pr
       - series: liberty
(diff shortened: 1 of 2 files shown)


rpc_jobs/multi_node_aio.yml
```diff
@@ -101,7 +101,11 @@
       # Leapfrog upgrades are only run for kilo --> newton141
       # as the upgrade method is not supported for any
       # other target series.
-      # the kilo --> newton141 job is triggered by PRs to either branch.
+      # the kilo --> newton141 job is only ran as a periodic job until
+      # it's more reliable and can be tested as a PR
+      - series: kilo
+        action: leapfrogupgrade
+        trigger: pr
```
mattt416: How about:

```
- action: leapfrogupgrade
   trigger: pr
```

antonym: K, updated with the two lines as it sounds like using three entries in an exclude doesn't work.

mattt416:
We'll need to delete jobs after this one merges.  Any issues w/ that @hughsaunders and @git-harry ?

hughsaunders:
project_builds failed as PR is from a fork.

PR #417, opened 2017-09-14
RE-221 Use latest openstack-ansible-ops

mattt416:
This commit makes a # of changes to support deploying from master of
openstack-ansible-ops.  In summary:

1. Remove overrides for OSA_OPS_REPO and OSA_OPS_BRANCH.
2. Remove MULTI_NODE_AIO_DIR param since there no longer two versions of
   the multi-node-aio tool to deploy.
3. Change any reference from deploy-vm to infra1-vm or anything requiring
   the deploy VM since upstream has removed the deploy VM and made
   deployment happen from infra1.
4. Rename COMPUTE_NODES and VOLUME_NODES to ADDITIONAL_COMPUTE_NODES
   and ADDITIONAL_VOLUME_NODES to better reflect the number of
   additional nodes to deploy, not the total.  This is done since we
   no longer use a simple hosts.json where we can override the default
   inventory.
5. In multi_node_aio_prepare.prepare(), we set all infra VMs to have a
   RAM allocation of 16 GB, since it defaults to 8 GB which is
   insufficient.  Here we also have to push the host's authorized_keys
   to infra1 so we can ssh in with our jenkins user.  This is no longer
   handled automatically upstream.
6. Further to the above, we create a dynamic inventory script to handle
   increasing compute/cinder node count.  In a nutshell, to add another
   compute, you'd have to modify the static inventory
   (playbooks/inventory/hosts) and you'd have to
   create a hosts_var file for the node.  With the dynamic inventory,
   we just provide environment variables to indicate how many additional
   compute/volume hosts we need and it will eliminate the need for the
   first two steps.
7. Lastly, we modify maas.get_mnaio_entity_names() to use the ansible
   inventory, since we no longer have access to a hosts.json file.

TODO:

1. Figure out why we're getting so many tempest failures (even with
   DEPLOY_TELEGRAF temporarily disabled (see [1]).

[1] https://rpc.jenkins.cit.rackspace.net/job/mattt-Multi-Node-AIO_newton-xenial-deploy-none-periodic/22/

Issue: [RE-221](https://rpc-openstack.atlassian.net/browse/RE-221)


mattt416:
Had a successful run [here](https://rpc.jenkins.cit.rackspace.net/job/mattt-Multi-Node-AIO_newton-xenial-deploy-none-periodic/30/) testing with `DEPLOY_TELEGRAF=no` and 
`tempest_test_sets='scenario'`.

rpc_jobs/params.yml
```diff
@@ -66,10 +66,10 @@
     parameters:
       - string:
           name: OSA_OPS_REPO
-          default: https://github.com/git-harry/openstack-ansible-ops
+          default: https://github.com/openstack/openstack-ansible-ops
       - string:
           name: OSA_OPS_BRANCH
-          default: 93ac83d3436fd078500d6fd85ec5078e357f4db4
+          default: master
```
odyssey4me: I think it'd be best if we use a SHA here to avoid upstream changes causing problems for us.

mattt416:
Rather than dropping the dynamic inventory, it may be cleaner to just carry an inventory template and then overwrite that based on how many compute/cinders have been specified.  I'll rework this today.

rpc_jobs/params.yml
mattt416: Fixed

mattt416:
@odyssey4me what I forgot to mention today in Slack (but is captured in the commit description) is that I'd also have to create separate host_vars files for the dynamically added nodes, which is handled nicely with the dynamic inventory.  I'll try to poke around in a bit, but do you know if this dynamic inventory as is will not work with ansible 2.4?  Surely it must be possible to create dynamic inventories that are supported by ansible <= 2.4?

mattt416:
That build failed pretty early on, unsurprisingly on something inventory-related.  I'll need to run through this end to end to test the latest ansible version which we weren't using previously.

mattt416:
Marking this card as WIP.  In light of newton issues, I went to build a mitaka rpco deploy and it looks like the latest openstack-ansible-ops changes don't permit booting trusty VMs.  The failures I'm seeing are related to grub-install and will need to be resolved before this merges.

mattt416:
We'll need to wait for https://review.openstack.org/#/c/507503/ to merge and bump OSA_OPS_BRANCH before this merges.

pipeline_steps/multi_node_aio_prepare.groovy
```diff
@@ -133,14 +142,16 @@ def prepare() {
...
+      scp -o StrictHostKeyChecking=no ${env.WORKSPACE}/user_zzz_gating_variables.yml infra1:/etc/openstack_deploy/user_zzz_gating_variables.yml
 
-      ssh -T -o StrictHostKeyChecking=no deploy1 << 'EOF'
+      ssh -T -o StrictHostKeyChecking=no infra1 << 'EOF'
       set -xe
       sudo cp /etc/openstack_deploy/user_variables.yml /etc/openstack_deploy/user_variables.yml.bak
       sudo cp -R /opt/rpc-openstack/openstack-ansible/etc/openstack_deploy /etc
       sudo cp /etc/openstack_deploy/user_variables.yml.bak /etc/openstack_deploy/user_variables.yml
+      # Write random var to user_variables.yml incase it's empty (which happens to be the case on stable/mitaka)
+      echo "osa_ops_mnaio: true" | sudo tee -a /etc/openstack_deploy/user_variables.yml
```
mattt416: We could upstream this, however I'm not sure it's worthwhile since stable/mitaka isn't really used upstream any more.  Thoughts?

rpc_jobs/params.yml
```diff
@@ -87,14 +87,6 @@
...
       - string:
-          name: MULTI_NODE_AIO_DIR
-          default: "multi-node-aio"
-          description: |
-            The directory to trigger build.sh from
-            Options:
-              multi-node-aio
-              multi-node-aio-xenial-ansible
-      - string:
           name: "COMPUTE_NODES"
```
git-harry: I'd be inclined towards changing the names of these parameters to be more descriptive/accurate, for example `ADDITIONAL_COMPUTE_NODES`. The parameter descriptions also seem misleading now.

pipeline_steps/multi_node_aio_prepare.groovy
```diff
@@ -91,10 +91,16 @@ def prepare() {
         git url: env.OSA_OPS_REPO, branch: "master"
         sh "git checkout ${env.OSA_OPS_BRANCH}"
       }
-      dir("openstack-ansible-ops/${env.MULTI_NODE_AIO_DIR}") {
+      dir("openstack-ansible-ops/multi-node-aio") {
+        // The multi-node-aio tool is quite modest when it comes to allocating
+        // RAM to VMs -- since we have RAM to spare we double that assigned to
+        // infra nodes.
+        sh """#!/bin/bash
+          sed -i 's/server_vm_ram: 8192/server_vm_ram: 16384/g' playbooks/host_vars/infra*.yml
```
git-harry: This comment conflicts with the commit message which describes the default allocation as insufficient, which is correct?

pipeline_steps/multi_node_aio_prepare.groovy
```diff
@@ -91,10 +91,16 @@ def prepare() {
...
+          cp -a ${WORKSPACE}/rpc-gating/scripts/dynamic_inventory.py playbooks/inventory
+          cp -a ${WORKSPACE}/rpc-gating/playbooks/templates/openstack_user_config.yml.j2 playbooks/osa/openstack_user_config.yml
+        """
         timeout(time: 45, unit: "MINUTES") {
-          add_nodes()
```
git-harry: This function is defined at the top of this file, it was only used here and now it's superseded it should probably be removed.

mattt416:
@git-harry I had a "successful" run here: https://rpc.jenkins.cit.rackspace.net/job/mattt-Multi-Node-AIO_newton-xenial-deploy-none-periodic/64/.  Not all tempest tests passed, but I'm not sure what the success rate on tempest tests on the MNAIO is.  The Trusty VM issues have been resolved, but sadly mitaka doesn't build because of the inline user_openstack_config.yml file we carry.  I'll need to resolve that today and also address all the feedback you've provided.

odyssey4me:
@mattt416 Perhaps the appropriate template openstack_user_config file could be held in rpc-openstack instead of the gating repo?

mattt416:
@odyssey4me yeah that's a good idea -- gets the file out of rpco and lets each branch carry its own file.

odyssey4me:
@mattt416 It may be worth a chat to @cloudnull to discuss the work you've done to see if it'd be worth upstreaming it. I'm thinking that the MNAIO toolset should be OSA version independent and should not actually be doing config prep for OSA at all (that should be handled in the OSA branch), so the OSA repo (and/or RPC repo) should be the place where the specifics of config for MNAIO tests should be done.

cloudnull:
I agree with @odyssey4me. Making the MNAIO OS agnostic would be great. If we could upstream changes to make this easier, even better.b

pipeline_steps/multi_node_aio_prepare.groovy
cloudnull: This seems fine however I think it'd better to get some type of an option into the host vars for us to use instead of relying on in place injections.

pipeline_steps/multi_node_aio_prepare.groovy
cloudnull: seems like a good idea to get that in when the deploy osa playbook is executed.

playbooks/templates/openstack_user_config.yml.j2
```diff
@@ -0,0 +1,199 @@
+---
```
cloudnull: the changes in this file seem like a good addition upstream too.

playbooks/templates/openstack_user_config.yml.j2
```diff
@@ -0,0 +1,199 @@
...
+cidr_networks:
+  container: 10.0.236.0/22
+  tunnel: 10.0.240.0/22
+  storage: 10.0.244.0/22
+  flat: 10.0.248.0/22
```
cloudnull: maybe we should make the base cidrs configurable upstream?

playbooks/vars/onmetal.yml
```diff
@@ -11,7 +11,7 @@ gating_overrides:
   tempest_testr_opts:
     - '--concurrency 3'
   tempest_run_tempest_opts: []
-  tempest_public_subnet_cidr: "172.29.248.0/22"
+  tempest_public_subnet_cidr: "10.0.248.0/22"
```
cloudnull: same as above. maybe it would help if we made the base cidrs configurable upstream?

pipeline_steps/maas.groovy
```diff
@@ -1,20 +1,15 @@
 String get_mnaio_entity_names() {
   String entities = sh (
     script: """#!/usr/bin/env python
-from __future__ import print_function
-from itertools import chain
+from ansible.parsing.dataloader import DataLoader
+from ansible.vars.manager import VariableManager
```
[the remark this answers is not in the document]
mattt416: Good catch, this is leftover from the pre-ansible 2.4 stuff that I did.

scripts/dynamic_inventory.py
```diff
@@ -0,0 +1,98 @@
...
+  'cinder_hosts': [],
+}
+hosts = []
+hostvars = {}
+
+compute_count = os.getenv('COMPUTE_NODES', 0)
+cinder_count = os.getenv('VOLUME_NODES', 0)
+
+for c in range(0, int(compute_count)):
+    # we want to start with compute3
```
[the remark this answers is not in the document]
mattt416: Fixing

scripts/dynamic_inventory.py
```diff
@@ -0,0 +1,98 @@
...
+for c in range(0, int(cinder_count)):
+    # we want to start with cinder3
+    host = "cinder%d" % (3 + c)
+    hosts.append(host)
+    inventory['all'].append(host)
+    inventory['pxe_servers'].append(host)
+    inventory["cinder_hosts"].append(host)
+
+for i, host in enumerate(hosts):
+    # we want to start with 10.0.236.144
```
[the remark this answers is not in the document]
mattt416: Yeah, old comment -- fixing!

pipeline_steps/multi_node_aio_prepare.groovy
mattt416: Deleting.

pipeline_steps/multi_node_aio_prepare.groovy
mattt416: I see your mnaio upstream change @cloudnull, will adapt once that merges.  Thanks!

pipeline_steps/multi_node_aio_prepare.groovy
mattt416: Thanks for upstreaming @cloudnull, will adapt once that merges!

rpc_jobs/params.yml
mattt416: Yeah, description was unclear, updating param names.

scripts/dynamic_inventory.py
```diff
@@ -0,0 +1,98 @@
...
+  'pxe_servers': [],
+  'cinder_hosts': [],
+}
+hosts = []
+hostvars = {}
+
+compute_count = os.getenv('COMPUTE_NODES', 0)
+cinder_count = os.getenv('VOLUME_NODES', 0)
+
+for c in range(0, int(compute_count)):
```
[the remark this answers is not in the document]
mattt416: Sensible, updating.

pipeline_steps/multi_node_aio_prepare.groovy
```diff
@@ -91,10 +91,16 @@ def prepare() {
...
       }
-      dir("openstack-ansible-ops/${env.MULTI_NODE_AIO_DIR}") {
+      dir("openstack-ansible-ops/multi-node-aio") {
+        // The multi-node-aio tool is quite modest when it comes to allocating
+        // RAM to VMs -- since we have RAM to spare we double that assigned to
+        // infra nodes.
+        sh """#!/bin/bash
+          sed -i 's/server_vm_ram: 8192/server_vm_ram: 16384/g' playbooks/host_vars/infra*.yml
+          cp -a ${WORKSPACE}/rpc-gating/scripts/dynamic_inventory.py playbooks/inventory
+          cp -a ${WORKSPACE}/rpc-gating/playbooks/templates/openstack_user_config.yml.j2 playbooks/osa/openstack_user_config.yml
```
cloudnull: I think with https://review.openstack.org/#/c/507886/ we could probably remove this

playbooks/templates/openstack_user_config.yml.j2
cloudnull: with pr https://review.openstack.org/#/c/507886/ i think this file could be dropped.

pipeline_steps/multi_node_aio_prepare.groovy
```diff
@@ -91,10 +91,16 @@ def prepare() {
...
         sh "git checkout ${env.OSA_OPS_BRANCH}"
       }
-      dir("openstack-ansible-ops/${env.MULTI_NODE_AIO_DIR}") {
+      dir("openstack-ansible-ops/multi-node-aio") {
+        // The multi-node-aio tool is quite modest when it comes to allocating
+        // RAM to VMs -- since we have RAM to spare we double that assigned to
+        // infra nodes.
+        sh """#!/bin/bash
+          sed -i 's/server_vm_ram: 8192/server_vm_ram: 16384/g' playbooks/host_vars/infra*.yml
+          cp -a ${WORKSPACE}/rpc-gating/scripts/dynamic_inventory.py playbooks/inventory
```
cloudnull: With PR https://review.openstack.org/#/c/507890/ we should be able to simple export our inventory file instead of overwriting the existing one.

pipeline_steps/multi_node_aio_prepare.groovy
mattt416: This is a great change, thanks @cloudnull.  Unfortunately we'll still be in the position where either rpc-gating or rpc-openstack needs to carry this because mitaka, liberty, kilo, etc. need to reference os-infra_hosts which isn't in this openstack_user_cnofig.yml file.  Do you have any suggestions to avoid us having to carry that?

playbooks/templates/openstack_user_config.yml.j2
cloudnull: the previous pr has merged.

pipeline_steps/multi_node_aio_prepare.groovy
```diff
@@ -107,10 +28,20 @@ def prepare() {
...
-      dir("openstack-ansible-ops/${env.MULTI_NODE_AIO_DIR}") {
+      dir("openstack-ansible-ops/multi-node-aio") {
+        sh """#!/bin/bash
+          # The multi-node-aio tool is quite modest when it comes to allocating
+          # RAM to VMs -- since we have RAM to spare we double that assigned to
+          # infra nodes.
+          # TODO(mattt): When https://review.openstack.org/#/c/507875/ merges
+          #              we can simply overwrite infra_vm_server_ram in group vars
+          sed -i 's/server_vm_ram: 8192/server_vm_ram: 16384/g' playbooks/host_vars/infra*.yml
+          cp -a ${WORKSPACE}/rpc-gating/scripts/dynamic_inventory.py playbooks/inventory
```
cloudnull: this like can now be a simple environmental export:

``export MNAIO_INVENTORY="${WORKSPACE}/rpc-gating/scripts/dynamic_inventory.py"``

pipeline_steps/multi_node_aio_prepare.groovy
```diff
@@ -107,10 +28,20 @@ def prepare() {
...
+          # TODO(mattt): Once https://review.openstack.org/#/c/507886/ we can
+          #              drop carrying this in rpc-gating
+          cp -a ${WORKSPACE}/rpc-gating/playbooks/templates/openstack_user_config.yml.j2 playbooks/osa/openstack_user_config.yml
```
cloudnull: this pr has merged.

playbooks/templates/openstack_user_config.yml.j2
cloudnull: this pr has merged

pipeline_steps/multi_node_aio_prepare.groovy
```diff
@@ -107,10 +28,20 @@ def prepare() {
...
       }
-      dir("openstack-ansible-ops/${env.MULTI_NODE_AIO_DIR}") {
+      dir("openstack-ansible-ops/multi-node-aio") {
+        sh """#!/bin/bash
+          # The multi-node-aio tool is quite modest when it comes to allocating
+          # RAM to VMs -- since we have RAM to spare we double that assigned to
+          # infra nodes.
+          # TODO(mattt): When https://review.openstack.org/#/c/507875/ merges
+          #              we can simply overwrite infra_vm_server_ram in group vars
+          sed -i 's/server_vm_ram: 8192/server_vm_ram: 16384/g' playbooks/host_vars/infra*.yml
```
cloudnull: the PR to set this via a variable has merged. we can now drop this injection and just use an override.

pipeline_steps/multi_node_aio_prepare.groovy
mattt416: Cool!  The dynamic_inventory.py is just to _extend_ the static inventory if a jenkins job needs additional compute or volume hosts, so I think we'll still need to put it into playbooks/inventory right?

mattt416:
Some run output:

https://rpc.jenkins.cit.rackspace.net/job/mattt-Multi-Node-AIO_newton-xenial-deploy-none-periodic/75/console (ran fine, but failed some tempest tests)

mattt416:
Hold onto em

PR #419, opened 2017-09-14
RLM-204 Fix build failure issue creation

hughsaunders:
Currently the mechanism that creates a jira issue when a periodic
build fails is not working. The problem is a call:
  common.create_jira_issue(project="RO")

This looks like a calll that passes a keyword argument, its not. Its
an assignment with a function call. In CPS groovy assignments return
null, so this is effectively common.create_jira_issue(null).

The fix is to remove project=.

Issue: [RLM-204](https://rpc-openstack.atlassian.net/browse/RLM-204)

### DIFF
--- a/rpc_jobs/rpc_aio.yml
+++ b/rpc_jobs/rpc_aio.yml
@@ -330,9 +330,9 @@
                 currentBuild.result = 'FAILURE'
                 if (env.TRIGGER == 'periodic'){{
                   print("Creating jira issue in project RO for failed periodic build")
-                  common.create_jira_issue(project="RO")
+                  common.create_jira_issue("RO")
                 }} else {{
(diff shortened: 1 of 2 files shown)


hughsaunders:
recheck_lint

pipeline_steps/common.groovy
```diff
@@ -458,7 +458,8 @@ void clone_with_pr_refs(
   String repo="git@github.com:${env.ghprbGhRepository}",
   String ref="origin/pr/${env.ghprbPullId}/merge",
   String refspec='+refs/pull/\\*:refs/remotes/origin/pr/\\*'\
-                +' +refs/heads/\\*:refs/remotes/origin/\\*'
+                +' +refs/heads/\\*:refs/remotes/origin/\\*',
```
mattt416: Is this here needed?

hughsaunders:
recheck_lint

PR #420, opened 2017-09-14
RQE-237 Add Ironic gate jobs

mkam:
Adds periodic and PR jobs to run ironic
deployments. PR jobs only run if trigger
phrase is used to reduce resource
consumption.

Issue: [RQE-237](https://rpc-openstack.atlassian.net/browse/RQE-237)

### DIFF
--- a/rpc_jobs/rpc_aio.yml
+++ b/rpc_jobs/rpc_aio.yml
@@ -183,6 +183,39 @@
     jobs:
       - 'RPC-AIO_{series}-{image}-{action}-{scenario}-{ztrigger}'
 
+- project:
+    name: 'RPC-AIO-Ironic-Jobs'
+    series:
+      - newton:
+          branch: newton
+          branches: "newton"
+      - master:
+          branch: master
+          branches: "master"
+    image:
+      - xenial:
+          IMAGE: "Ubuntu 16.04.2 LTS prepared for RPC deployment"
+    action:
+      - deploy:
+          ACTION_STAGES: >-
+            Install Tempest,
+            Tempest Tests,
+            Prepare Kibana Selenium,
+            Kibana Tests
+    scenario:
+      - ironic-multihypervisor:
+          DEPLOY_IRONIC: "yes"
+          TRIGGER_PR_PHRASE_ONLY: true
+    ztrigger:
+      - pr:
+          CRON: ""
+          TRIGGER_USER_VARS: "maas_use_api: false"
+      - periodic:
+          branches: "do_not_build_on_pr"
+          NUM_TO_KEEP: 10
+    jobs:
+      - 'RPC-AIO_{series}-{image}-{action}-{scenario}-{ztrigger}'
+
 - job-template:
     # DEFAULTS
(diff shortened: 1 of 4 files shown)


rpc_jobs/rpc_ironic.yml
```diff
@@ -0,0 +1,137 @@
...
+      - string:
+          name: "DEPLOY_ELK"
+          default: "{DEPLOY_ELK}"
+          description: "Deploy ELK? yes/no"
+      - kibana_selenium_params:
+          KIBANA_SELENIUM_BRANCH: "{branch}"
+      - rpc_gating_params
+      - instance_params:
+          IMAGE: "{IMAGE}"
+          FLAVOR: "{FLAVOR}"
```
hughsaunders: Does this job need the default performance2-15 instance?

rpc_jobs/rpc_aio.yml
```diff
@@ -79,6 +79,15 @@
                     default_volume_type: ceph
             cinder_service_backup_driver: cinder.backup.drivers.ceph
             tempest_service_available_swift: false
+      - ironic-multihypervisor:
```
hughsaunders: How do you test multi-hypervisor in an AIO job? I think we only one hyp not in a container. Is it that multiple virtual ironic nodes are created for testing?

mkam:
@hughsaunders When I originally made the separate job, deploying ironic had different steps than the standard AIO job. Those differences eventually got folded into deploy.sh, so I updated the rpc_ironic.yml but then thought to reuse the existing rpc_aio.yml. At this point, I've got a successful run with rpc_ironic.yml and not rpc_aio.yml, which is why I haven't deleted the original one yet.

rpc_jobs/rpc_aio.yml
mkam: Multi-hypervisor refers to being able to create baremetal nodes and virtual nodes. The other scenario for ironic will be "standalone," which won't deploy nova compute so that only ironic builds are supported.

mkam:
Got the rpc_aio.yml to work, so I've deleted the old separate rpc_ironic.yml. I had to open https://github.com/rcbops/rpc-openstack/pull/2552 to fix some issues where the ironic compute container was installing the wrong packages because of the hardcoded nova_virt_type value.

rpc_jobs/rpc_aio.yml
```diff
@@ -180,6 +183,20 @@
       # Ocata onwards.
       - series: master
         image: trusty
+      # Ironic builds are not supported for series before newton
```
hughsaunders: Something we've been exploring recently is adding new JJB projects that used the same template as previous projects, this can help reduce the complexity of projects and the number of excludes that are required. 

See https://github.com/rcbops/rpc-gating/blob/master/rpc_jobs/irr_role_test.yml#L56-L69 for an example where ceph jobs were added as a separate project rather than new axis/item to the existing project.

mkam: Done!

PR #416, opened 2017-09-13
RLM-58 Adds periodic AIO leapfrog jobs for deployed versions of kilo

antonym:
Creates a seperate rpc_aio_leapfrog job to avoid stepping on
existing jobs in place.  Tests desired kilo branches for leapfrogs
as periodic jobs.

Issue: [RLM-58](https://rpc-openstack.atlassian.net/browse/RLM-58)

PR #422, opened 2017-09-14
RLM-58: Adds periodic AIO leapfrog jobs for deployed versions of kilo

antonym:
Creates a seperate rpc_aio_leapfrog job to avoid stepping on
existing jobs in place.  Tests desired kilo branches for leapfrogs
as periodic jobs.

Reopening as a branch on the main repo, closed https://github.com/rcbops/rpc-gating/pull/416

Issue: [RLM-58](https://rpc-openstack.atlassian.net/browse/RLM-58)

### DIFF
--- a/rpc_jobs/rpc_aio_leapfrog.yml
+++ b/rpc_jobs/rpc_aio_leapfrog.yml
@@ -8,61 +8,73 @@
           branch: "r14.2.0"
           branches: "r14.2.0"
           KIBANA_SELENIUM_BRANCH: "newton"
+          DEPLOY_TELEGRAF: "yes"
       - newton-r11.1.15:
           UPGRADE_FROM_REF: "r11.1.15"
           branch: "r14.2.0"
           branches: "r14.2.0"
           KIBANA_SELENIUM_BRANCH: "newton"
+          DEPLOY_TELEGRAF: "yes"
       - newton-r11.1.14:
           UPGRADE_FROM_REF: "r11.1.14"
           branch: "r14.2.0"
           branches: "r14.2.0"
           KIBANA_SELENIUM_BRANCH: "newton"
+          DEPLOY_TELEGRAF: "yes"
       - newton-r11.1.10:
           UPGRADE_FROM_REF: "r11.1.10"
           branch: "r14.2.0"
           branches: "r14.2.0"
           KIBANA_SELENIUM_BRANCH: "newton"
+          DEPLOY_TELEGRAF: "yes"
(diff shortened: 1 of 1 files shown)


claco:
@antonym Question: What version of Newton does this target?  r14.2.0, which is what we're targeting currently, or the newton branch so we catch things earlier before they hist a newton release?

antonym:
It looks like it's just targeting the newton branch from looking at this code if that's what's setting things up:

https://github.com/rcbops/rpc-gating/blob/master/pipeline_steps/deploy.groovy#L88

I wonder if we'd need to add something like:

  if (env.SERIES == "kilo-*" && env.TRIGGER == "periodic"){
    branch="r14.2.0"
  }

mattt416:
@antonym Maybe we need something like this:

```
    series:
      - newton-r11.1.18:
          UPGRADE_FROM_REF: r11.1.18
          branch: r14.2.0
          branches: "r11.1.18"
          KIBANA_SELENIUM_BRANCH: newton
      - newton-r11.1.15:
          UPGRADE_FROM_REF: r11.1.15
          branch: r14.2.0
          branches: "r11.1.15"
          KIBANA_SELENIUM_BRANCH: newton
      - newton-r11.1.14:
          UPGRADE_FROM_REF: r11.1.14
          branch: r14.2.0
          branches: "r11.1.14"
          KIBANA_SELENIUM_BRANCH: newton
```

antonym:
Pushed changes up based on your recommendations, will squash branches if things look alright.

rpc_jobs/rpc_aio_leapfrog.yml
```diff
@@ -0,0 +1,93 @@
+- project:
+    name: 'RPC-AIO-Leapfrog-Testing'
+    # Note: branch is the branch for periodics to build
+    #       branches is the branch pattern to match for PR Jobs.
+    series:
+      - newton-r11.1.18:
+          UPGRADE_FROM_REF: "r11.1.18"
+          branch: "r14.2.0"
+          branches: "r14.2.0"
+          KIBANA_SELENIUM_BRANCH: "newton"
```
mattt416: Think we're about there.  Last question (I hope) that I have: do we need `DEPLOY_TELEGRAF: "yes"` like we have [here](https://github.com/rcbops/rpc-gating/blob/master/rpc_jobs/rpc_aio.yml#L21)?

antonym: Yeah, went ahead and added them for consistency.

PR #423, opened 2017-09-14
RLM-209 Set newton excludes for leapfrogupgrades

antonym:
Looks like we never added excludes for newton on leapfrogupgrades, so this removes those jobs.

Issue: [RLM-209](https://rpc-openstack.atlassian.net/browse/RLM-209)

### DIFF
--- a/rpc_jobs/multi_node_aio.yml
+++ b/rpc_jobs/multi_node_aio.yml
@@ -109,6 +109,8 @@
         action: leapfrogupgrade
       - series: mitaka
         action: leapfrogupgrade
+      - series: newton
+        action: leapfrogupgrade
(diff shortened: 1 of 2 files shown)


antonym:
Ah, ok, I think it's making a bit more sense to me...  we still need the Newton Jobs for commits to the Newton branch which then does an UPGRADE_FROM_REF: "kilo" branch to validate that Newton still works.  

The leapfrog jobs will handle mainly testing from the various kilo branches to the latest Newton so we probably could pull the kilo leapfrog job out of rpc_aio instead since it will be getting tested in the new rpc_aio_leapfrog.yml anyways.

I'll close this guy out.

PR #425, opened 2017-09-15
LA-436 Re-add Python path in influx-report.sh

git-harry:
The use of python in the virtual env `.venv` was incorrectly removed.
This change adds it back so the script `influx.py` is run with the
required dependencies.

Issue: [LA-436](https://rpc-openstack.atlassian.net/browse/LA-436)


mancdaz:
LGTM - commit where the PYTHON_PATH was errrantly removed from the invocation line: 6a367444de7e6340a1191899cf573f499a829df7

PR #427, opened 2017-09-15
RLM-58 Multi-node AIO gate jobs for Kilo Leapfrog testing

antonym:
Adds multi-node AIO gate jobs for Kilo Leapfrog Testing.  Uses small resources only to limit the number of jobs, can be adjusted as we determine our testing needs.

Issue: [RLM-58](https://rpc-openstack.atlassian.net/browse/RLM-58)

### DIFF
--- a/rpc_jobs/multi_node_aio_leapfrog.yml
+++ b/rpc_jobs/multi_node_aio_leapfrog.yml
@@ -7,7 +7,10 @@
           branches: "r14.2.0"
           KIBANA_SELENIUM_BRANCH: "newton"
           SERIES_USER_VARS: |
-            tempest_test_sets: 'scenario defcore api heat_api smoke'
+            tempest_test_sets: 'all'
+            lxc_container_vg_name: "USE_DIR_NOT_VG"
+            glance_default_store: "swift"
+            neutron_legacy_ha_tool_enabled: true
           DEPLOY_TELEGRAF: "yes"
       - newton-r11.1.15:
           UPGRADE_FROM_REF: "r11.1.15"
@@ -15,7 +18,10 @@
           branches: "r14.2.0"
           KIBANA_SELENIUM_BRANCH: "newton"
           SERIES_USER_VARS: |
-            tempest_test_sets: 'scenario defcore api heat_api smoke'
+            tempest_test_sets: 'all'
+            lxc_container_vg_name: "USE_DIR_NOT_VG"
+            glance_default_store: "swift"
+            neutron_legacy_ha_tool_enabled: true
           DEPLOY_TELEGRAF: "yes"
       - newton-r11.1.14:
           UPGRADE_FROM_REF: "r11.1.14"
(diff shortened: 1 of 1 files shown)


mattt416:
LGTM, though what if we ran these once a week rather than daily?

antonym:
I think we're fine with doing once a week to start things out and get things green but I think ultimately we'd like to get to daily.

Is there any particular time we should set in cron to align us to the release schedule so we'd have enough time to scramble to fix any issues that might crop up and so we can optimize the available resources?  I mainly just copied from multi-node-aio for now.

claco:
At this point, weekly would be fine, or how about every other day? We're going to have more drop in with Liberty. And we can run them on demand as well. Ultimately, if this causes resource contention, I'd like to surface that sooner rather than later. We're only going to have more and more of these for the foreseeable future.

claco:
Scheduling aside, I've be happy to get them in even if it was once a month so we can start button pushing to get everything to a green stable state.

mattt416:
Why don't we start off w/ every two days and see how we get on with that.  Perhaps something like this will work?

```
CRON: "H H * * H/2"
```

I think @mkam originally had these jobs run at a random hour/minute between midnight and 8 AM CST to give us as much capacity during the day for per-commit and manual testing, but perhaps as we add more and more jobs we need to spread this usage throughout the day (and into the weekend also)?

antonym:
Yeah, that works.  I pushed the updated cron up and I'm sure we'll be tweaking things as we go.

rpc_jobs/multi_node_aio_leapfrog.yml
```diff
@@ -0,0 +1,108 @@
+- project:
+    name: 'Multi-Node-AIO-Leapfrog-Testing'
+    series:
+      - kilo-r11.1.18:
+          branch: r11.1.18
+          branches: "r11.1.18"
+          KIBANA_SELENIUM_BRANCH: "liberty-12.2"
+          SERIES_USER_VARS: |
+            tempest_test_sets: 'scenario defcore api heat_api smoke'
+          DEPLOY_TELEGRAF: "yes"
```
mattt416: Think I jumped the gun with my LGTM, because I suspect we're hitting the same question here as we have in https://github.com/rcbops/rpc-gating/pull/422.  Notably, `branch` should be what we're upgrading to and `UPGRADE_FROM_REF` should be what we're upgrading from.

Also, we already have leapfrog jobs in `multi_node_aio.yml`.  Should all these MNAIO jobs be consolidated into this one file?

Lastly, so it's crystal clear, we're doing a leapfrog from kilo->liberty here?  I'm a bit confused by the liberty KIBANA_SELENIUM_BRANCH.

mattt416: Forget about consolidating leapfrog jobs, that's something we can do later.

antonym: Pushed new rev up based on the recommendations and fixes on the AIO.

rpc_jobs/multi_node_aio_leapfrog.yml
```diff
@@ -0,0 +1,120 @@
+- project:
+    name: 'Multi-Node-AIO-Leapfrog-Testing'
+    series:
+      - newton-r11.1.18:
+          UPGRADE_FROM_REF: "r11.1.18"
+          branch: "r14.2.0"
+          branches: "r14.2.0"
+          KIBANA_SELENIUM_BRANCH: "newton"
+          SERIES_USER_VARS: |
+            tempest_test_sets: 'scenario defcore api heat_api smoke'
```
mattt416: The `SERIES_USER_VARS` in newton [here](https://github.com/rcbops/rpc-gating/blob/master/rpc_jobs/multi_node_aio.yml#L27-L31) are significantly different.  Is that by design?

antonym: The remaining vars were copied based on it being for kilo, but if it's newton we're building for, I probably should grab those vars for newton and update it for that as well.

PR #429, opened 2017-09-19
RE-458 Build RAX instances with build_config=core

major:
This patch allows us to build RAX instances without automatic
updates applied by the Rackspace agent. Automatic updates are causing
problems with artifacting and general gate jobs.

The documentation for this option is here:

  https://support.rackspace.com/how-to/cloud-server-configuration-options/

Issue: [RE-458](https://rpc-openstack.atlassian.net/browse/RE-458)


major:
This also prevents the monitoring agent from being installed or configured. I see that as a good thing, but I figured it was worth bringing up in case it's a problem.

PR #432, opened 2017-09-20
[WIP] RLM-211 Add hooks for Lifecycle to slip fixes in

antonym:
In some cases, older branches have experienced bit rot or contain
bugs that were fixed in newer branches.  This enables us to jump
in and make changes during gate runs without having to check things
into each branch.

Also open to any ideas if this is the wrong approach.

Issue: [RLM-211](https://rpc-openstack.atlassian.net/browse/RLM-211)

### DIFF
--- a/pipeline_steps/lifecycle.groovy
+++ b/pipeline_steps/lifecycle.groovy
@@ -1,0 +1,38 @@
+def lifecycle_integration(){
+  common.conditionalStage(
+    stage_name: "Prepare Lifecycle",
+    stage: {
+      if (env.UPGRADE_FROM_REF.contains("kilo") || env.UPGRADE_FROM_REF.contains("r11.")) {
+        kilo_prepare()
+      }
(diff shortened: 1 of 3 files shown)


odyssey4me:
We're desperately trying to avoid increasing the complexity of the gate jobs and recommending that fixes for products go into the product's repo. There are several fixes in rpc-openstack's kilo branch, including patches to openstack-ansible. I would recommend using those mechanisms instead of doing all that here.

antonym:
Yeah, I agree with less complexity, this wasn't going to work anyways as I'd need to slip in the checkout procedure and before deploy runs which is in the same stage to actually get this fix in.  

Ultimately what I think we need if we don't have it, is a way to layer cherry picks on top of a branch that can assist the branch we're testing to complete so that we can get around bit rot issues but at the same time use the commits in the primary branch.  Then we can avoid rewriting tag history which probably isn't the way to go.  

Closing this out.

PR #433, opened 2017-09-21
RE-458 Prevent unattended upgrades executing

odyssey4me:
The default public cloud image includes the unattended
upgrades package which automatically upgrades all
installed packages when booting the instance.

RPC-O instance builds must not automatically upgrade
packages on boot as the test may require the use of
the apt artifacts repo which has differing, often older,
packages than those available upstream.

This patch uses cloud-init to prevent unattended
upgrades executing on boot.

Issue: [RE-473](https://rpc-openstack.atlassian.net/browse/RE-473)

Issue: [RE-458](https://rpc-openstack.atlassian.net/browse/RE-458)

### DIFF
--- a/playbooks/allocate_pubcloud.yml
+++ b/playbooks/allocate_pubcloud.yml
@@ -4,7 +4,8 @@
   gather_facts: False
   vars:
     count: 1
-    inventory_path: '{{lookup("env", "WORKSPACE")}}/rpc-gating/playbooks/inventory'
+    inventory_path: "{{ lookup('env', 'WORKSPACE') }}/rpc-gating/playbooks/inventory"
+    user_data_path: "{{ lookup('env', 'WORKSPACE') }}/rpc-gating/scripts/user_data_pubcloud.sh"
   tasks:
     - name: Provision a set of public cloud instances
       local_action:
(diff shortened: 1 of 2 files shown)


mancdaz:
👍   Pending test job: https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_master-xenial-deploy-swift-pr/462/console

playbooks/allocate_pubcloud.yml
```diff
@@ -15,6 +16,12 @@
           count: "{{ count }}"
           key_name: "jenkins"
           region: "{{ region }}"
+          # TODO(odyssey4me):
+          # We no longer need to use a lookup here once
+          # we shift to using Ansible 2.3+. In Ansible
+          # 2.2 only a string works here, not a file
+          # path.
```
hughsaunders: This comment should make it clear that its a problem with the rax module, not ansible lookups.

Eg: In Ansible<2.3 there is a bug which prevents the rax module from reading user_data files. To work around that bug, we supply the contents of the user_data file rather than the path to it.  This bug is fixed in 2.3, so once we upgrade to 2.3+, we can set `user_data: user_data_path`

odyssey4me: Done

odyssey4me:
All tests passed, including the RPC-O test.

major:
Nice!

PR #437, opened 2017-09-26
RE-207 Standardise elements of archiving artefacts

git-harry:
Artefacts, such as logs and configuration files, need to be stored for
finished builds. This commit changes some aspects of how this is done to
provide a cleaner separation between the responsibilities of RE and its
customers.

With this change `upload_to_cloud_files.yml` is now responsible for
taking the data to be archived, creating an appropriately named
compressed tarball, and uploading it to Cloud Files. This reduces the
responsibility of `archive_artifacts.yml`, the RPCO-specific artefact
script, to collecting all the required data into one directory.

`html_report_dest` is removed given there is no need to be able to
override it, the report file is created temporarily and is not something
that needs to be configured for a job.

The result of these changes is that any job must store all of its
artefacts in `$WORKSPACE/artifacts`.

Issue: [RE-207](https://rpc-openstack.atlassian.net/browse/RE-207)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -356,8 +356,6 @@ def archive_artifacts(String build_type = "AIO"){
       print "Uploading artifacts to Cloud Files..."
       pubcloud.uploadToCloudFiles(
         container: "jenkins_logs",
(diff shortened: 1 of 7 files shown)


git-harry:
recheck_all

PR #438, opened 2017-09-26
RE-197 Dependency Update Mechanism

hughsaunders:
#386  Need to go in first, but lets get the review process started on this one, but make sure you're only looking at the dep update commit, not the commits that are also in #386 (RE-196)

Issue: [RE-197](https://rpc-openstack.atlassian.net/browse/RE-197)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -601,13 +601,14 @@ def create_jira_issue(String project="RE",
     sh """#!/bin/bash -xe
       cd ${env.WORKSPACE}
       set +x; . .venv/bin/activate; set -x
-      python rpc-gating/scripts/jirautils.py create_issue\
-        --tag '$tag'\
-        --link '$link'\
-        --project '$project'\
+      python rpc-gating/scripts/jirautils.py \
         --user '$JIRA_USER' \
         --password '$JIRA_PASS' \
-        --type '$type'
+        create_issue \
+          --summary "JBF: ${tag}" \
+          --description "Jenkins Build Failed :( [${tag}|${link}]" \
+          --project '$project' \
+          --type '$type'
     """
   }
 }
--- a/rpc_jobs/dep_update.yml
+++ b/rpc_jobs/dep_update.yml
@@ -1,0 +1,51 @@
+- project:
+    name: "Dependency-Update-Jobs"
+    repo:
+      - rpc-openstack:
+          URL: "https://github.com/rcbops/rpc-openstack"
+          BRANCH: "master"
+          #CRON: "@weekly" <-- commented as inherits default
+          #JIRA_ISSUE_TYPE: "task"
+          JIRA_PROJECT_KEY: "RO"
+    jobs:
+      - 'Dep-Update_{repo}-{BRANCH}'
+
+
+- job-template:
+    CRON: "@weekly"
+    RPC_GATING_BRANCH: master
+    JIRA_ISSUE_TYPE: Task
+    name: 'Dep-Update_{repo}-{BRANCH}'
+    project-type: pipeline
+    pipeline-scm:
+      scm:
+        - git:
+            url: https://github.com/rcbops/rpc-gating
+            branches:
+              - "${{RPC_GATING_BRANCH}}"
+            credentials-id: "github_account_rpc_jenkins_svc"
+      script-path: job_dsl/dep_update.groovy
+    concurrent: false
+    properties:
+      - build-discarder:
+          num-to-keep: 30
+    triggers:
+        - timed: "{CRON}"
+    parameters:
+      - rpc_gating_params
+      - string:
+          name: "URL"
+          description: URL of the repo to be updated
+          default: "{URL}"
+      - string:
+          name: "BRANCH"
+          description: Branch of the repo to be updated
+          default: "{BRANCH}"
+      - string:
+          name: "JIRA_PROJECT_KEY"
+          description: Project key to use when creating JIRA issues
+          default: "{JIRA_PROJECT_KEY}"
+      - string:
+          name: "JIRA_ISSUE_TYPE"
+          description: "Type of issue to create"
+          default: "{JIRA_ISSUE_TYPE}"
--- a/scripts/dep_update.sh
+++ b/scripts/dep_update.sh
@@ -1,0 +1,77 @@
+#!/bin/bash -xe
+# Run pre and run hooks
+# Detect whether the repo has been modified
+# Create pr if required
+# Run post hook
+run_hook(){
+  if [[ -x "$1" ]]
+  then
+    $1 || {
+      echo "Hook script failed: $1"
+      exit 1
+    }
+  else
+    if [[ "$1" =~ .*/run ]]
+    then
+      echo "Run hook $1 not found."
+      exit 1
+    fi
+  fi
+}
+cd ${WORKSPACE}/repo
+run_hook "gating/update_dependencies/pre"
+run_hook "gating/update_dependencies/run"
+if [[ -z "$(git status -s)" ]]
+then
+  echo "Repo is clean, prep script made no changes to be committed."
+else
+  echo "Repo is dirty, preparing to propose changes."
+  # split URL into array separated by '/'s
+  IFS='/ ' read -r -a url_split <<< "$URL"
+  # owner/org is whatever is between the second and third / in the url
+  owner="${url_split[3]}"
+  # repo is anything after owner/
+  repo=${URL#https://github.com/${owner}/}
+  ssh_url="git@github.com:${owner}/${repo}"
+  message="This change is the result of running gating/update_dependencies/*
+  See update_dependencies in https://rpc-openstack.atlassian.net/wiki/spaces/RE/pages/19005457/RE+for+Projects for more details."
+
+  # Get / Create Jira issue
+  echo "Looking for Jira issue, will create if not found."
+  jira_summary="Update ${repo}:${BRANCH} dependencies"
+  issue=$(python ${WORKSPACE}/rpc-gating/scripts/jirautils.py \
+        --user "${JIRA_USER}" \
+        --password "${JIRA_PASS}" \
+        create_issue \
+          --project "${JIRA_PROJECT_KEY}" \
+          --summary "${jira_summary}" \
+          --description "${message}" \
+          --type "${JIRA_ISSUE_TYPE}" \
+          --label RE_DEP_UPDATE \
+          --existing-issue-query "project = ${JIRA_PROJECT_KEY} AND status = BACKLOG AND summary ~ \"${jira_summary}\""
+  )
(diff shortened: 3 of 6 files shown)


hughsaunders:
Test Job: https://rpc.jenkins.cit.rackspace.net/job/Dep-Update_rpc-openstack-newton/15/console ✅ 
PR Created by test job: https://github.com/rcbops/rpc-openstack/pull/2555 (I closed the pr to reduce clutter)

scripts/dep_update.sh
```diff
@@ -0,0 +1,47 @@
...
+  # owner/org is whatever is between the second and third / in the url
+  owner="${url_split[3]}"
+  # repo is anything after owner/
+  repo=${URL#https://github.com/${owner}/}
+  ssh_url="git@github.com:${owner}/${repo}"
+  pr_branch="${BRANCH}_dep_update_$(date +%H%M_%d%m%Y)"
+  echo "Repo is dirty, proposing changes"
+  git commit -a -m "${pr_branch}"
+  git push "$ssh_url" HEAD:${pr_branch}
+  python ${WORKSPACE}/rpc-gating/scripts/ghutils.py \
```
git-harry: I think this should be creating a card to which the pull request is linked.

scripts/dep_update.sh
```diff
@@ -0,0 +1,47 @@
...
+  # split URL into array seperated by '/'s
+  IFS='/ ' read -r -a url_split <<< "$URL"
+  # owner/org is whatever is between the second and third / in the url
+  owner="${url_split[3]}"
+  # repo is anything after owner/
+  repo=${URL#https://github.com/${owner}/}
+  ssh_url="git@github.com:${owner}/${repo}"
+  pr_branch="${BRANCH}_dep_update_$(date +%H%M_%d%m%Y)"
+  echo "Repo is dirty, proposing changes"
+  git commit -a -m "${pr_branch}"
```
git-harry: The commit message on the test job is:
```
RE-197_dep_update_1716_27092017
```
I'd like to see this be a bit more human readable. While everyone would eventually understand what that meant, it doesn't cost us to be a bit more descriptive and then the commit will never require explanation. For example,
```
RE-XXX Update {branch} dependencies

This change updates the branch dependencies with the repository
modifications made by running the Jenkins hook in
`gating/update_dependencies`.
```
I added a reference to an issue (see other comments), dropped the date and time because they are recorded by the commit, and added an explanation as to how this commit was generated.

scripts/dep_update.sh
```diff
@@ -0,0 +1,47 @@
...
+  fi
+}
+cd ${WORKSPACE}/repo
+run_hook "gating/update_dependencies/pre"
+run_hook "gating/update_dependencies/run"
+if [[ -z "$(git status -s)" ]]
+then
+  echo "Repo is clean, prep script made no changes to be comitted"
+else
+  # split URL into array seperated by '/'s
```
git-harry: separated

rpc_jobs/dep_update.yml
```diff
@@ -0,0 +1,40 @@
+- project:
+    name: "Dependency-Update-Jobs"
+    repo:
+      - rpc-openstack:
+          URL: "https://github.com/rcbops/rpc-openstack"
+          BRANCH: "newton"
```
git-harry: Why only newton?

scripts/dep_update.sh
hughsaunders: Done

scripts/dep_update.sh
hughsaunders: Done.

scripts/dep_update.sh
hughsaunders: Done

rpc_jobs/dep_update.yml
hughsaunders: Mostly because I wanted to implement an example. If rpco want to update other branches or deps, they'll need to modify the jobs. OSA probably isn't the only dep that RPCO wants to track.

pipeline_steps/common.groovy
```diff
@@ -601,13 +601,14 @@ def create_jira_issue(String project="RE",
...
-        --tag '$tag'\
-        --link '$link'\
-        --project '$project'\
+      python rpc-gating/scripts/jirautils.py \
         --user '$JIRA_USER' \
         --password '$JIRA_PASS' \
-        --type '$type'
+        create_issue \
+          --summary "JBF: ${tag}"\
+          --description "Jenkisn Build Failed :( [${tag}|${link}])"
```
odyssey4me: Spelling error.

scripts/dep_update.sh
```diff
@@ -0,0 +1,76 @@
...
+      exit 1
+    fi
+  fi
+}
+cd ${WORKSPACE}/repo
+run_hook "gating/update_dependencies/pre"
+run_hook "gating/update_dependencies/run"
+if [[ -z "$(git status -s)" ]]
+then
+  echo "Repo is clean, prep script made no changes to be comitted"
```
odyssey4me: I think that it should be ``committed``.

scripts/dep_update.sh
```diff
@@ -0,0 +1,76 @@
...
+  issue=$(python ${WORKSPACE}/rpc-gating/scripts/jirautils.py \
+        --user "${JIRA_USER}" \
+        --password "${JIRA_PASS}" \
+        create_issue \
+          --project "${JIRA_PROJECT_KEY}" \
+          --summary "${jira_summary}" \
+          --description "${message}" \
+          --type "${JIRA_ISSUE_TYPE}" \
+          --label RE_DEP_UPDATE \
+          --existing-issue-query "project = ${JIRA_PROJECT_KEY} AND resolution = Unresolved AND summary ~ \"${jira_summary}\""
```
git-harry: My understanding of this is that any new dependency update will be linked to an existing issue, if one exists, by updating the original pull request. This may end up reducing throughput, for example if the existing issue was in needs review and only waiting to be merged, updating the pull request would require all the tests to be rerun.

I would be inclined to only use an existing issue if it is in the backlog.

scripts/dep_update.sh
```diff
@@ -0,0 +1,77 @@
...
+
+  echo "Committing changes"
+  pr_branch="${issue}_${BRANCH}_dep_update"
+  title="${issue} Update ${BRANCH} dependencies"
+  git commit -a -m "${title}
+
+  ${message}"
+
+  echo "Pushing changes to repo branch"
+  git push -f "$ssh_url" "HEAD:${pr_branch}"
```
git-harry: Jenkins should maintain a fork for creating pull requests. RPCO has chosen to move away from creating pull request branches directly on rcbops/rpc-openstack, it would make sense to be consistent with that approach. Also given the permissions Jenkins has it probably reduces the risk of error.

hughsaunders:
Ran a job to check that common.create_jira_issue still works:
```
library "rpc-gating@RE-197"
env.RPC_GATING_BRANCH="RE-197"
common.shared_slave(){
    common.create_jira_issue()
}
```

Job: https://rpc.jenkins.cit.rackspace.net/job/scratchpipeline/431/ ✅ 
Created Issue: https://rpc-openstack.atlassian.net/browse/RE-658

scripts/dep_update.sh
odyssey4me: That's probably a good idea, although I'd vote for that as an improvement in another card.

scripts/dep_update.sh
odyssey4me: Perhaps the right way to think about this is that if there is an unresolved card which is not in backlog then do not execute the dep update. The thinking there would be that in this situation the team may be dealing with fallout in the PR and the PR needs manual updates to get it into a working state - and preventing the dep update from running also prevents the PR being changed from the team's work to something else.

The only trouble I'm seeing here is that the column it's in could be named differently for each board. :/

hughsaunders: BACKLOG seems quite standard I'll change the query to `project = ${JIRA_PROJECT_KEY} AND status = Backlog AND summary ~ \"${jira_summary}\"`

scripts/dep_update.sh
hughsaunders: I think it woud make sense to make an exception to that rule for Jenkins. Automated branches are namespaced and will not conflict with well known branch names as has happened recently.

If the consensus is that Jenkins must fork, then I agree with @odyssey4me that should be another card. We'll have to auto create forks and update pr creation code to create cross repo PRs.

hughsaunders:
Testing reuse of exisitng issue and PR, but only when the existing issue is in backlog:

[Run 1](https://rpc.jenkins.cit.rackspace.net/job/Dep-Update_rpc-openstack-newton/25/) ✅ : Create issue [663](https://rpc-openstack.atlassian.net/browse/RE-663) and pr rcbops/rpc-openstack#2571
[Run 2](https://rpc.jenkins.cit.rackspace.net/job/Dep-Update_rpc-openstack-newton/26) ✅ : Issue still in backlog, Same issue used, PR updated
[Run 3](https://rpc.jenkins.cit.rackspace.net/job/Dep-Update_rpc-openstack-newton/27/) ✅ : Issue in doing, new issue [664](https://rpc-openstack.atlassian.net/browse/RE-664) and pr rcbops/rpc-openstack#2572 created.

rpc_jobs/dep_update.yml
```diff
@@ -0,0 +1,51 @@
+- project:
+    name: "Dependency-Update-Jobs"
+    repo:
+      - rpc-openstack:
+          URL: "https://github.com/rcbops/rpc-openstack"
+          BRANCH: "master"
```
odyssey4me: How do we make sure that newton is also updated? Would this be the same project or a separate project? Would it be another ``repo`` in the list?

hughsaunders: Another repo in the list.

scripts/dep_update.sh
```diff
@@ -0,0 +1,79 @@
...
+          --type "${JIRA_ISSUE_TYPE}" \
+          --label RE_DEP_UPDATE \
+          --existing-issue-query "project = ${JIRA_PROJECT_KEY} AND status = BACKLOG AND summary ~ \"${jira_summary}\""
+  )
+  echo "Issue: ${issue}"
+
+  echo "Committing changes"
+  pr_branch="${issue}_${BRANCH}_dep_update"
+  title="${issue} Update ${BRANCH} dependencies"
+  git commit -a -m "${title}
```
odyssey4me: You can just do this instead, which is slightly neater:

```
git commit -a -m "${title}" -m "${message}"
```

hughsaunders: Done

hughsaunders:
Test without modifying params: https://rpc.jenkins.cit.rackspace.net/job/Dep-Update_rpc-openstack-master/1/ ✅ 
Issue: https://rpc-openstack.atlassian.net/browse/RO-2730

PR #443, opened 2017-09-28
RE-641 Don't deploy metering services

mattt416:


mattt416:
This is dependent on functionality in https://review.openstack.org/#/c/508269/, which has yet to merge.  Once https://review.openstack.org/#/c/508269/ merges we'll need to update this review with the correct openstack-ansible-ops SHA.

rpc_jobs/params.yml
```diff
@@ -69,7 +69,7 @@
           default: https://github.com/openstack/openstack-ansible-ops
       - string:
           name: OSA_OPS_BRANCH
-          default: 89e45170fe6043ae014c975f391cf3effc54476f
+          default: bc2e128143ed9397e1452bde9987e4a59b39881a
```
hughsaunders: I should have read your comment about this, that SHA is available from github, but isn't reachable by any of the branches or tags. I had to `git fetch origin $(git ls-remote origin |awk '{print $2}')` to find it.

PR #450, opened 2017-10-02
RE-613 Set artifact environment vars in IRR jobs

hughsaunders:
Recently we standardised the artifact location to WORKSPACE/artifacts
but also decided that projects should not rely on that, instead
they should read the RE_HOOK_ARTIFACTS_DIR environment variable
in order to determine the correct place to place artifacts.

This commit ensures that the artifacts and test result env vars
are available in IRR jobs.

Issue: [RE-613](https://rpc-openstack.atlassian.net/browse/RE-613)


odyssey4me:
recheck_project_builds

PR #453, opened 2017-10-02
Update Ansible and adjust MNAIO setup

cloudnull:
Ansible 2.2.2.x seems to have issues with ssh disconnects as noted here:
[Related-Issue]. Reports indicate going to 2.3.x resolves some of the issues.

We don't need to run the "openstack-service-setup.yml" playbook. We deploy
RPC-O in our own way, not following the method in the MNAIO, so this playbook
is not used by us.

SSH params are now being passed into the jenkins slave setup playbook. This is
done to ensure we're not causing ourselves issues with SSH configs as we setup
the Jenkins slave on the infra1 VM instead of the OnMetal Host; failures with
SSH disconnects and host key verifications have been repeatedly observed
in the gate.

To resolve an issue with the RPC-O extra inventory the `dynamic-inventory.py`
script has been converted to write a YAML based inventory instead of using
the JSON type which does not directly support the standard Ansible inventory
extensions.

Related-Issue: https://github.com/ansible/ansible/issues/25941
Issue: https://rpc-openstack.atlassian.net/browse/RE-632
Co-Authored-By: Matt Thompson <matt.thompson@rackspace.com>

Issue: [RE-632](https://rpc-openstack.atlassian.net/browse/RE-632)


mattt416:
Kicked off https://rpc.jenkins.cit.rackspace.net/user/amesserl/my-views/view/Leapfrogs/job/OnMetal-Multi-Node-AIO_r14.2.0-kilo-trusty-leapfrogupgrade-large-periodic/13/ against this patch.  Sadly, it's churning on `TASK [Set proper permissions for Jenkins home directory] ***********************` doing apparently nothing.  I'll let it continue to run and check back in the morning.

I've also kicked off https://rpc.jenkins.cit.rackspace.net/user/amesserl/my-views/view/Leapfrogs/job/OnMetal-Multi-Node-AIO_r14.2.0-kilo-trusty-leapfrogupgrade-large-periodic/14/ and https://rpc.jenkins.cit.rackspace.net/user/amesserl/my-views/view/Leapfrogs/job/OnMetal-Multi-Node-AIO_r14.2.0-kilo-trusty-leapfrogupgrade-large-periodic/15/.

From my own testing, just adding the flush handler helped a fair bit, but didn't resolve the connection issues 100%.

cloudnull:
@mattt416 looks like 13 worked, I think 14 was not using the correct branch/checkout as it tried to run the openstack service setup plays (or the env var was not being passed in correctly), and 15 seems to have failed to provision a node.

Can we retest these three again?

mattt416:
@cloudnull 14 was definitely configured to use `ansible-rev`, but I'll retrigger a few jobs so we can see.

cloudnull:
tracking test builds on:
* https://rpc.jenkins.cit.rackspace.net/user/amesserl/my-views/view/Leapfrogs/job/OnMetal-Multi-Node-AIO_r14.2.0-kilo-trusty-leapfrogupgrade-large-periodic/15
* https://rpc.jenkins.cit.rackspace.net/user/amesserl/my-views/view/Leapfrogs/job/OnMetal-Multi-Node-AIO_r14.2.0-kilo-trusty-leapfrogupgrade-large-periodic/16
* https://rpc.jenkins.cit.rackspace.net/user/amesserl/my-views/view/Leapfrogs/job/OnMetal-Multi-Node-AIO_r14.2.0-kilo-trusty-leapfrogupgrade-large-periodic/17

scripts/dynamic_inventory.py
```diff
@@ -1,61 +1,78 @@
...
+
+
+def host_skel():
+    """Prep the dynamic inventory extension."""
+
+    node_count = 0
+    for k, v in EXTRA_NODES.items():
+        for c in range(0, v['nodes']):
+            node_count += 1
+            host = "%s-e%d" % (k, c)
```
mattt416: Small nit, but how about:

```
host = "%s-e%d" % (k, c+1)
```

cloudnull: done.

pipeline_steps/multi_node_aio_prepare.groovy
```diff
@@ -34,7 +34,7 @@ def prepare() {
           # RAM to VMs -- since we have RAM to spare we double that assigned to
           # infra nodes.
           echo "infra_vm_server_ram: 16384" | sudo tee -a playbooks/group_vars/all.yml
-          cp -a ${WORKSPACE}/rpc-gating/scripts/dynamic_inventory.py playbooks/inventory
+          ${WORKSPACE}/rpc-gating/scripts/dynamic_inventory.py playbooks/inventory/rpc-inventory.yaml
```
odyssey4me: It looks like the ``dynamic_inventory.py`` script is an inventory generator, rather than a dynamic inventory. In other words - it is used to generate a static inventory in ``playbooks/inventory/rpc-inventory.yaml``. If so, perhaps the name should change to something more appropriate, like ``mnaio-inventory-generate.py`` so that where it's used is clear?

playbooks/setup_jenkins_slave.yml
```diff
@@ -26,11 +37,17 @@
         ClientAliveInterval: 15
         Compression: "yes"
   tasks:
+    - meta: flush_handlers
```
odyssey4me: Flushing the handlers reloads sshd, right? If so, do we need to add a wait here to verify that sshd is ready to receive connections again? With Ansible 2.3 I'm thinking something like https://github.com/openstack/openstack-ansible/blob/77ac88197c223968e845e57706b81af24d6aa443/playbooks/lxc-containers-create.yml#L28-L33

playbooks/setup_jenkins_slave.yml
```diff
@@ -26,11 +37,17 @@
...
+
     - name: Install apt packages
       apt:
         pkg: "{{ item }}"
         state: installed
         update_cache: yes
+      register: install_packages
+      until: install_packages | success
+      retries: 5
+      delay: 2
```
odyssey4me: How are retries related here? While this is no harm, I'm just confused about why this is being done - perhaps to accommodate the potential for the connection failing (see comment in task above)?

playbooks/setup_jenkins_slave.yml
```diff
@@ -2,6 +2,17 @@
...
+      -o UserKnownHostsFile=/dev/null
+      -o StrictHostKeyChecking=no
+      -o ServerAliveInterval=64
+      -o ServerAliveCountMax=1024
+      -o Compression=no
+      -o TCPKeepAlive=yes
+      -o VerifyHostKeyDNS=no
+      -o ForwardX11=no
+      -o ForwardAgent=yes
+      -T
```
odyssey4me: @hughsaunders Do you think that these parameters should be added to group_vars to make them used globally by all rpc-gating playbooks or do you think they're better left in this playbook exclusively?

playbooks/setup_jenkins_slave.yml
mattt416: The reason for adding this, to capture, is because at present the restart happens after the slave is connected, which we think is contributing to the slave disconnection that we're facing.

pipeline_steps/multi_node_aio_prepare.groovy
cloudnull: I could rename it. I didn't write it I just updated what we had however I'd be happy to rename it if you think it best.

playbooks/setup_jenkins_slave.yml
cloudnull: there seem to be issues running apt the first time, some times. I guess this could be a connection issue with SSH spinning back up, but it could also be a caching issue in the MNAIO. That said, the retry seems to ensure we get past this spot.

mattt416: I think we saw this in a job, and it was added as a result:

```
TASK [Install apt packages] ****************************************************
failed: [omnarktllp-18-fa33-infra1-vm] (item=[u'git-core', u'default-jre-headless', u'python-dev', u'gcc', u'libffi-dev', u'libssl-dev']) => {"failed": true, "item": ["git-core", "default-jre-headless", "python-dev", "gcc", "libffi-dev", "libssl-dev"], "module_stderr": "Warning: Permanently added '[104.130.18.4]:2222' (RSA) to the list of known hosts.\r\n/usr/bin/python: can't open file '/root/.ansible/tmp/ansible-tmp-1507037495.84-117600361617508/apt.py': [Errno 2] No such file or directory\n", "module_stdout": "", "msg": "MODULE FAILURE", "rc": 0}
	to retry, use: --limit @/var/lib/jenkins/workspace/OnMetal-Multi-Node-AIO_r14.2.0-kilo-trusty-leapfrogupgrade-large-periodic@3/rpc-gating/playbooks/setup_jenkins_slave.retry
```

playbooks/setup_jenkins_slave.yml
cloudnull: I personally think this should be done for all RPC playbooks however to limit impact I've kept it in the one playbook. @odyssey4me @hughsaunders if you'd like to have this in the group vars I'd be happy to move it there.

playbooks/setup_jenkins_slave.yml
cloudnull: done.

pipeline_steps/multi_node_aio_prepare.groovy
cloudnull: done.

cloudnull:
recheck_cit_project_builds

hughsaunders:
The rpc-gating jenkins library is always loaded from rcbops/rpc-gating, as the library load mechanism doesn't have a way of specifying an alternate repo. Even though cloudnull:ansible-rev was rebased, project-builds is using rcbops:ansible-rev which still needs a rebase. Because of this confusion, PRs to rpc-gating should not come from a fork.

hughsaunders:
recheck_cit_project_bulds I've pushed cloudnull:ansible-rev --> rcbops:ansible-rev

hughsaunders:
recheck_cit_project_bulds I've pushed cloudnull:ansible-rev --> rcbops:ansible-rev

hughsaunders:
recheck_project_builds I've pushed cloudnull:ansible-rev --> rcbops:ansible-rev

playbooks/setup_jenkins_slave.yml
hughsaunders: I'm happy for that to go into group vars.

playbooks/vars/all.yml
```diff
@@ -27,3 +27,15 @@ default_gating_overrides:
   # too low for the liberty->mitaka upgrade job where more space is used by
   # additional packages, venvs, logs, etc.
   percent_used_critical_threshold: 95
+
+ansible_ssh_extra_args: >
```
odyssey4me: These need to be in ``playbooks/inventory/group_vars/all.yml`` - this vars file is only used for some RPC-O configuraiton playbooks and does not apply to the single use slave setup.

PR #454, opened 2017-10-04
HUM-330 add ceph job for rpc-maas 

hughsaunders:
This axis will be used for the MaaS jobs to deploy with ceph
or no additional storage backend.

Note that this commit does not implement a ceph deploy for rpc-maas,
it provides a new environment variable (IRR_SCENARIO=deploy/ceph)
which rpc-maas/run_tests.sh can use to deploy and test the correct
scenario. This variable will also be provided to all the other IRR
jobs, but they can safely ignore it until they need multiple scenarios.

Issue: [HUM-330](https://rpc-openstack.atlassian.net/browse/HUM-330)

### DIFF
--- a/rpc_jobs/irr_role_test.yml
+++ b/rpc_jobs/irr_role_test.yml
@@ -20,6 +20,7 @@
       - mitaka
       - liberty
       - kilo
+      - ceph
     exclude:
       # NOTE(cloudnull): The image context is only applied to certain releases.
       - image: trusty
@@ -34,6 +35,8 @@
         context: liberty
       - image: xenial
         context: kilo
+      - image: trusty
+        context: ceph
     jobs:
       - 'RPC-IRR_{repo}-{series}-{image}-{context}'
 


rpc_jobs/irr_role_test.yml
```diff
@@ -34,8 +37,18 @@
         context: liberty
       - image: xenial
         context: kilo
+      - scenario: ceph
```
andymcc: This would just be for master, we aren't deploying rpc-o + ceph, so the kilo/liberty/mitaka bits are meaningless.

rpc_jobs/irr_role_test.yml
```diff
@@ -20,6 +20,9 @@
       - mitaka
       - liberty
       - kilo
+    scenario:
```
odyssey4me: ``context`` was intended initially as ``scenario``, with ``series`` being used for the branch. I'd suggest changing this up so that ``series`` refers to the branch and rename ``context`` to ``scenario`` and using the scenario there.

rpc_jobs/irr_role_test.yml
odyssey4me: @andymcc Yeah, this is excluding all of those branches so that it only uses the ``ceph`` scenario for the master branch.

rpc_jobs/irr_role_test.yml
hughsaunders: Series is the branch of the repo under test (rpc-maas), context in this case is the branch of rpco to test against. As only master is required, ceph could be an additional context, then we could add another axis later if required.

andymcc:
recheck_cit_project_builds

PR #458, opened 2017-10-06
RE-674 Adjust apt sources if necessary

odyssey4me:
When setting up a single-use slave, we need to install
packages. Some of those packages may also be installed
later after apt sources are reconfigured (as is the case
with RPC-O), but the packages in those sources are older
and the package database therefore errors.

In this patch we try to detect whether we need to use
the apt artifacts as a source and configure them before
installing any packages. This hopes to prevent jobs
failures due to the packages we install for Jenkins.


odyssey4me:
Tested using this branch in https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_newton-trusty-deploy-swift-periodic/67 and it appears to be doing the right thing.

cloudnull:
This looks like a fine approach. @odyssey4me would it be possible to have a "lastest" repo for our apt archive and simply have gating point at that via pre-built images by default?

odyssey4me:
@cloudnull It might be possible to do that, but that would cause problems for jobs that implement older versions of RPC-O for the purpose of upgrade testing. The current approach of looking at the rpc_release and using that repo is the best option to cater for the broadest range of jobs as far as I can see.

antonym:
The approach makes sense, seems a bit safer than downgrades for now.

mattt416:
I think we're holding off on this while @hughsaunders tests some things (as stated in RE-674).

PR #459, opened 2017-10-06
RE-506 Remove rpc-gating project_builds job

odyssey4me:
With the implementation of the pre_merge_test
template this job is no longer required.

Issue: [RE-506](https://rpc-openstack.atlassian.net/browse/RE-506)


rpc_jobs/single_use_slave_template.yml
```diff
@@ -1,58 +0,0 @@
-- job:
```
hughsaunders: I think the single use slave template still has some value as an example. So its enough to remove the project builds job.

PR #461, opened 2017-10-06
RE-677 Fix KIBANA_SELENIUM_BRANCH override

antonym:
Issue: [RE-677](https://rpc-openstack.atlassian.net/browse/RE-677)

### DIFF
--- a/rpc_jobs/rpc_aio.yml
+++ b/rpc_jobs/rpc_aio.yml
@@ -284,7 +284,10 @@
                 && env.ACTION == "leapfrogupgrade" ){{
                 env.UPGRADE_FROM_REF="origin/pr/${{env.ghprbPullId}}/merge"
                 kibana_branch = "kilo"
-                env.KIBANA_SELENIUM_BRANCH="newton-14.1"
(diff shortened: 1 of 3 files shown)


rpc_jobs/rpc_aio.yml
```diff
@@ -286,6 +286,8 @@
                 kibana_branch = "kilo"
                 env.KIBANA_SELENIUM_BRANCH="newton-14.1"
             }}
+            // set kibana branch to newton for leapfrogs, see RE-677
+            if (env.ACTION == "leapfrogupgrade" {{ env.KIBANA_SELENIUM_BRANCH="newton" }}
```
mattt416: You could remove the assignment on line 287 since this new if block here will essentially cover that case also.

antonym:
Looks like this fixed the job:

https://rpc.jenkins.cit.rackspace.net/view/Leapfrog/job/RPC-AIO_r14.2.0-kilo-trusty-leapfrogupgrade-swift-periodic/

antonym:

PR #465, opened 2017-10-10
RLM-228 Change kernel to linux-image-generic

antonym:
The package, linux-image-generic, should be used by the MNAIO to ensure
we're testing with the supported LTS kernel for the distro being
deployed.

Upstream:

https://github.com/openstack/openstack-ansible-ops/commit/cc9294bb0d4191bf6a78ef15b02bcaed86cdd709

Issue: [RLM-228](https://rpc-openstack.atlassian.net/browse/RLM-228)

### DIFF
--- a/rpc_jobs/params.yml
+++ b/rpc_jobs/params.yml
@@ -84,7 +84,7 @@
           description: Version of Ubuntu image to use for VMs (14.04.5 or 16.04.2)
       - string:
           name: DEFAULT_KERNEL
-          default: "linux-image-4.4.0-66-generic"
+          default: "linux-image-generic"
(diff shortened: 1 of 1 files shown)


weezer:
i think this one is related to the mn-aio env, 
https://github.com/openstack/openstack-ansible-ops/blob/master/multi-node-aio/README.rst
Instruct the VM to use the selected kernel meta package, eg. linux-generic:
DEFAULT_KERNEL=${DEFAULT_KERNEL:-linux-image-generic}

at the beginning of this year when you point to linux-image-generic it will give you 3.xx kernel not the 4.xxx if you dont see this issue i think it good to go.

PR #467, opened 2017-10-10
RE-634 Update Merge-Trigger-JJB

mattt416:
This job is currently not firing, presumably because of the change in
[1].  This change moves the conditional directly around the build step
so that the git step is not masked by the conditional.

We also update the conditional to trigger if ${env.GIT_BRANCH} is null
so that we can trigger this job manually for testing.

Lastly, we tweak the print statement to include the branch, for
debugging purposes.

[1] https://github.com/rcbops/rpc-gating/pull/434

Issue: [RE-634](https://rpc-openstack.atlassian.net/browse/RE-634)


mattt416:
WIP, changes are still being tested :)

PR #471, opened 2017-10-11
K8S-434 Add rpc-octavia to CI

xgerman:
Adds the pre-merge job for rpc-octavia

Issue: [K8S-434](https://rpc-openstack.atlassian.net/browse/K8S-434)

### DIFF
--- a/rpc_jobs/rpc_octavia.yml
+++ b/rpc_jobs/rpc_octavia.yml
@@ -1,0 +1,30 @@
+ - project:
+    name: "rpc-octavia"
+
+    # URL of the rpc-octavia repository
+    repo_url: "https://github.com/rcbops/rpc-octavia"
+
+    # currently we only do master. Once RPC-O switches
+    # to Pike we will retire os-octavia and use the
+    # upstream installer
+    branches:
+      - "master"
+
+    # Use image for RPC-O install
+    image:
+      - xenial:
+          FLAVOR: "performance2-15"
+          IMAGE: "Ubuntu 16.04.2 LTS prepared for RPC deployment"
+
+    # Octavia ignores that setting for now
+    scenario:
+      - "functional"
+
+    # Octavia ignores that setting for now
+    action:
+      - "test"
+
+    # Link to the standard pre-merge-template
+    jobs:
+      - 'PR_{name}-{image}-{scenario}-{action}'
+    # add post_merge whe available


rpc_jobs/rpc_octavia_gating.yml
```diff
@@ -0,0 +1,46 @@
+ - project:
+    name: "rpc-octavia"
+
+    # This is the URL for the project's repository
```
odyssey4me: nit: You can remove all the comments and just use your own. The comments are in the documentation anyway.

rpc_jobs/rpc_octavia_gating.yml
```diff
@@ -0,0 +1,46 @@
...
+      - "master"
+
+    # This is a list of names to designate the image
+    # used when preparing a cloud instance. Each name
+    # must also include the key IMAGE with a value
+    # that corresponds to the appropriate image UUID
+    # or name of the image available in the cloud.
+    image:
+      - xenial:
+          IMAGE: "Ubuntu 16.04 LTS (Xenial Xerus) (PVHVM)"
```
odyssey4me: What flavor would you need to do your tests on - the default is performance1-1 which will not be able to run RPC-O. If you want the same as RPC-O you'll need to use the following immediately under the IMAGE.

```
FLAVOR: "performance2-15"
```

Also, if you're planning on executing RPC-O installs, then you should be using this value for IMAGE: ``Ubuntu 16.04.2 LTS prepared for RPC deployment``

rpc_jobs/rpc_octavia_gating.yml
```diff
@@ -0,0 +1,46 @@
+ - project:
+    name: "rpc-octavia"
```
evrardjp: nit: this should probably be renamed, for consistency, RPC-octavia-Jobs

The filename itself should probably be rpc-octavia.yml also for consistency.

rpc_jobs/rpc_octavia_gating.yml
evrardjp: +1

rpc_jobs/rpc_octavia_gating.yml
evrardjp: +1

rpc_jobs/rpc_octavia_gating.yml
```diff
@@ -0,0 +1,46 @@
...
+    # for the 'scenario' which affect the code path
+    # tested.
+    # Octavia ignores that
+    action:
+      - "test"
+
+    # This final configuration links this project to the
+    # standard template used.
+    jobs:
+      - 'PR_{name}-{image}-{scenario}-{action}'
```
evrardjp: Are you sticking with the gating project PR job?
If you need very different behavior, you may need to define your own job-template and job.
It looks like in the repo you're planning to use this, and the pre/run steps, so I am expecting a yes there.

In that case, shouldn't a ``Execute Pre Script`` stage be included?
Because they aren't defined in the template:
https://github.com/rcbops/rpc-gating/blob/e1a2a69b24e37a93d2a7d2e1b6c0c44f207b3012/rpc_jobs/standard_job_premerge.yml#L33-L46

evrardjp:
FYI, you should probably source the bash script instead of bash <script> in https://github.com/rcbops/rpc-octavia/blob/master/gating/scripts/pre.sh#L37

rpc_jobs/rpc_octavia_gating.yml
odyssey4me: @evrardjp That is the old standard. The name here is correct.

rpc_jobs/rpc_octavia_gating.yml
odyssey4me: @evrardjp This is implementing the new standard job. Please see https://rpc-openstack.atlassian.net/wiki/spaces/RE/pages/19005457/RE+for+Projects

evrardjp: Oh ok. I thought it was ALL steps, not only optional steps.

odyssey4me: @evrardjp In the template there are no conditional steps, so the STAGES var is now only useful for adjusting the setup/teardown of the single use slave. eg: Adding the ``Pause`` stage.

evrardjp: That is a different conversation, shouldn't be on this review.

rpc_jobs/rpc_octavia_gating.yml
evrardjp: OK.

xgerman:
https://github.com/rcbops/rpc-octavia/blob/master/gating/scripts/pre.sh#L37 I was going for isolating the two since in a later install the os-octavia step might be weeks later after the rpc-o install so I don't like the env variables set in the RPC-O deploy script impact the later script

rpc_jobs/rpc_octavia.yml
```diff
@@ -0,0 +1,30 @@
...
+      - "functional"
+
+    # Octavia ignores that setting for now
+    action:
+      - "test"
+
+    # Link to the standard pre-merge-template
+    jobs:
+      - 'PR_{name}-{image}-{scenario}-{action}'
+    # add post_merge whe available
```
git-harry: nit: s/whe/when/

PR #475, opened 2017-10-12
RE-643 Phobos <--> Webhook Proxy Tunnel

hughsaunders:
As internal routing between CIT jenkins and the Phobos environment
is not possible in the short term, a tunnel is added.

The deploy-phobos container in the phobos environment will connect
to the webhook proxy node via ssh. The ssh connection will carry
a reverse port forward such that any connections to port 2223 on
the webhook proxy  node will be forwarded to port 22 on deploy-phobos.

Jenkins will ssh to webhookproxy:2223 when it requires a connection
to deploy-phobos.

Issue: [RE-643](https://rpc-openstack.atlassian.net/browse/RE-643)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -48,12 +48,17 @@ void install_ansible(){
     np=\${PWD}/.venv                    # New Path
     grep -ri --files-with-match \$op \
       |while read f; do sed -i.bak "s|\$op|\$np|" \$f; done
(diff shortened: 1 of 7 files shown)


hughsaunders:
Note that the changes in this PR have already been enacted, and the [phobos deployment host is now a jenkins slave](https://rpc.jenkins.cit.rackspace.net/computer/deploy-phobos/).

PR #477, opened 2017-10-12
RE-634 Change Merge-Trigger-JJB to freestyle job

mattt416:
Currently, Merge-Trigger-JJB fires irrespective of the branch on
rcbops/rpc-gating being pushed to.  We tried to circumvent this by
using GIT_BRANCH inside the pipeline script, however that environment
variable is not being passed in and therefore this job will never build
the Jenkins-Job-Builder job.

Since there is no clear way to get a pipeline script to only trigger
when a specific branch is being pushed to, we need to convert this job
to a freestyle job where we can specify the scm and specific branch(es)
we wish to build against (master, in this case).

NOTE: It's likely we'll need to delete the existing Merge-Trigger-JJB
      job when this merges, as I don't believe JJB will change project
      types on an existing job.

Issue: [RE-634](https://rpc-openstack.atlassian.net/browse/RE-634)

### DIFF
--- a/rpc_jobs/jjb_setup.yml
+++ b/rpc_jobs/jjb_setup.yml
@@ -1,32 +1,27 @@
-# the github trigger for pipeline/worflow jobs will only trigger for pushes
-# to the branch that was built previously. This means that the job must
-# be run manually once to establish the branch.
-# This is also why there is separate merge trigger job, as it ensures that
-# JJB still runs on pushes to master, even if the JJB job is used to test_client
-# an alternative branch.
+# This job is specified as `freestyle` as using a pipeline job with github
+# trigger causes the job to be triggered irrespective of which branch was
+# pushed to.  Additionally, the environment variables we were trying to use to
+# determine the branch to pushed to were not being passed into the job by the
+# plugins in question.
+# Also, having a separate merge trigger job ensures that JJB still runs on
+# pushes to master, even if the JJB job is used to test an alternative branch.
 - job:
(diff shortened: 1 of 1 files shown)


rpc_jobs/jjb_setup.yml
```diff
@@ -1,32 +1,27 @@
...
-            } // stage
-          } // node
-      } else {
-        print ("Not running JJB as job was triggered by a push to a branch other than master")
-      }
+        - github
+    builders:
+      - trigger-builds:
+        - project:
+            - "Jenkins-Job-Builder"
```
mattt416: To point out, builds on this job aren't blocked while Jenkins-Job-Builder runs, but I don't believe that's an issue.  Please correct me if I'm wrong!

git-harry: I may be misunderstanding what you're referring to but isn't the default value for `concurrent: false` and so they would sit in a queue if an existing build was in progress.

mattt416: You are right -- but you can have this job basically wait for the triggered `Jenkins-Job-Builder` to finish before it moves on -- and that's what I was referring to.  However, since I didn't block, it means jobs here will run super quickly, so having `concurrent: false` as the default shouldn't be a problem.

We can always flip things around if there's a compelling reason to do so.

mattt416:
@git-harry I did test this as a separate job but, didn't leave it around since we had two jobs triggering `Jenkins-Job-Builder` jobs.

PR #486, opened 2017-10-17
[WIP] RLM-259 Tune sshd settings on setup_jenkins_slave

antonym:
Attempting to keep the connection to the slave a bit
more stable to avoid the channel termination errors.

Issue: [RLM-259](https://rpc-openstack.atlassian.net/browse/RLM-259)

### DIFF
--- a/playbooks/setup_jenkins_slave.yml
+++ b/playbooks/setup_jenkins_slave.yml
@@ -31,6 +31,9 @@
         PrintLastLog: "no"
         GatewayPorts: "yes"
         Compression: "yes"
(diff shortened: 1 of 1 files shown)


playbooks/setup_jenkins_slave.yml
```diff
@@ -130,6 +133,15 @@
           retries: 3
           delay: 2
 
+    - meta: flush_handlers
```
odyssey4me: I don't see any handlers between the last one and this one. Which handlers are being flushed here?

PR #488, opened 2017-10-18
RE-759 Fix notification e-mail message

git-harry:
Issues:
The e-mail message body is set to the file path where the release notes
are saved. The command-line option `--body` is being incorrectly passed
the path instead of the actual text to use in the message.

In general, the sub-command `mailgun` should be run after
`generate_release_notes` and so the message body should not need to be
supplied however the click context object is not updated to include the
release notes and so the incorrectly supplied path is used instead.

Release notes should be formatted in GitHub Markdown, e-mail clients are
not guaranteed to present this in a reader-friendly format.

Change:
This change removes from `ghutils.create_release` the responsibility of
defining the message subject and consolidates all control of the e-mail
to the notifications file. To help to keep the separation of concerns,
`ghutils.create_release` now updates the context object to include the
link to the new release so it can be used in the message instead of
sending the release notes text.

`notifications.generate_message_data` is used to unify all the common
code, in `notifications.mail` and `notifications.mailgun`, that is being
used to inspect the context object and determine the sender, subject,
and body of the message.

Issue: [RE-759](https://rpc-openstack.atlassian.net/browse/RE-759)

Tested with: https://rpc.jenkins.cit.rackspace.net/job/RE-Release/107/
@hughsaunders would you mind checking the e-mail you received for this test.

### DIFF
--- a/scripts/notifications.py
+++ b/scripts/notifications.py
@@ -34,6 +34,51 @@ def try_context(ctx_obj, var, var_name, context_attr):
     return var
 
 
+def generate_message_data(subject, body):
+    """Return e-mail message data."""
+    ctx_obj = click.get_current_context().obj
+    owner = ctx_obj.owner.login
+    repo = ctx_obj.name
+
+    if not subject:
+        try:
+            version = ctx_obj.version
+        except AttributeError:
+            logger.error(
+                "`version` is missing from the Click context object, if "
+                "this notification is not being sent at the same time as the "
+                "release is created, `--subject` must be supplied."
+            )
+            raise
+        subject = "New release: {o}/{r} version {v}".format(
+            v=version,
+            o=owner,
+            r=repo,
+        )
+
+    if not body:
+        try:
+            url = ctx_obj.release_url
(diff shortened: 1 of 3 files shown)


scripts/notifications.py
```diff
@@ -48,21 +93,18 @@ def cli(debug):
...
     "--subject",
     help="Subject of release announcement message."
-         " May be omitted if create_release is used as that"
-         " generates a subject.")
+         " May be omitted if create_release is used.")
 @click.option(
     "--body",
-    help="Contents of release announcement message"
-         " May be omitted if generate_release_notes is used.")
+    help="Body of release announcement message."
```
hughsaunders: Please add the output of [usage](https://github.com/rcbops/rpc-gating/blob/master/scripts/release.py#L174) to [release.yml](https://github.com/rcbops/rpc-gating/blob/master/rpc_jobs/release.yml#L71) so that the help text is in sync with the script.

git-harry: Done.

PR #492, opened 2017-10-19
RE-784 Standardise pre and post job names

mattt416:
Currently, standard_job_postmerge.yml and standard_job_premerge.yml use
the project name in the job template name.  For rpc_openstack.yml as it
stands, this worked fine.  However, as we go to add post merge jobs into
that file we run into a problem where we cannot use duplicate project
names and need to start injecting additional identifiers into the
project names to avoid collision.  The downside here is that these
names are injected into the job names meaning we'll end up with a slew
of job names that are inconsistent and contain potentially redundant
information.

To address, this commit does the following:

1. Changes the job template name in standard_job_postmerge.yml to
   `PM_{repo_name}-{branch}-{image}-{scenario}-{action}`.  The idea
   here is that we specify a `repo_name` in each project, leaving the
   project's name to be descriptive for the purpose of the project and
   not used in the job name at all.
2. Defaults `branch` in standard_job_postmerge.yml to `master` so that
   this variable can be left out if the project is testing their master
   branch.
3. Similarily, we update standard_job_premerge.yml by changing the job
   template name to
   `PR_{repo_name}-{series}-{image}-{scenario}-{action}`.  In addition
   to `repo_name`, we add `series` as an identifier for the series
   being tested.  This too defaults to master and most likely won't be
   specified outside of rpc-openstack.
4. Updates all existing projects to use the updated templates.
5. Adds a new post-merge project for rpc-openstack's master branch,
   which depends on PR [1]
6. Updates the DSL in standard_job_postmerge.yml to remove the
   splitting of env.REPO_URL, since we now pass repo_name into the
   template.
7. Updates lint_naming_conventions.py since the underscore in
   `repo_name` was causing the regex to fail against
   `PM_{repo_name}-{branch}-{image}-{scenario}-{action}`.
8. Removes `properties([pipelineTriggers([githubPush()])])` from
    standard_job_postmerge.yml since this is causing PM jobs to trigger
    on GitHub pushes.
9. Add an exclusion to rpc_aio.yml so we do not run periodics for the
     master deploy job twice.

PLEASE DO NOT MERGE UNTIL [1] MERGES

[1] https://github.com/rcbops/rpc-openstack/pull/2610

Issue: [RE-784](https://rpc-openstack.atlassian.net/browse/RE-784)

### DIFF
--- a/rpc_jobs/standard_job_postmerge.yml
+++ b/rpc_jobs/standard_job_postmerge.yml
@@ -1,16 +1,19 @@
 - project:
     name:       "gating-post-merge"
+    repo_name:  "rpc-gating"
     repo_url:   "https://github.com/rcbops/rpc-gating"
-    branch:     "master"
+    # `branch` defaults to `master` and can be omitted.
+    #branch:     "master"
     image:      "xenial"
     scenario:   "functional"
     action:     "test"
     jira_project_key: "RE"
     jobs:
-      - 'PM_{name}-{branch}-{image}-{scenario}-{action}'
+      - 'PM_{repo_name}-{branch}-{image}-{scenario}-{action}'
 
 - job-template:
-    name: 'PM_{name}-{branch}-{image}-{scenario}-{action}'
+    name: 'PM_{repo_name}-{branch}-{image}-{scenario}-{action}'
+    branch: "master"
     project-type: pipeline
     concurrent: true
     FLAVOR: "performance1-1"
@@ -40,7 +43,6 @@
 
     dsl: |
       library "rpc-gating@${{RPC_GATING_BRANCH}}"
-      properties([pipelineTriggers([githubPush()])])
 
       env.STAGES="Allocate Resources, Connect Slave, Cleanup, Destroy Slave"
 
@@ -53,6 +55,7 @@
       env.RE_JOB_ACTION = "{action}"
       env.RE_JOB_FLAVOR = "{FLAVOR}"
       env.RE_JOB_TRIGGER = "PM"
+      env.RE_JOB_REPO_NAME = "{repo_name}"
       env.RE_JOB_BRANCH = "{branch}"
 
       // Not part of the published interface, used by this job later on
@@ -72,8 +75,7 @@
 
             try {{
               ansiColor('xterm') {{
-                String repo_name = env.REPO_URL.split('/')[-1]
-                dir("${{env.WORKSPACE}}/${{repo_name}}") {{
+                dir("${{env.WORKSPACE}}/${{env.RE_JOB_REPO_NAME}}") {{
                   withCredentials(common.get_cloud_creds()) {{
 
                     stage('Checkout') {{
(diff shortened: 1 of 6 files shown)


mattt416:
This still needs some testing, and we'll need to coordinate with projects that have jobs rename.  :-/

mattt416:
@odyssey4me good catch, I've excluded the periodic for master.

Yeah, I wanted to get a single job in which we can verify works, and then I'll backport everything in rpc-openstack and then move everything over wholesale in rpc-gating.

odyssey4me:
@mattt416 It probably makes sense to do all the 'deploy' actions first in this card, then do cards for each other action (minorupgrade, etc), because they'll likely take quite a bit more time.

mattt416:
https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-master-xenial-swift-deploy/7/ is testing the post merge PR in rpc-openstack.

mattt416:
@odyssey4me That's a good point.  By excluding the PR jobs when we added the new standard pre merge template we inadvertently removed the minorupgrade, majorupgrade, leapfrog, etc. jobs which we don't have coverage for yet w/ the in-repo scripts.  The same will apply when we start excluding periodics.  Do we want to just remove those exclusions and have double-testing in the interim, or continue on as we've been doing?

odyssey4me:
@mattt416 The exclusion of upgrade tests for PR's was on purpose - all testing for those has been moved to periodics only. So when excluding the periodics, initially it'd have to be only for those tests actually replaced by the PM tests. Double-testing would be ok given it's periodics, but a waste of resources.

mattt416:
@odyssey4me I'll remove the exclusion then, since we're not ready to exclude all periodics in rpc_aio.yml and it not being possible to exclude the specific master periodic deploy job.

odyssey4me:
@mattt416 I vaguely remember @hughsaunders mentioning that it is possible to use more than two axes in an exclusion.

mattt416:
Running a quick test w/ a 3 axis exclusion.

mattt416:
@odyssey4me seemed to work!  I tried excluding this (newton because it has far more combinations):

```
      - series: newton
        ztrigger: periodic
        action: deploy
```

... and this is what got removed:

(.venv) MS43FCG8WN:~ matt0199$ diff 1.jjb 3.jjb
67d66
< INFO:jenkins_jobs.builder:Job name:  RPC-AIO_newton-trusty-deploy-swift-periodic
71d69
< INFO:jenkins_jobs.builder:Job name:  RPC-AIO_newton-xenial-deploy-swift-periodic
(.venv) MS43FCG8WN:~ matt0199$

rpc_jobs/standard_job_postmerge.yml
```diff
@@ -53,6 +55,7 @@
       env.RE_JOB_ACTION = "{action}"
       env.RE_JOB_FLAVOR = "{FLAVOR}"
       env.RE_JOB_TRIGGER = "PM"
+      env.RE_JOB_REPO_NAME = "{repo_name}"
```
hughsaunders: This should be documented in RE-For-Projects, or not injected into the environment (remove env.)

PR #499, opened 2017-10-24
RLM-275 Install python-minimal on xenial hosts for ansible

antonym:
Ansible bombs without this in place on the OnMetal Xenial image as python 2.x isn't installed by default.  This leverages cloud-init to install the package.  Could also load this via the first ansible playbook as the first option we run using the raw module but this seems a bit cleaner.

python-yaml is installed at this time too since mnaio_inventory_generate.py fires early.

I could also maybe drop the apt-get update, but wasn't sure if a cache exists or not on these images.

Issue: [RLM-275](https://rpc-openstack.atlassian.net/browse/RLM-275)

### DIFF
--- a/pipeline_steps/pubcloud.groovy
+++ b/pipeline_steps/pubcloud.groovy
@@ -106,6 +106,7 @@ String getPubCloudSlave(Map args){
           env.RAX_CREDS_FILE = pyrax_cfg
           common.venvPlaybook(
             playbooks: ["allocate_pubcloud.yml",
+                        "instance_prep.yml",
                         "drop_ssh_auth_keys.yml"],
             args: [
               "-i inventory",
--- a/playbooks/instance_prep.yml
+++ b/playbooks/instance_prep.yml
@@ -1,0 +1,6 @@
+- hosts: job_nodes
+  remote_user: root
+  gather_facts: False
+  tasks:
+    - name: RLM-275 Adds python packages if python is not present
+      raw: test -e /usr/bin/python || (apt -y update && apt install -y python-minimal python-yaml)


mattt416:
Hmmm, the functional test didn't look happy:

```
TASK [Fail if the required number of instances aren't available] ***************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "At least one public cloud instance failed to start :("}
	to retry, use: --limit @/var/lib/jenkins/workspace/PR_rpc-gating-master-xenial-functional-test/rpc-gating/playbooks/allocate_pubcloud.retry
```

mattt416:
Testing this [here](https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-master-xenial-swift-deploy/20/console), but based on the failure of the functional job I suspect this too is going to fail.

antonym:

mattt416:
Good point @odyssey4me -- for onmetal instances this is a non-issue since we're installing the package on the host and won't use the rpco frozen repos on the host itself, but this will affect aio instances.

odyssey4me:
@mattt416 @hughsaunders It might be sensible for us to actually only use raw/shell tasks for the slave setup and convert our venv into something that includes its own python executable so that we require as little as possible on the target and all our bits are self-contained.

antonym:
Pushed up some changes, was going to do a when: flavor | search("/onmetal/") but that requires for facts to be gathered.  This at least does a simple test to see if python is available and if not attempts to install.  This seemed like the least intrusive way to do it.

antonym:
Tested on Xenial here:

https://rpc.jenkins.cit.rackspace.net/job/OnMetal-Multi-Node-AIO_newton-xenial-deploy-none-periodic/124/console

antonym:
It works in user data as well but I'll give this a try.  If we have issues with it we can move it there.  This gives us a bit more visibility and logging output if it fails for some reason.

PR #501, opened 2017-10-27
RE-645 Add phobos deployment job

hughsaunders:
This job will be used to deploy RPC Openstack into the phobos
environment. It includes slack warnings for pending deploys and
Jira issue creation on failure.

Depends on:
- [x] https://github.com/rcbops/rpc-maas/pull/380
- [x] https://github.com/rcbops/rpc-eng-ops/pull/123

Notes:
- ⚠️  Please review the deploy portion of this job in detail, as **it has not been tested**. This is because I wanted to have the code reviewed before running it against a production cluster.
- ✅  The verify portion of this job is currently [running every 20 minutes](https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/Phobos-Deploy/activity), and so far all runs of the latest version have succeeded. 

Issue: [RE-645](https://rpc-openstack.atlassian.net/browse/RE-645)

### DIFF
--- a/rpc_jobs/phobos.yml
+++ b/rpc_jobs/phobos.yml
@@ -1,0 +1,194 @@
+- job:
+    name: Phobos-Deploy
+    project-type: workflow
+    concurrent: false
+    properties:
+      - build-discarder:
+          num-to-keep: 60
+    parameters:
+      # See params.yml
+      - rpc_gating_params
+      - rpc_repo_params:
+          RPC_BRANCH: newton
+      - rpc_eng_ops_params:
+          RPC_ENG_OPS_BRANCH: master
+      - string:
+          name: RPC_MAAS_BRANCH
+          default: master
+          description: |
+            Branch of rpc-maas repo used for verify-maas in the verification
+            stages. This does not affect the version of rpc-maas that is
+            deployed. That is determined by the version required by rpco.
+      - bool:
+          name: DEPLOY
+          default: true
+          description: Execute a deploy? Disable for verify only. Jira issues are created for failures only if deploy is enabled.
+      - bool:
+          name: SLACK_NOTIFICATIONS
+          default: true
+          description: Send slack notifications pre and post deploy?
+      - bool:
+          name: PRE_DEPLOY_PAUSE
+          default: true
+          description: Wait 15 minutes between warning users and executing the deploy?
+      - choice:
+          name: MAAS_CREDENTIALS_METHOD
+          choices:
+            - "Token obtained via Jenkins Service Account"
+            - "Token supplied at runtime via Job Parameter"
+          description: "How to get the MaaS token."
+      - string:
+          name: MAAS_TOKEN
+          description: |
+            Required if MAAS_CREDENTIALS_METHOD is "Token supplied at runtime via Job Parameter"
+            This should be an authentication token valid against the MaaS api for the phobos dedicated account.
+            Tokens can be obtained from pitchfork.
+
+    dsl: |
+      void slack_notify(message, color){
+        if(env.SLACK_NOTIFICATIONS == "true"){
+          msg_plus_link="${message} (${env.BUILD_URL})"
+          slackSend(channel: "#rpc-releng", message: msg_plus_link, color: color)
+          slackSend(channel: "#rpc-eng-ops", message: msg_plus_link, color: color)
+        }
+      }
+      void verify(){
+        sshagent (credentials:['rpc-jenkins-svc-github-ssh-key']){
+          sh """#!/bin/bash
+      # Deliberate outdent to match bash's heredoc requirements.
+      sudo -E bash <<EOF
+              cd rpc-eng-ops/phobos-re-automation
+              # Need a newer ansible than newton supports, for using the
+              # shade openstack modules
+              rm -rf phobos_verification_venv
+              virtualenv phobos_verification_venv >/dev/null
+              . phobos_verification_venv/bin/activate > /dev/null
+              set -xeu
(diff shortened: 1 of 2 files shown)


rpc_jobs/phobos.yml
```diff
@@ -0,0 +1,194 @@
...
+          name: RPC_MAAS_BRANCH
+          default: master
+          description: |
+            Branch of rpc-maas repo used for verify-maas in the verification
+            stages. This does not affect the version of rpc-maas that is
+            deployed. That is determined by the version required by rpco.
+      - bool:
+          name: DEPLOY
+          default: true
+          description: Execute a deploy? Disable for veryify only. Jira issues are created for failures only if deploy is enabled.
```
mattt416: s/veryify/verify/

rpc_jobs/phobos.yml
```diff
@@ -0,0 +1,194 @@
...
+              common.internal_slave(){
+                dir("rpc-eng-ops"){
+                  git branch: env.RPC_ENG_OPS_BRANCH, url: env.RPC_ENG_OPS_REPO, credentialsId: 'rpc-jenkins-svc-github-ssh-key'
+                }
+                sh """#!/bin/bash
+                  scl enable python27 ${WORKSPACE}/rpc-eng-ops/phobos-re-automation/maas_creds_check.sh
+                """
+                // Ensure that the MAAS_TOKEN env var is correct
+                // whether the token was obtained from the service account
+                // or supplied via job param.
```
mattt416: Do these comments need to be moved above the actual execution of `maas_creds_check.sh`?

rpc_jobs/phobos.yml
```diff
@@ -0,0 +1,194 @@
...
+            stage("Deploy"){
+              if (env.DEPLOY == "true"){
+                sh """#!/bin/bash -xeu
+                  sudo -E rpc-eng-ops/phobos-re-automation/deploy.sh
+                """
+              }
+            }
+            stage("Post Deployment Verification"){
+              // no point verifying twice if we're not running a deploy
+              if (env.DEPLOY == "true"){
```
mattt416: Would it make sense to move the stages inside the conditionals so that they only run if the condition is true?  Would that make tracing the job in the UI more straight-forward because otherwise these jobs are going to have empty stages.

rpc_jobs/phobos.yml
```diff
@@ -0,0 +1,194 @@
...
+            common.create_jira_issue("RE", "PHOBOS DEPLOY / ${env.BUILD_TAG}")
+          } else {
+            slack_notify("Phobos Verification Failed", "warning")
+          }
+          throw e
+        } finally {
+          sh """#!/bin/bash -xeu
+            # All files in the workspace must be owned by jenkins
+            # so they can be cleaned up by jenkins.
+            sudo chown -R jenkins:jenkins . &>/dev/null
```
mattt416: This seems dangerous.  It may be guaranteed that you're in ${WORKSPACE} here, but I'd do an explicit cd incase something doesn't end up where you think it is.

rpc_jobs/phobos.yml
```diff
@@ -0,0 +1,194 @@
...
+                if (env.PRE_DEPLOY_PAUSE == "true"){
+                  slack_notify("Phobos Deploy starting in 15 minutes.", "warning")
+                  sleep(time: 15, unit: "MINUTES")
+                }
+                slack_notify("Phobos Deploy starting now.", "warning")
+            }
+            stage("Deploy"){
+              if (env.DEPLOY == "true"){
+                sh """#!/bin/bash -xeu
+                  sudo -E rpc-eng-ops/phobos-re-automation/deploy.sh
```
mattt416: Has the env been (re)deployed with this tool already?

hughsaunders: rcbops/rpc-eng-ops#123 is listed as a dep in the PR description. But no, the deploy script has not been run, I wanted it to be reviewed before its executed against the live cluster.

rpc_jobs/phobos.yml
hughsaunders: Empty stages are ok in blue ocean - they show as empty so theres no confusion. Also keeping empty stages prevents the stage table from resetting in the old UI.

rpc_jobs/phobos.yml
hughsaunders: No because they related to the line below, where the env var MAAS_TOKEN is re-set.

rpc_jobs/phobos.yml
hughsaunders: Fixed the typo, I think the defaults should reflect the best settings for running a deploy. This will be important when we eventually automate this. 

This job is not safe, and shouldn't be executed by anyone who doesn't know what they are doing..

rpc_jobs/phobos.yml
hughsaunders: Yes.

hughsaunders:
@mattt416 Addressed your issues except backup, there is currently no attempt to backup pre deploy in this PR or rcbops/rpc-eng-ops#123.

hughsaunders:
@mattt416 added [backup config function](https://github.com/rcbops/rpc-eng-ops/pull/123/files#diff-58ab054b72a04c165a33016528f6b983R39).

rpc_jobs/phobos.yml
mattt416: Awesome, thanks @hughsaunders !

hughsaunders:
Dependencies have merged, 2x +2, going in.

PR #503, opened 2017-10-27
RE-581 JJB Job deletion fix

hughsaunders:
Currently the JJB job has two boolean params:
 * ignore cache
 * delete old

Enabling both of these is dangerous, as the cache may contain jobs
from a testing branch, which may not include current production jobs.

To avoid this situation, both params are removed. Ignore cache will
always be enabled, delete old will only be enabled when using
the master branch.

Issue: [RE-581](https://rpc-openstack.atlassian.net/browse/RE-581)


hughsaunders:
recheck_lint

PR #506, opened 2017-10-30
RE-1046 Bump openstack-ansible-ops

mattt416:
This is necessary to work around failures we're hitting as a result of
ansible 2.4.1.0 being installed.

Note that in addition to bumping the SHA, we also need to introduce a
new host var, `server_domain_name`.

Lastly, because openstack-ansible-ops now reverts to using ansible
2.3.2.0, we have to revert some of the python ansible code in
pipeline_steps/maas.groovy.  However, instead of reverting wholesale we
update it to work with both ansible < 2.4 and >= 2.4.

Issue: [RE-1046](https://rpc-openstack.atlassian.net/browse/RE-1046)

### DIFF
--- a/pipeline_steps/maas.groovy
+++ b/pipeline_steps/maas.groovy
@@ -2,13 +2,26 @@ String get_mnaio_entity_names() {
   String entities = sh (
     script: """#!/usr/bin/env python
 from ansible.parsing.dataloader import DataLoader
-from ansible.inventory.manager import InventoryManager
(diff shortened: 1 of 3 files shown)


mattt416:
Testing this [here](https://rpc.jenkins.cit.rackspace.net/job/OnMetal-Multi-Node-AIO_newton-xenial-deploy-none-periodic/129/).

PR #510, opened 2017-11-02
RE-1098 Don't create issues for timeout exceptions

hughsaunders:
Ideally timeouts would be within the node block so that clenaup
cannot fail due to timeout. However that is more complicated
when the node block is reused in different situations (eg
runonpubcloud).
As we are planning to move to node pool, its not worth investing
the time to refactor that, so drop timeout exceptions instead
of creating useless jira issues.

Issue: [RE-1098](https://rpc-openstack.atlassian.net/browse/RE-1098)


pipeline_steps/pubcloud.groovy
```diff
@@ -202,9 +202,20 @@ def runonpubcloud(Map args=[:], body){
...
-        common.create_jira_issue("RE", "Cleanup Failure: ${env.BUILD_TAG}")
+        // Ideally timeouts would be within the node block so that clenaup
+        // cannot fail due to timeout. However that is more complicated
+        // when the node block is reused in different situations (eg
+        // runonpubcloud).
+        // As we are planning to move to node pool, its not worth investing
+        // the time to refactor that, so drop timeout exceptions instead
+        // of creating useless jira issues.
+        if(e.getClass().name == "org.jenkinsci.plugins.workflow.steps.FlowInterruptedException"){
+          print "Caught timeout exception, dropping"
```
mattt416: Better to just catch this exception instead?

```
try {
  ...
} catch (org.jenkinsci.plugins.workflow.steps.FlowInterruptedException fie) {
  ...
} catch (e) {
  ...
}
```

hughsaunders: No idea what I was thinking doing that.. have fixed it, but it bought more complication.

mattt416: Ugh!

lint.sh
```diff
@@ -47,11 +47,8 @@ check_groovy(){
          return
        }
   extract_groovy_from_jjb
-  # pipeline_steps/*.groovy includes pipeline_steps/NonCPS.groovy, However
-  # NonCPS must be loaded before any scripts that use it, so it's explicitly
-  # included first.
```
mattt416: This removed comment is still applicable right?

hughsaunders: Not really as all the support scripts are now loaded before the test scripts.

pipeline_steps/pubcloud.groovy
```diff
@@ -201,6 +201,17 @@ def runonpubcloud(Map args=[:], body){
     if (fileExists("${args.inventory_path}/hosts")){
       try {
         delPubCloudSlave(args)
+      // catch timeouts
+      } catch (org.jenkinsci.plugins.workflow.steps.FlowInterruptedException e){
+          // Ideally timeouts would be within the node block so that clenaup
```
odyssey4me: nit: s/clenaup/cleanup/

hughsaunders: Done

PR #511, opened 2017-11-02
RE-1102 Remove nested parallelism from groovy

hughsaunders:
Nested parallelism causes unexpected behaviour and may cause
errors to go unreported. To work around this, the one usage of
nested parallelism is removed by flattening all the branches and
sub branches into a single parallel dictionary.

Issue: [RE-1102](https://rpc-openstack.atlassian.net/browse/RE-1102)

### DIFF
--- a/rpc_jobs/rpc_artifact_build.yml
+++ b/rpc_jobs/rpc_artifact_build.yml
@@ -141,7 +141,7 @@
           // We can run all the artifact build processes in parallel
           // for PR's because they do not upload anything.
           if (env.trigger == "pr") {{
-            parallel(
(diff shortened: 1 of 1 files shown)


hughsaunders:
The structure works..
```
Map ab = ["a" :{print "a"}, "b": {print "b"}]
Map cd = ["c" :{print "c"}, "d": {print "d"}]
Map branches = ab + cd + ["e": {print "e"}]
parallel branches
```
screen shot 2017-11-02 at 12 27 07

odyssey4me:
@hughsaunders Your jenkins-groovy-fu is strong.

PR #514, opened 2017-11-02
RE-678 Add merge trigger job for post merge job

hughsaunders:
Currently the post merge standard job is triggered by cron/timer only.
This commit adds a freestyle job which is triggered by pushes to github
which in turn triggers the post merge standard job.

The reason for this is that github push trigger is not reliable with
pipeline jobs (unless using orgfolder/multibranch pipeline which we
aren't).

Each project that wants to use a push trigger, will create a JJB
project to instantiate the job for their repo. Re-For-Projects
will be updated with instructions.

⚠️  Please also review the [documentation](https://rpc-openstack.atlassian.net/wiki/pages/diffpagesbyversion.action?pageId=19005457&selectedPageVersions=49&selectedPageVersions=47) when reviewing this PR.
- [x] doc review 1
- [ ] doc review 2 

Issue: [RE-678](https://rpc-openstack.atlassian.net/browse/RE-678)

### DIFF
--- a/rpc_jobs/standard_job_postmerge.yml
+++ b/rpc_jobs/standard_job_postmerge.yml
@@ -2,14 +2,36 @@
     name:       "gating-post-merge"
     repo_name:  "rpc-gating"
     repo_url:   "https://github.com/rcbops/rpc-gating"
-    # `branch` defaults to `master` and can be omitted.
-    #branch:     "master"
+    branch:     "master"
     image:      "xenial"
     scenario:   "functional"
     action:     "test"
     jira_project_key: "RE"
+    trigger_build: "PM_{repo_name}-{branch}-{image}-{scenario}-{action}"
     jobs:
       - 'PM_{repo_name}-{branch}-{image}-{scenario}-{action}'
+      - 'PM-push-trigger_{repo_name}-{branch}-{image}-{scenario}-{action}'
+
+# This job is specified as `freestyle` as using a pipeline job with github
+# trigger causes the job to be triggered irrespective of which branch was
+# pushed to.
+- job-template:
+    name: "PM-push-trigger_{repo_name}-{branch}-{image}-{scenario}-{action}"
+    project-type: freestyle
+    scm:
+      - git:
+          url: "{repo_url}"
+          branches:
+           - "{branch}"
+    properties:
+      - build-discarder:
+          num-to-keep: 30
+    triggers:
+        - github
+    builders:
+      - trigger-builds:
+        - project:
+            - "{trigger_build}"
 
 - job-template:
     name: 'PM_{repo_name}-{branch}-{image}-{scenario}-{action}'


hughsaunders:
Tested this by merging another PR:

- Merged https://github.com/rcbops/rpc-gating/pull/513 
- which triggered https://rpc.jenkins.cit.rackspace.net/job/PM-push-trigger_rpc-gating-master-xenial-functional-test/1/ 
- which triggered https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/PM_rpc-gating-master-xenial-functional-test/detail/PM_rpc-gating-master-xenial-functional-test/11/pipeline

Success ✅

hughsaunders:
Documentation still required, so don't close the issue when this is merged.

rpc_jobs/standard_job_postmerge.yml
```diff
@@ -2,14 +2,36 @@
...
+# trigger causes the job to be triggered irrespective of which branch was
+# pushed to.
+- job-template:
+    name: "PM-push-trigger_{repo_name}-{branch}-{image}-{scenario}-{action}"
+    project-type: freestyle
+    scm:
+      - git:
+          url: "{repo_url}"
+          branches:
+           - "{branch}"
```
odyssey4me: As with the standard PM job, can the default ``branch`` value be set to 'master' so that the setting can be omitted in the project (to simplify the usage of the template).

hughsaunders: The problem is that `branch` can't be omitted from the project when it's referenced in the `trigger_build` variable.

odyssey4me: Ah, best to note that in the docs then.

hughsaunders: Added [note to docs](https://rpc-openstack.atlassian.net/wiki/pages/diffpagesbyversion.action?pageId=19005457&selectedPageVersions=49&selectedPageVersions=48).

rpc_jobs/standard_job_postmerge.yml
```diff
@@ -2,14 +2,36 @@
...
+          num-to-keep: 30
+    triggers:
+        - github
+    builders:
+      - trigger-builds:
+        - project:
+            - "{trigger_build}"
 
 - job-template:
     name: 'PM_{repo_name}-{branch}-{image}-{scenario}-{action}'
```
git-harry: I think this name should be changed as well, e.g. `PM-periodic-trigger_`. Having one called `PM_` and the other `PM-push-trigger_` makes for an awkward naming scheme that lacks clarity when reading the list of jobs.

hughsaunders: I think the difference is justified as the push trigger requires a separate job, but the timer/periodic trigger is part of the main job. 

Otherwise the push trigger job will trigger the periodic trigger job, which doesn't make so much sense to me.

git-harry: That's a fair point. I'm not familiar enough with this to suggest an alternative but these types of naming inconsistency tend to build over time and create a whole load of unnecessary knowledge that must be learnt. If it can be removed I think that would be preferable.

mattt416: I'm not swayed enough one way or another to suggest further edits -- but I am now mindful of increasingly long job names and for that am happy to keep the job name as it stands.

PR #516, opened 2017-11-02
RE-1103 Update PM to use clone_with_pr_refs

mattt416:
Currently, the PM template is using the `git` pipeline step but this
isn't initialising submodules, which is necessary by rpc-openstack.
This commit switches to the helper `clone_with_pr_refs` instead.

Issue: [RE-1103](https://rpc-openstack.atlassian.net/browse/RE-1103)


mattt416:
Testing this change [here](https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-liberty-trusty-ceph-deploy/2/).

PR #517, opened 2017-11-03
RE-1111 Notify #ceph and #rpc-engineering of phobos deploys

hughsaunders:
This is so the storage team who are a current user of phobos
are aware of potential disruption.

Issue: [RE-1111](https://rpc-openstack.atlassian.net/browse/RE-1111)


odyssey4me:
Shouldn't we just notify ``#rpc-engineering`` - otherwise this list is going to continue growing.

hughsaunders:
I'll add rpc-engineering as that was requested previously, but @toan-nguyen specifically requested #ceph.

PR #518, opened 2017-11-03
RE-1114 Remove generator usage from cleanup script

hughsaunders:
Generators can only be iterated once, and in one case, we
were doing two membership tests on one generator, causing incorrect
results for the second test.

This commit removes all generator usage, even when it wouldn't
cause a problem, as the data set is small enough that we don't need
the optimisation. This may also prevent bugs being introduced in future.

Issue: [RE-1114](https://rpc-openstack.atlassian.net/browse/RE-1114)

### DIFF
--- a/scripts/periodic_cleanup.py
+++ b/scripts/periodic_cleanup.py
@@ -95,14 +95,14 @@ class Cleanup:
     def cache_servers(self):
         self.servers = self.cs.servers.list()
         self.servers_by_region[self.region] = self.servers
-        self.unprotected_servers = (
+        self.unprotected_servers = [
             server for server in self.servers
             if not re.match(self.protected_prefix, server.name)
-        )
+        ]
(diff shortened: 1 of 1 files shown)


scripts/periodic_cleanup.py
```diff
@@ -188,7 +188,7 @@ def cleanup_maas_entities(self):
 
             # sort agents with most recently connected agent last
             agents = sorted(
-                (a for a in self.agents if a.id == entity.agent_id),
+                [a for a in self.agents if a.id == entity.agent_id],
```
git-harry: This doesn't need to be changed, sorted only does one pass and the generator can't be accessed outside the function.

hughsaunders: Yeah, sorted returns a list, but I changed it for consistency.

PR #519, opened 2017-11-03
RE-508 Remove pyrax

git-harry:
In preparation for using Phobos, the use of pyrax and rax Ansible
modules needs to be replaced with the generic OpenStack versions. The
Python modules have conflicting dependencies and therefore cannot both
be installed and working in the gating venv. This change removes pyrax
from the list of requirements and adds shade to enable use of the os_*
Ansible modules. rackspaceauth is required to use the os_* modules with
Rackspace public cloud.

These changes do not, by themselves, allow for Phobos to be used by
rpc-gating. Use of Phobos will need to be added separately once the
nature of its use is decided.

common.groovy

`writeCloudsCfg` is added to create a configuration file compatible with
the new os_* Ansible modules, it is used instead of `writePyraxCfg` to
make the necessary credentials available. `writeRaxmonCfg` is for use
by the raxmon CLI tool.

allocate_pubcloud.yml

This change updates allocation_pubcloud.yml to use the community
OpenStack Ansible modules instead of Rackspace-specific ones.

The switch from `rax` to `os_server` is mostly a case of modifying
parameter names. The exceptions to that are:
- `os_server` requires the section of clouds.yaml to be specified with
  the parameter `cloud`
- `os_server` does not support building multiple instances. Instead of
  attempting to retain this functionality in the playbook, it has been
  removed give that `count` is always set to 1.

cleanup_playbook.yml

Use of `rax` is replaced with `os_server`. Given that creating multiple
instances is no longer supported by allocate_pubcloud.yml this playbook
has been simplified to remove a single instance.

create_cloud_image.py

The script rax_create_image.py has been renamed to create_cloud_image.py
given that in is no longer tied to Rackspace public cloud. It has been
mostly rewritten due to the differences between the pyrax and openstack
APIs however, the suggestion in the comment at the top of the file to
switch to the Ansible modules has not been followed because it would
change how the image is updated. Changes beyond the switch away from
pyrax are being left to future refactoring to simplify this patch.

To use the openstacksdk configuration file, the script required a couple
of changes.  The file location is now specified using the environment
variable `OS_CLIENT_CONFIG_FILE` and the cloud to select from the file
is set using the option `--cloudname`.

upload_to_swift.yml

The Ansible module os_object does not support setting the object to be
automatically deleted by Swift after a defined period of time. The
`swift` CLI tool is used to upload the archive, to enable this to be
configured.

The public object URL for CloudFiles is determined via the CDN API, this
is not something supported by the OpenStack Swift tools, and so there is
now additional code to get the correct URL based on whether or not
CloudFiles is being used.

maasutils.py

pyrax is replaced here with rackspace_monitoring, this is a relatively
trivial change given that pyrax itself uses rackspace_monitoring to
support the monitoring API.

multi_node_aio_maas_entities.yml and multi_node_aio_maas_cleanup.yml

To remove pyrax, the CLI tool raxmon is called to create and delete the
entities. Given there is no longer a single entry point to these two
actions the role maas_entities is removed and the required tasks are
added to the playbooks.

periodic_cleanup.py

This change replaces the usage of pyrax with openstacksdk.  At present
the script does not handle command-line arguments, so to keep changes at
a minimum we hard-code the clouds.yaml cloud to be 'public_cloud'.

Co-Authored-By: Matt Thompson <mattt@defunct.ca>

Issue: [RE-508](https://rpc-openstack.atlassian.net/browse/RE-508)

I've run this through some testing:
periodic_cleanup - https://rpc.jenkins.cit.rackspace.net/job/Periodic-Cleanup/5006
allocate_pubcloud/cleanup_pubcloud/upload_to_cloudfiles - https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_newton-trusty-minorupgrade-swift-periodic/72
save_pubcloud (Cloud-Image-Build) - isn't properly tested because the job is currently broken and is something being actively used.


mattt416:
Scanned through the change set and reviewed job logs and don't see anything out of the ordinary.  Will review the change set in more detail before reviewing.

playbooks/upload_to_swift.yml
```diff
@@ -0,0 +1,90 @@
...
+      os_object:
+        container: "{{ container }}"
+        container_access: "public"
+        region_name: "{{ region }}"
+        cloud: "{{ cloud_name }}"
+
+    - name: Authenticate to the cloud and retrieve the service catalog
+      os_auth:
+        cloud: "{{ cloud_name }}"
+        region_name: "{{ region }}"
```
mattt416: Any issues/downsides to having artifacts scattered across multiple regions?  I'm not aware of any (and will presumably make uploading faster since you're uploading from an instance in the same region), but just pointing it out incase anyone can think of any.

mattt416:
Testing an [MNAIO](https://rpc.jenkins.cit.rackspace.net/job/OnMetal-Multi-Node-AIO_liberty-trusty-deploy-none-periodic/80/) to validate that the MaaS bits still work as expected.

playbooks/upload_to_swift.yml
git-harry: The region is set to `"DFW"` at the bottom of this file so this change should keep things as they currently are.

mattt416: Ah ok, thought that was coming in from Jenkins.

mattt416:
The [MNAIO](https://rpc.jenkins.cit.rackspace.net/job/OnMetal-Multi-Node-AIO_liberty-trusty-deploy-none-periodic/80/) test failed, but seemed to create/delete MaaS entities without issue.

scripts/create_cloud_image.py
```diff
@@ -0,0 +1,95 @@
...
+            break
+
+    timeout = 900
+    while timeout > 0 and saved_image.status not in ("active", "error"):
+        timeout -= 60
+        sleep(60)
+        saved_image = conn.image.get_image(saved_image)
+
+    if saved_image.status == "active":
+        for image in current_images:
```
hughsaunders: Should this use current_ids instead, as it looks like this is setup to handle multiple images with the same name.

playbooks/upload_to_swift.yml
```diff
@@ -0,0 +1,90 @@
...
+    - set_fact:
+        rax_pub_cloud: "{{ 'clouddrive.com' in object_store['endpoints'][0]['publicURL'] }}"
+
+    - set_fact:
+        object_cdn: "{{ service_catalog | selectattr('type', 'equalto', 'rax:object-cdn') | first }}"
+      when: rax_pub_cloud
+
+    # os_object does not currently support setting object expiration header field
+    - name: Upload file to Cloud Files
+      command: "swift upload {{ container }} {{ archive_name }} --header 'X-Delete-After:2592000'"
```
hughsaunders: Swift client has been added to the venv ✅

playbooks/multi_node_aio_maas_cleanup.yml
```diff
@@ -2,5 +2,20 @@
...
+      failed_when: "{{ list_entities.stdout | regex_search('Total: [0-9]+$') == None }}"
+      delegate_to: localhost
+      register: list_entities
+      until: list_entities|success
+      retries: 3
+      delay: 5
+
+    - name: Delete host entities
+      command: "raxmon-entities-delete --id {{ list_entities.stdout | regex_search(item) | regex_search('en[a-zA-Z0-9]+') | quote }}"
+      with_items: "{{ ([] | zip_longest(entity_labels, fillvalue='id=en[a-zA-Z0-9]+ label=') | map('join', '') | list) }}"
```
hughsaunders: Could this be rewritten as follows? Or have I missed something?
```
regex_search('id=en[a-zA-Z0-9]+ label='+item)
with_items: "{{ entity_labels }}"
```

hughsaunders:
Kicked off another MNAIO https://rpc.jenkins.cit.rackspace.net/job/OnMetal-Multi-Node-AIO_liberty-trusty-deploy-none-periodic/83/ ▶️

scripts/create_cloud_image.py
git-harry: Either an image object or ID string is acceptable - https://developer.openstack.org/sdks/python/openstacksdk/users/proxies/image_v1.html#openstack.image.v1._proxy.Proxy.delete_image

hughsaunders: Ah yes, its an object that contains the id.

playbooks/multi_node_aio_maas_cleanup.yml
git-harry: I've made the following change:
```diff
diff --git a/playbooks/multi_node_aio_maas_cleanup.yml b/playbooks/multi_node_aio_maas_cleanup.yml
index 545b45c..aa86ad7 100644
--- a/playbooks/multi_node_aio_maas_cleanup.yml
+++ b/playbooks/multi_node_aio_maas_cleanup.yml
@@ -13,8 +13,8 @@
       delay: 5
 
     - name: Delete host entities
-      command: "raxmon-entities-delete --id {{ list_entities.stdout | regex_search(item) | regex_search('en[a-zA-Z0-9]+') | quote }}"
-      with_items: "{{ ([] | zip_longest(entity_labels, fillvalue='id=en[a-zA-Z0-9]+ label=') | map('join', '') | list) }}"
+      command: "raxmon-entities-delete --id {{ list_entities.stdout | regex_search('id=en[a-zA-Z0-9]+ label=' + item + ' ') | regex_search('en[a-zA-Z0-9]+') | quote }}"
+      with_items: "{{ entity_labels }}"
       register: delete_entities
       until: delete_entities|success
       retries: 3
```

hughsaunders:
https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-newton-xenial-swift-deploy/5/ ▶️

PR #520, opened 2017-11-03
RE-1116 Correctly fail build when failure occurs

mattt416:
Currently, builds that fail in the `Execute Run Script` stage do not
fail the build.  This commit sets the build to `FAILURE` in the catch
and re-throws the exception.  This is a similar pattern that exists in
standard_job_premerge.yml.

Issue: [RE-1116](https://rpc-openstack.atlassian.net/browse/RE-1116)

### DIFF
--- a/rpc_jobs/standard_job_postmerge.yml
+++ b/rpc_jobs/standard_job_postmerge.yml
@@ -122,10 +122,13 @@
                 }} // dir
               }} // ansiColor
             }} catch (e) {{
+              print(e)
+              currentBuild.result="FAILURE"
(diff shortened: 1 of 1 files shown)


mattt416:
Manually mod'd https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-mitaka-trusty-ceph-deploy with this change and added an explicit sh "exit 1" to verify that https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-mitaka-trusty-ceph-deploy/2/ would fail as expected.

PR #521, opened 2017-11-03
RE-1120 Unset delete old when building subset

hughsaunders:
JJB can be restricted to building a subset of jobs, this option
must not be used with --delete-old, as then jobs not in that
subset will be deleted.

This commit ensures --delete-old is not enabled when building
a subset of jjb jobs.

Issue: [RE-1120](https://rpc-openstack.atlassian.net/browse/RE-1120)


hughsaunders:
Tweaked commit message, but no code change since @mattt416's review.

PR #523, opened 2017-11-06
RE-703 Merge rpc_gating_lint with pre_merge_test

git-harry:
`rpc_gating_lint.yml` is a pre-merge job used by rpc-gating to run
`lint.sh`. This change removes `rpc_gating_lint.yml` and adds lint to
the list of scenarios covered by the rpc-gating pre-merge project.

Separate scripts for functional and lint tests have been created in
`gating/common/`. `gating/pre_merge_test/run` and
`gating/post_merge_test/run` have been updated to run the appropriate
script based on the scenario name, this reduces duplication and makes
them work in a consistent manner.

The comment in `gating/post_merge_test/run` was removed (not copied to
`gating/common/run_lint.sh`) because it appeared to be referencing the
implementation in `rpc_gating_lint.yml` which is removed by this change.

Issue: [RE-703](https://rpc-openstack.atlassian.net/browse/RE-703)

PR #525, opened 2017-11-07
RE-53 Add lint check to validate job name length

mattt416:
We've run into situations where long job names cause python scripts in
virtualenvs to fail.  This is due to a kernel limitation that imposes
the max number of characters that can exist in an interpreter line in a
script.

This commit adds a new check that scans all job name lengths and then
errors if any exceeds 77 chars.

Issue: [RE-53](https://rpc-openstack.atlassian.net/browse/RE-53)

### DIFF
--- a/lint.sh
+++ b/lint.sh
@@ -20,13 +20,16 @@ install(){
   pip install -c constraints.txt -r test-requirements.txt >/dev/null
 }
 
+create_jjb_ini(){
+  # work around for ip6 issues in docker image used for gating UG-652
+  echo -e "[jenkins]\nurl=http://127.0.0.1:8080" > lint_jjb.ini
+}
+
 check_jjb(){
   which jenkins-jobs >/dev/null \
     || { echo "jenkins-jobs unavailble, please install jenkins-job-builder from pip"
          return
        }
-  # work around for ip6 issues in docker image used for gating UG-652
-  echo -e "[jenkins]\nurl=http://127.0.0.1:8080" > lint_jjb.ini
   jenkins-jobs --conf lint_jjb.ini test -r rpc_jobs >/dev/null \
     && echo "JJB Syntax ok" \
     || { echo "JJB Syntax fail"; rc=1; }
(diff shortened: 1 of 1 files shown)


mattt416:
Modified this script to error on job names > 50 chars for testing, and you can see the results [here](https://rpc.jenkins.cit.rackspace.net/job/RPC-Gating-Lint/1900/console):

```
** Scanning for job names for those that are longer than 50 characters ... **
55	OnMetal-Multi-Node-AIO_kilo-trusty-deploy-none-periodic
57	OnMetal-Multi-Node-AIO_kilo-trusty-deploy-none-post-merge
58	OnMetal-Multi-Node-AIO_liberty-trusty-deploy-none-periodic
60	OnMetal-Multi-Node-AIO_liberty-trusty-deploy-none-post-merge
57	OnMetal-Multi-Node-AIO_mitaka-trusty-deploy-none-periodic
59	OnMetal-Multi-Node-AIO_mitaka-trusty-deploy-none-post-merge
72	OnMetal-Multi-Node-AIO_newton-kilo-trusty-leapfrogupgrade-large-periodic
73	OnMetal-Multi-Node-AIO_newton-kilo-trusty-leapfrogupgrade-medium-periodic
72	OnMetal-Multi-Node-AIO_newton-kilo-trusty-leapfrogupgrade-small-periodic
75	OnMetal-Multi-Node-AIO_newton-liberty-trusty-leapfrogupgrade-large-periodic
76	OnMetal-Multi-Node-AIO_newton-liberty-trusty-leapfrogupgrade-medium-periodic
75	OnMetal-Multi-Node-AIO_newton-liberty-trusty-leapfrogupgrade-small-periodic
57	OnMetal-Multi-Node-AIO_newton-trusty-deploy-none-periodic
59	OnMetal-Multi-Node-AIO_newton-trusty-deploy-none-post-merge
57	OnMetal-Multi-Node-AIO_newton-xenial-deploy-none-periodic
59	OnMetal-Multi-Node-AIO_newton-xenial-deploy-none-post-merge
73	OnMetal-Multi-Node-AIO_r14.2.0-kilo-trusty-leapfrogupgrade-large-periodic
74	OnMetal-Multi-Node-AIO_r14.2.0-kilo-trusty-leapfrogupgrade-medium-periodic
73	OnMetal-Multi-Node-AIO_r14.2.0-kilo-trusty-leapfrogupgrade-small-periodic
60	RPC-AIO_master-xenial-deploy-ironic-multihypervisor-periodic
54	RPC-AIO_master-xenial-deploy-ironic-multihypervisor-pr
57	RPC-AIO_newton-kilo-trusty-leapfrogupgrade-swift-periodic
60	RPC-AIO_newton-liberty-trusty-leapfrogupgrade-swift-periodic
60	RPC-AIO_newton-xenial-deploy-ironic-multihypervisor-periodic
54	RPC-AIO_newton-xenial-deploy-ironic-multihypervisor-pr
58	RPC-AIO_r14.2.0-kilo-trusty-leapfrogupgrade-swift-periodic
** 26 problematic job name(s) found! **
```

git-harry:
As it stands, the only issues I've seen with this relate to the use of Python virtual environments. While very long job names will be difficult to read, I wouldn't want us to be forcing the awkward naming of jobs just to meet this limit. Instead of creating the venv in the workspace, could we use /tmp instead?

odyssey4me:
@git-harry That would only be helpful for the rpc-gating venv(s) - what about venvs created in the test itself?

FYI this issue is informed by https://github.com/pypa/pip/issues/1773 and https://github.com/pypa/virtualenv/issues/596.

git-harry:
@odyssey4me you're assuming any venv created by the test is in the workspace, what about venvs created in subdirectories of the workspace as happens here https://github.com/rcbops/rpc-gating/pull/523/files#diff-6ab1426463e7a9fa99ef939e5f30246c

This change isn't guaranteed to protect against the issue in all cases and the subset of cases for which it does guarantee to solve the problem can be achieved in a different manner without placing a restriction on the job name length.

odyssey4me:
@git-harry It's not foolproof, no, but this still looks like something to at least try to prevent new jobs going in until we can figure out a better way to solve this. Ideally, in my mind, we should be able to have Jenkins use some sort of short-form directory name for the workspace. This would be ideal as it will clean up after itself so that the jobs can still be universally used on single-use-slaves and long-lived-slaves. If we do workarounds like using temp directories outside of the workspace, then we have to make sure that the jobs clean up after themselves otherwise long-lived-slaves get cluttered or fail jobs due to clashes when two jobs use the same spaces (we've seen this in the apt artifact builder).

hughsaunders:
All nodes where a venv is used are allocated via common.use_node, so we can modify that function to allocate a workspace with a shorter path (using the ws step) and ensure it is cleaned up. However we'd have to work out where to put it, so the jenkins user has all the required perms. Tmp may not be the best especially if it is tmpfs?

odyssey4me:
@hughsaunders Would it perhaps be possible for the workspace directory name to be a fixed size hash of the job name instead of the full job name string? That way we could have the workspace folder just be ``/var/lib/jenkins/<16 or perhaps 24 char hash>`` and move on with our lives.

mattt416:
I'm not married to this idea -- I went with the job name length checking idea as that was what was suggested by @git-harry in RE-53.  :) 

As we're moving away from CIT Jenkins (where /var/lib/jenkins is necessary because that's the partition with the most available space), we can easily move away from /var/lib/jenkins/workspace to another directory higher up the tree.  To be honest though, /var/lib/jenkins/workspace only accounts for 17% of the overall max job name length.

hughsaunders:
In answer to @odyssey4me's question, we could use a hash, and a 8 char hash would prob do.

PR #528, opened 2017-11-09
RE-912 Remove majorupgrade from matrix

mattt416:
At present, we only test a major upgrade from liberty -> mitaka and
according to @claco this is not an upgrade that support would do going
forward.  As such, this commit removes all reference to majorupgrade in
rpc_aio.yml.

Issue: [RE-912](https://rpc-openstack.atlassian.net/browse/RE-912)

### DIFF
--- a/rpc_jobs/rpc_aio.yml
+++ b/rpc_jobs/rpc_aio.yml
@@ -12,7 +12,6 @@
       - mitaka:
           branch: mitaka
           branches: "mitaka.*"
-          UPGRADE_FROM_REF: "liberty"
       - newton:
           branch: newton
           branches: "newton.*"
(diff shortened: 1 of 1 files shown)


mattt416:
A diff of jobs before and after this change:

$ diff ~/master.jjb ~/912.jjb
70,71d69
< RPC-AIO_mitaka-trusty-majorupgrade-ceph-periodic
< RPC-AIO_mitaka-trusty-majorupgrade-swift-periodic
$

odyssey4me:
Given that the tests are being reduced for RPC-O, I'd prefer at least @antonym or @claco marking their approval for this review.

mattt416:

PR #529, opened 2017-11-10
RE-1139 Fix retry instance creation

git-harry:
The task "Provision a cloud instance" fails without trying the next
region under certain types of build failure, for example if the
requested image does not exist in the current region.

This change updates the when clause so that the next region will be
tried if no 'server' key is present in the result object.

Issue: [RE-1139](https://rpc-openstack.atlassian.net/browse/RE-1139)

### DIFF
--- a/playbooks/allocate_pubcloud.yml
+++ b/playbooks/allocate_pubcloud.yml
@@ -52,7 +52,7 @@
       register: result
       with_items: "{{ regions_shuf + fallback_regions_shuff }}"
       when:
-        - result is undefined or result.server.status != 'ACTIVE'
+        - result is undefined or 'server' not in result or result.server.status != 'ACTIVE'
       failed_when: false
 
     - name: Extract instance boot result from loop results array


git-harry:
Without this change https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-newton-trusty-swift-deploy/9:
```
TASK [Provision a cloud instance] **********************************************
ok: [localhost] => (item=ORD)
fatal: [localhost]: FAILED! => {"failed": true, "msg": "The conditional check 'result is undefined or result.server.status != 'ACTIVE'' failed. The error was: error while evaluating conditional (result is undefined or result.server.status != 'ACTIVE'): 'dict object' has no attribute 'server'\n\nThe error appears to have been in '/var/lib/jenkins/workspace/PM_rpc-openstack-newton-trusty-swift-deploy/rpc-gating/playbooks/allocate_pubcloud.yml': line 37, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n      # When is used to skip further attempts once instances have booted.\n    - name: Provision a cloud instance\n      ^ here\n"}
```

With this change https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-newton-trusty-swift-deploy/11:
```
TASK [Provision a cloud instance] **********************************************
ok: [localhost] => (item=ORD)
changed: [localhost] => (item=DFW)
```

PR #532, opened 2017-11-13
RLM-352 Bump timeout to 6 hours

antonym:
Leap PM jobs seem to run around 4 hours so the PR
jobs should be similar.  Adding an extra hour for
buffer.

Issue: [RLM-352](https://rpc-openstack.atlassian.net/browse/RLM-352)


mattt416:
The timeout in standard_job_postmerge.yml is 6 hours, shall we just set this timeout to the same for consistency's sake?

antonym:
Yeah, I'm fine with that... wasn't sure how necessary the 3 hours was.  If we ever needed lower, maybe we make that configurable later on?

PR #535, opened 2017-11-14
RE-1142 Add python-openstacksdk to requirements

git-harry:
python-openstacksdk is required by the periodic clean-up job. It was
being installed as a dependency of python-ironicclient which is itself a
dependency of shade. python-ironicclient has now removed this dependency
causing jobs to fail. python-openstacksdk should be an explicit
requirement of this project and this commit makes that so.

Issue: [RE-1142](https://rpc-openstack.atlassian.net/browse/RE-1142)

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -41,6 +41,7 @@ netaddr==0.7.19
 netifaces==0.10.5
 nose==1.3.7
 oauthlib==2.0.3
+openstacksdk==0.9.19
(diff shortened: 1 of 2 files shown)


git-harry:
recheck_cit_xenial_lint_test

git-harry:
Successful test build: https://rpc.jenkins.cit.rackspace.net/job/Periodic-Cleanup/5275

PR #538, opened 2017-11-15
RE-913 Add xenial minor upgrade to rpc_upgrades

git-harry:
This commit adds an additional minor upgrade job to rpc_upgrades so that
newton minor upgrades are tested on both trusty and xenial.

The rpc_aio code for testing minor upgrades is removed given this
testing is now the responsibility of rpc_upgrades.

Issue: [RE-913](https://rpc-openstack.atlassian.net/browse/RE-913)


git-harry:
Yep, it only matters when it's `"mnaio"`.

PR #540, opened 2017-11-15
RE-774 Project specific creds for standard jobs

hughsaunders:
This collection of commits allows a project to specify the credentials that their jobs require via JJB. RPC-Repo creds are added to octavia as they need them and also as an example. 

Issue: [RE-774](https://rpc-openstack.atlassian.net/browse/RE-774)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -684,4 +684,102 @@ void internal_slave(body){
   use_node("CentOS", body)
 }
 
+// Build an array suitable for passing to withCredentials
+// from a space or comma separated list of credential IDs.
+@NonCPS
+List build_creds_array(String list_of_cred_ids){
+    print("Building credentials array from the following list of IDs: ${list_of_cred_ids}")
+    Map creds_bundles = [
+      "cloud_creds": ['dev_pubcloud_username',
+                      'dev_pubcloud_api_key',
+                      'dev_pubcloud_tenant_id'],
+      "rpc_repo": ['RPC_REPO_IP',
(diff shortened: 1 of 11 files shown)


pipeline_steps/common.groovy
```diff
@@ -684,4 +684,102 @@ void internal_slave(body){
...
+// Build an array suitable for passing to withCredentials
+// from a space or comma separated list of credential IDs.
+@NonCPS
+List build_creds_array(String list_of_cred_ids){
+    print("Building credentials array from the following list of IDs: ${list_of_cred_ids}")
+    Map creds_bundles = [
+      "cloud_creds": ['dev_pubcloud_username',
+                      'dev_pubcloud_api_key',
+                      'dev_pubcloud_tenant_id'],
+      "rpc_repo": ['RPC_REPO_IP',
```
hughsaunders: This is inconsistent, cloud_creds is using the id, while rpc_repo creds are using the variable name. I think using the variable would be better for users, will change all to that.

hughsaunders: On second thoughts, using variable names instead of IPs may lead to problems with credentials that provide multiple variables. Also All the above are in fact IDs, its just inconsistent capitalisation that makes it look like one set are IDs and the other are variables.

hughsaunders:
Example run of the with requested creds test job: https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/RE-unit-test-withRequestedCredentials/detail/RE-unit-test-withRequestedCredentials/9/pipeline ✅

PR #541, opened 2017-11-15
RLM-366 Reduces PR tests for rpc-upgrades

antonym:
Reduces PR tests to kilo-newton and liberty-newton.
This provides a basic sanity check and then relies
on the PM jobs to surface issues.  Because the jobs
take several hours, it's currently overkill to test
every version on PR.

Also makes image names consistent and switches to
16.04 for mnaio host.

Issue: [RLM-366](https://rpc-openstack.atlassian.net/browse/RLM-366)

### DIFF
--- a/rpc_jobs/rpc_upgrades.yml
+++ b/rpc_jobs/rpc_upgrades.yml
@@ -1,11 +1,27 @@
 ### tests rpc-upgrades to <scenario> leaped from <action>
 - project:
-    name: "rpc-upgrades-aio-newton-head"
+    name: "rpc-upgrades-aio-newton-head-pr"
     repo_name: "rpc-upgrades"
     repo_url: "http://github.com/rcbops/rpc-upgrades"
-    series: "master"
     branches:
      - "master"
+    image:
+      - trusty_aio:
+          FLAVOR: "performance2-15"
+          IMAGE: "Ubuntu 14.04.5 LTS prepared for RPC deployment"
+    scenario:
+      - "swift"
+    action:
+      - "liberty_to_newton_leap"
+      - "kilo_to_newton_leap"
+    jira_project_key: "RLM"
+    jobs:
+      - 'PR_{repo_name}-{series}-{image}-{scenario}-{action}'
+
+- project:
+    name: "rpc-upgrades-aio-newton-head-pm"
+    repo_name: "rpc-upgrades"
+    repo_url: "http://github.com/rcbops/rpc-upgrades"
     image:
       - trusty_aio:
           FLAVOR: "performance2-15"
@@ -25,7 +41,6 @@
       - "r14.4.1_to_newton_minor"
     jira_project_key: "RLM"
     jobs:
-      - 'PR_{repo_name}-{series}-{image}-{scenario}-{action}'
       - 'PM_{repo_name}-{branch}-{image}-{scenario}-{action}'
     exclude:
       - image: "xenial_aio"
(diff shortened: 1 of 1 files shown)


antonym:
recheck_cit_all

rpc_jobs/rpc_upgrades.yml
```diff
@@ -63,19 +82,18 @@
...
 - project:
-    name: "rpc-upgrades-mnaio-newton-head"
+    name: "rpc-upgrades-mnaio-newton-head-pm"
     repo_name: "rpc-upgrades"
     repo_url: "http://github.com/rcbops/rpc-upgrades"
     branches:
      - "master"
     image:
-      - mnaio:
+      - xenial_mnaio:
```
odyssey4me: Are you using ``xenial`` here to denote the mnaio host, or the guest OS that it's running? How're you informing the MNAIO setup to deploy one distro versus another. I would think it might be best to use this to denote the VM OS so that it's similar in connotation to the AIO setups. That way you can easily inform the MNAIO setup tooling which VM OS to prep too.

antonym: Good point... I'll change it to reflect the guest OS which will give me one more variable to use.

rpc_jobs/rpc_upgrades.yml
```diff
@@ -63,19 +82,18 @@
...
     repo_name: "rpc-upgrades"
     repo_url: "http://github.com/rcbops/rpc-upgrades"
     branches:
      - "master"
     image:
-      - mnaio:
+      - trusty_mnaio:
           FLAVOR: "onmetal-io1"
-          IMAGE: "OnMetal - Ubuntu 14.04 LTS (Trusty Tahr)"
+          IMAGE: "OnMetal - Ubuntu 16.04 LTS (Xenial Xerus)"
```
odyssey4me: It's probably worth adding a comment here to explain why the IMAGE value does not match the image axis name, otherwise both reviewers and later job changers are going to be very confused.

rpc_jobs/rpc_upgrades.yml
```diff
@@ -1,6 +1,26 @@
...
+      - "kilo_to_newton_leap"
+    jira_project_key: "RLM"
+    jobs:
+      - 'PR_{repo_name}-{series}-{image}-{scenario}-{action}'
+
+- project:
+    name: "rpc-upgrades-aio-newton-head-pm"
     repo_name: "rpc-upgrades"
     repo_url: "http://github.com/rcbops/rpc-upgrades"
     series: "master"
```
mattt416: Nit, series and branches can be removed from the PM jobs.

rpc_jobs/rpc_upgrades.yml
```diff
@@ -42,14 +61,14 @@
         action: "kilo_to_newton_leap"
 
 - project:
-    name: "rpc-upgrades-aio-newton-release"
+    name: "rpc-upgrades-aio-newton-release-pm"
     repo_name: "rpc-upgrades"
     repo_url: "http://github.com/rcbops/rpc-upgrades"
     series: "master"
```
mattt416: Same as above, series/branches not needed for PM.

rpc_jobs/rpc_upgrades.yml
```diff
@@ -63,19 +82,19 @@
...
     jobs:
-      - 'PR_{repo_name}-{series}-{image}-{scenario}-{action}'
       - 'PM_{repo_name}-{branch}-{image}-{scenario}-{action}'
 
 - project:
-    name: "rpc-upgrades-mnaio-newton-head"
+    name: "rpc-upgrades-mnaio-newton-head-pm"
     repo_name: "rpc-upgrades"
     repo_url: "http://github.com/rcbops/rpc-upgrades"
     branches:
```
mattt416: Same as above, branches not needed for PM.

rpc_jobs/rpc_upgrades.yml
antonym: Reason I did that is because then I collide on the two first projects since I split the first two out to avoid having to make a ton of excludes.

antonym:
Cleaned up the series and branches as I'm using the default master.

PR #542, opened 2017-11-16
RE-1129 Implement nodepool setup job

odyssey4me:
Currently most RPC tests use single-use slaves which are requested,
built, used and deleted within the test time of every job. This adds
a minimum of 10 minutes of test time to every job and often results
in failures. This mechanism also adds a bunch of management code and
administration which we can displace quite easily by using nodepool.

Nodepool is what OpenStack uses to prepare images and cloud instances
to test on. It handles the preparation and teardown so that the tests
only have to request and release the nodes.

This patch adds a job to do the build of nodepool hosts to manage the
system. The intent for this job is to continuously deliver the software
and configuration to the instances running as nodepool servers.

For now this only handles the initial setup. Follow on work will add
the jenkins enablement, monitoring and the image preparation.

Included in this patch are also the following changes:

* A new groovy function is added to use the rpc-gating venv to
  execute ansible-galaxy to download the contents of a standard
  ansible-role-requirements.yml file.

* The clouds.yaml configuration implemented is changed to include
  options to allow the file to be used along with the Ansible
  OpenStack dynamic inventory. The regions are also specified
  in clouds.yaml as that is required when using the dynamic
  inventory.

* The drop_ssh_auth_keys playbook is adjusted to allow it to target
  a given set of hosts. The default remains 'job_nodes' to ensure
  that it does not break anything else using it. The concatenation
  for the file lookup is adjusted to ensure that it does a string
  concatenation - this is just good practise.

* The openstack dynamic inventory script for Ansible 2.3.2.0-1
  is vendored into the git tree as it is the most reliable way
  of accessing it to use it. As the script is a contrib script,
  it is unfortunately not included in the pip package for ansible.

* python-novaclient is added to requirements.txt in order to
  support the os_server_actions ansible module. The entry for
  python-swiftclient is moved so that the file's entries are
  sorted alphabetically.

* The slave_open_ports.yml playbook has been consolidated into
  the slave_security.yml playbook as that was the only place it
  was called from and there is no need for it to be a seperate
  playbook.

* The slave_security playbook is adjusted to allow it to target
  a given set of hosts. The default remains 'job_nodes' to ensure
  that it does not break anything else using it. It now also
  implements sshd configuration to improve the security posture
  for any instances it executes against.

Issue: [RE-1129](https://rpc-openstack.atlassian.net/browse/RE-1129)


playbooks/setup-nodepool.yml
```diff
@@ -0,0 +1,230 @@
...
+      file:
+        path: "/var/lib/nodepool/.config/openstack"
+        owner: "nodepool"
+        group: "nodepool"
+        mode: "0700"
+        state: directory
+
+    - name: Copy clouds.yaml
+      copy:
+        src: "~/.config/openstack/clouds.yaml"
```
odyssey4me: wrong location

playbooks/setup-nodepool.yml
```diff
@@ -0,0 +1,230 @@
...
+
+    - name: Install prerequisite packages
+      apt:
+        name:
+          - gcc
+          - git
+          - fail2ban
+          - libffi-dev
+          - libssl-dev
+          - lvm2
```
odyssey4me: no disk management is being done, so no lvm required

playbooks/setup-nodepool.yml
```diff
@@ -0,0 +1,230 @@
...
+      file:
+        path: "/home/nodepool/.ssh"
+        owner: "nodepool"
+        group: "nodepool"
+        mode: "0700"
+        state: directory
+
+    - name: Copy private key
+      copy:
+        src: "~/.ssh/id_rsa"
```
odyssey4me: env.ANSIBLE_PRIVATE_KEY_FILE

playbooks/setup-nodepool.yml
mattt416: Yep, AFAIK writeCloudsCfg() won't put clouds.yaml in ~.

playbooks/setup-nodepool.yml
```diff
@@ -0,0 +1,230 @@
...
+    - name: Install prerequisite packages
+      apt:
+        name:
+          - gcc
+          - git
+          - fail2ban
+          - libffi-dev
+          - libssl-dev
+          - lvm2
+          - openjdk-8-jre-headless
```
mattt416: Is this needed by nodepool?

mattt416:
Looks good as a first pass -- two things I think need addressing from the get-go is configuration off sshd_config and configuration of unattended-upgrades (both handled by ansible roles for slave setup).

playbooks/setup-nodepool.yml
odyssey4me: It's needed by zookeeper.

playbooks/setup-nodepool.yml
```diff
@@ -0,0 +1,272 @@
...
+          - daily
+          - notifempty
+      - name: nodepool-launcher
+        log: /var/log/nodepool/launcher-debug.log /var/log/nodepool/nodepool-launcher.log
+        options:
+          - compress
+          - missingok
+          - rotate 7
+          - daily
+          - notifempty
```
odyssey4me: We may need a zookeeper log rotation setup too.

playbooks/setup-nodepool.yml
```diff
@@ -0,0 +1,277 @@
...
+          - python-minimal
+          # requirements for security
+          - fail2ban
+          - ufw
+        update_cache: yes
+
+    - name: Install prerequisite pip packages
+      pip:
+        name:
+          - rackspaceauth
```
odyssey4me: We should probably notify the services to restart if something changes here.

playbooks/setup-nodepool.yml
```diff
@@ -0,0 +1,277 @@
...
+          - python-virtualenv
+          # requirement for zookeeper
+          - openjdk-8-jre-headless
+          # requirements for ansible
+          - python-apt
+          - python-minimal
+          # requirements for security
+          - fail2ban
+          - ufw
+        update_cache: yes
```
odyssey4me: We should probably notify the services to restart if something changes here.

playbooks/setup-nodepool.yml
```diff
@@ -0,0 +1,277 @@
...
+
+    - name: Add vhd-util PPA
+      apt_repository:
+        filename: "vhd-util"
+        repo: "ppa:openstack-ci-core/vhd-util"
+        update_cache: yes
+
+    - name: Install vhd-util
+      apt:
+        name: "vhd-util"
```
odyssey4me: We should probably notify the services to restart if something changes here.

playbooks/setup-nodepool.yml
```diff
@@ -0,0 +1,277 @@
...
+    # used by openstack-infra which has the modified version available.
+    # https://launchpad.net/~openstack-ci-core/+archive/ubuntu/vhd-util
+    # built from: https://github.com/emonty/vhd-util
+    # deployed by: https://github.com/openstack-infra/puppet-diskimage_builder/blob/339340409823927bb987f0195c6cedfdace05f4a/manifests/init.pp#L26
+
+    - name: Add vhd-util PPA
+      apt_repository:
+        filename: "vhd-util"
+        repo: "ppa:openstack-ci-core/vhd-util"
+        update_cache: yes
```
odyssey4me: We should probably notify the services to restart if something changes here.

playbooks/setup-nodepool.yml
```diff
@@ -0,0 +1,277 @@
...
+          ff00::0 ip6-mcastprefix
+          ff02::1 ip6-allnodes
+          ff02::2 ip6-allrouters
+          ff02::3 ip6-allhosts
+          # zookeeper hosts
+          {% for host in zookeeper_hosts %}
+          {{ hostvars[host].ansible_default_ipv4.address }} {{ host }}
+          {{ hostvars[host].ansible_default_ipv6.address }} {{ host }}
+          {% endfor %}
+        dest: "/etc/hosts"
```
odyssey4me: We should probably notify the services to restart if something changes here.

playbooks/setup-nodepool.yml
```diff
@@ -0,0 +1,277 @@
...
+      ufw:
+        rule: allow
+        from_ip: "{{ hostvars[item].ansible_default_ipv4.address }}"
+      with_items: "{{ zookeeper_hosts }}"
+
+    - name: Configure firewall to allow cluster traffic (ipv6)
+      ufw:
+        rule: allow
+        from_ip: "{{ hostvars[item].ansible_default_ipv6.address }}"
+      with_items: "{{ zookeeper_hosts }}"
```
odyssey4me: We should probably notify the services to restart if something changes here.

playbooks/setup-nodepool.yml
```diff
@@ -0,0 +1,277 @@
...
+    - name: Configure firewall to allow ssh
+      ufw:
+        rule: allow
+        name: OpenSSH
+
+    - name: Configure firewall to allow cluster traffic (ipv4)
+      ufw:
+        rule: allow
+        from_ip: "{{ hostvars[item].ansible_default_ipv4.address }}"
+      with_items: "{{ zookeeper_hosts }}"
```
odyssey4me: We should probably notify the services to restart if something changes here.

rpc_jobs/setup_nodepool.yml
```diff
@@ -0,0 +1,78 @@
+- job:
+    name: Setup-Nodepool
+    project-type: pipeline
+    concurrent: false
+    properties:
+      - build-discarder:
+          num-to-keep: 30
+    parameters:
+      - rpc_gating_params
```
odyssey4me: The job should run at least daily, if not more frequently.

playbooks/setup-nodepool.yml
odyssey4me: Done

playbooks/setup-nodepool.yml
odyssey4me: Done

playbooks/setup-nodepool.yml
odyssey4me: Done

playbooks/setup-nodepool.yml
odyssey4me: Done

playbooks/setup-nodepool.yml
odyssey4me: Done

playbooks/setup-nodepool.yml
odyssey4me: Done

playbooks/setup-nodepool.yml
odyssey4me: Done

rpc_jobs/setup_nodepool.yml
odyssey4me: Done

odyssey4me:
recheck_cit_xenial_lint_test

playbooks/setup_openstack_instances.yml
```diff
@@ -0,0 +1,73 @@
...
+        userdata: "{{ instance['userdata'] | default(omit) }}"
+        config_drive: yes
+        meta: "{{ default_metadata | combine(instance['metadata'] | default({})) }}"
+        wait: yes
+        timeout: 900
+      with_items: "{{ instance_list | default([]) }}"
+      loop_control:
+        loop_var: instance
+
+    - name: Lock instances to prevent accidental deletion
```
mattt416: Can we move the lock/unlock actions into separate plays?  I want to use this setup_openstack_instances.yml playbook, but I can't attach my volumes afterwards as the instances are locked.  If we separate these actions out and include them here, I can just re-include them before and after I attach volumes.

odyssey4me: Locking prevents changes? Oh - that's news to me.

I think it's a good idea to add the volume stuff as part of the same data structure and include the cloud actions in the same play. That way this play always unlocks, then does things to instances, then locks them again. It keeps it tight and does a fair job of trying to prevent humans from meddling through any means other than a patch being merged.

mattt416: OK, I'll adjust this playbook to handle volumes also once this PR merges.

odyssey4me:
FYI, this has successfully run in https://rpc.jenkins.cit.rackspace.net/job/Setup-Nodepool/9

playbooks/slave_security.yml
```diff
@@ -1,19 +1,25 @@
...
-  hosts: job_nodes
+  hosts: "{{ target_hosts | default('job_nodes') }}"
   user: root
   vars:
     ip_address_whitelist: "{{ lookup('env', 'SSH_IP_ADDRESS_WHITELIST') }}"
-  pre_tasks:
-    - name: Install apt packages
+    ports_whitelist: "{{ lookup('env', 'FIREWALL_OPEN_PORTS') }}"
+  tasks:
+
```
mattt416: Nit: unnecessary newline

playbooks/slave_security.yml
```diff
@@ -28,12 +34,27 @@
...
-    - role: jnv.unattended-upgrades
+    - include_role:
+        name: willshersystems.sshd
+      vars:
+        sshd:
+          PasswordAuthentication: "no"
+          X11Forwarding: "no"
+          PrintLastLog: "no"
+          GatewayPorts: "yes"
+          Compression: "yes"
```
mattt416: If I'm reading this correctly, some instances like long running slaves will have the role included here, and then again in setup_jenkins_slave.yml (with a marginally different config).  Wonder if we can remove the inclusion of the role in setup_jenkins_slave.yml and then any sensitive instance just needs to ensure they're including slave_security.yml.

rpc_jobs/setup_nodepool.yml
```diff
@@ -0,0 +1,109 @@
...
+                  args: [
+                    "-v",
+                    "-e '@${env.WORKSPACE}/rpc-gating/nodepool/nodepool-instances.yml'"
+                  ]
+                ) //venvPlaybook
+
+                // We need an inventory refresh, so we execute
+                // this using a fresh shell instead of including
+                // the playbooks in the previous ansible-playbook
+                // play list.
```
mattt416: Note, I'll be adding a refresh inside setup_openstack_instances.yml before we unlock instances (as we need to first ensure the instance exists).  We'll probably want to combine these again to avoid the extra inventory lookup.

playbooks/slave_security.yml
odyssey4me: I think we should register a card to figure out the best approach for the jenkins slave thing. Right now the two are separate and we'll need to flip the long running slave setup to use this mechanism IMO, and in so doing also figure things out like how we best handle their security, the ports, etc. I think this might be easy enough to handle with group_vars.

For this PR I''ll add another commit which ensures that the two role executions have the same results.

playbooks/slave_security.yml
odyssey4me: It was intentionally added to make it a little easier to read. I find that when a bunch of text (ie the task and the vars above) have no whitespace between them it gets harder to read it clearly.

playbooks/slave_security.yml
mattt416: OK ... in retrospect I don't think removing the ssh role from setup_jenkins_slave.yml would be a good idea since we don't want to allow password logins into our single use slaves since some of those create artifacts.    We also can't apply slave_security.yml to the single use slaves since the fw config, etc. will affect RPCO AIO deploys.

playbooks/setup_nodepool.yml
```diff
@@ -0,0 +1,244 @@
...
+
+  tasks:
+    - name: Wait for a successful connection
+      wait_for_connection:
+        connect_timeout: 2
+        delay: 5
+        sleep: 5
+        timeout: 60
+
+    - name: Gather facts
```
mattt416: Do you need the `Gather facts` and `Wait for a successful connection` tasks here since they're both run above?

odyssey4me: Done.

mattt416:
Looks like https://rpc.jenkins.cit.rackspace.net/job/RE-unit-test-withRequestedCredentials/97/console failed with:

```
ERROR: Could not resolve origin/pr/542/merge
hudson.plugins.git.GitException: Command "git rev-parse origin/pr/542/merge^{commit}" returned status code 128:
stdout: origin/pr/542/merge^{commit}

stderr: fatal: ambiguous argument 'origin/pr/542/merge^{commit}': unknown revision or path not in the working tree.
```

odyssey4me:
@mattt416 I'm under the impression that the unit tests are failing generally, not just in this PR. I think @hughsaunders is working on something to fix them?

scripts/ansible_v2_3_2_0_1_contrib_inventory_openstack.py
```diff
@@ -0,0 +1,246 @@
+#!/usr/bin/env python
```
git-harry: Copying this in isn't particularly nice. It feels like tech debt. I don't think it stops the pull request from being merged but I think there should be a separate issue created to address how this is managed.

requirements.txt
```diff
@@ -11,6 +11,8 @@ jmespath
 openstacksdk
 packaging
 python-dateutil
+python-novaclient
```
git-harry: Based on the documentation and module source code, os_server_actions should only require shade.

https://docs.ansible.com/ansible/2.3/os_server_actions_module.html
https://github.com/ansible/ansible/blob/v2.3.2.0-1/lib/ansible/modules/cloud/openstack/os_server_actions.py

playbooks/setup_openstack_instances.yml
```diff
@@ -0,0 +1,73 @@
...
+        userdata: "{{ instance['userdata'] | default(omit) }}"
+        config_drive: yes
+        meta: "{{ default_metadata | combine(instance['metadata'] | default({})) }}"
+        wait: yes
+        timeout: 900
+      with_items: "{{ instance_list | default([]) }}"
+      loop_control:
+        loop_var: instance
+
+    - name: Lock instances to prevent changes
```
git-harry: Maybe a block should be used to ensure this task is always run.

playbooks/slave_security.yml
hughsaunders: Currently infrastructure instances are registered as Jenkins slaves, not for running normal jobs, but for targeting for maintenance etc. That doesn't seem to be the case here. Is the idea that we will use dynamic inventory instead of Jenkins for connecting to infrastructure nodes? If so why have a different mechanism for infrastructure nodes and short lived nodes?

nodepool/nodepool-config.yml
```diff
@@ -0,0 +1,89 @@
...
+      - name: ubuntu-xenial
+        config-drive: true
+    pools: &provider_pools_block
+      - name: main
+        max-servers: 2
+        availability-zones: []
+        labels:
+          - name: ubuntu-xenial
+            diskimage: ubuntu-xenial
+            flavor-name: general1-8
```
hughsaunders: How do I overwrite this? Do I have to create another pool with a different flavour?

scripts/ansible_v2_3_2_0_1_contrib_inventory_openstack.py
hughsaunders: Can you suggest a better alternative?

nodepool/nodepool-config.yml
```diff
@@ -0,0 +1,89 @@
...
+      - openssh-server
+      - simple-init
+      - vm
+    release: xenial
+    env-vars:
+      TMPDIR: /opt/nodepool/dib_tmp
+      DIB_CHECKSUM: '1'
+      DIB_IMAGE_CACHE: /opt/nodepool/dib_cache
+      DIB_GRUB_TIMEOUT: '0'
+      DIB_DISTRIBUTION_MIRROR: 'http://mirror.rackspace.com/ubuntu'
```
git-harry: https?

requirements.txt
mattt416: Without python-novaclient installed, using os_server_actions to lock/unlock throws this:

```
UserWarning: Using shade to get a novaclient object is deprecated. If you need a novaclient object, please use make_legacy_client in os-client-config instead
```

pipeline_steps/common.groovy
```diff
@@ -381,6 +394,23 @@ def writeCloudsCfg(Map args){
...
+    # The default regions include LON which is not
+    # in the same catalog, causing errors when
+    # using the ansible dynamic inventory due to
+    # missing endpoints. We therefore specify all
+    # the regions in the US catalog.
+    regions:
+      - IAD
+      - DFW
+      - ORD
+      - HKG
```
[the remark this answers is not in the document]
mattt416: https://github.com/rcbops/rpc-gating/commit/bdb59a65524b7655cb80776840ced6f5a4364077 removed SYD/HKG from being used by standard tests, having these regions listed here just gives us the ability to target them if we need to.  For example, we may need long-running instances in either, but unless they're explicitly targeted we won't have builds ending up there.

scripts/ansible_v2_3_2_0_1_contrib_inventory_openstack.py
git-harry: One thought would be to manage it as part of the gating venv job and carry the script in the venv so that it is updated at the same time as the version of ansible.

odyssey4me: That's a pretty good idea. I like it.

playbooks/slave_security.yml
odyssey4me: In this case we're both building and executing the builds from a remote 'deployment host' which is a jenkins slave. If the target is missing, it is replaced automatically. If were setting these up as slaves then it would be more complex to do that model I think. If we want to set these up as jenkins slaves too, I don't really understand what we gain from it.

nodepool/nodepool-config.yml
```diff
@@ -0,0 +1,89 @@
...
+    # existing nodes that haven't been used will get recreated
+    # every hour.
+    max-ready-age: 3600
+
+providers:
+  - name: pubcloud-iad
+    region-name: 'IAD'
+    cloud: public_cloud
+    api-timeout: 60
+    boot-timeout: 120
```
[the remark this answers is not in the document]
odyssey4me: From https://docs.openstack.org/infra/nodepool/feature/zuulv3/configuration.html this is defined as:

```
Once an instance is active, how long to try connecting to the image via SSH. If the timeout is exceeded, the node launch is aborted and the instance deleted.
```

Are you sure that would be an issue?

Even if it is, these are largely testing values right now - we can alsways change them later as different use-cases come along.

nodepool/nodepool-config.yml
odyssey4me: There are bits here to provide combinations:

label - an arbitrary name (could be ubuntu-xenial-big or ubuntu-xenial-metal)
diskimage - defines the image to use (which effectively defined the contents)
flavor-name - for this provider, the flavor that goes with this label... for others it might be different (ie pubcloud and phobos may use different flavor names)

There are also multiple labels per pool, and multiple pools per provider. Our jenkins integration would then request a particular label from a particular pool for the job. So what you do depends on what you're looking to achieve. If you just want access to more flavor profiles, then we create more labels in the same pool. If you want to perhaps use the same label, but have different flavor profiles per pool then I think that's doable.

playbooks/setup_nodepool.yml
```diff
@@ -0,0 +1,233 @@
...
+
+- name: Setup nodepool servers serially
+  hosts: "nodepool_server"
+  gather_facts: no
+  user: root
+  # NOTE(odyssey4me):
+  # The zookeeper service loses all its data when it restarts,
+  # which may happen when packages are upgraded. To ensure that
+  # this is not an issue for us we execute the playbook in
+  # serial so that the cluster always retains its quorum.
```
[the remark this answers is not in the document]
odyssey4me: The way this has been built allows us to destroy one or all nodes and re-execute the playbooks. Absolutely none of the data needs to be retained. The reason for creating the cluster was just for smoother operations.

So no, there is no documentation right now. I'll add a card to the epic to add a page to confluence for it as I think that's a good idea.

nodepool/nodepool-config.yml
odyssey4me: Can do.

playbooks/setup_openstack_instances.yml
odyssey4me: Hmm, so the task should always run even if the changes above fail somehow? That's a good idea.

odyssey4me: I'll register a card for this. As discussed @mattt416 has some improvements incoming, but we should revisit this concept after that.

playbooks/slave_security.yml
mattt416: It's easier having a node as a slave, because it means you can execute ad-hoc commands on it.  However, for long-running infrastructure I think it makes more sense to have everything applied to a node captured in a playbook that is run end-to-end.

nodepool/nodepool-config.yml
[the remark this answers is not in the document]
hughsaunders: I didn't realise its post-active. Snapshots will cause builds to take longer as the hyps won't have the image cached, but thats all pre-active.

hughsaunders:
Deleted the nodepool related instances and running a complete rebuild as a final test: https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/Setup-Nodepool/detail/Setup-Nodepool/19/pipeline ▶️

hughsaunders:
Sadly that [didn't go so well](https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/Setup-Nodepool/detail/Setup-Nodepool/19/pipeline#step-62-log-19).

```
TASK [Create/delete OpenStack Instances] ***************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: keystoneauth1.exceptions.connection.SSLError: SSL exception connecting to https://iad.images.api.rackspacecloud.com/v2/images: [Errno 2] No such file or directory
failed: [localhost] (item={u'region': u'IAD', u'metadata': {u'group': u'nodepool_server'}, u'flavor': u'general1-8', u'name': u'nodepool-server-iad-1'}) => {"failed": true, "instance": {"flavor": "general1-8", "metadata": {"group": "nodepool_server"}, "name": "nodepool-server-iad-1", "region": "IAD"}, "module_stderr": "Traceback (most recent call last):\n  File \"/tmp/ansible_jUUWLT/ansible_module_os_server.py\", line 685, in <module>\n    main()\n  File \"/tmp/ansible_jUUWLT/ansible_module_os_server.py\", line 674, in main\n    _create_server(module, cloud)\n  File \"/tmp/ansible_jUUWLT/ansible_module_os_server.py\", line 510, in _create_server\n    _exit_hostvars(module, cloud, server)\n  File \"/tmp/ansible_jUUWLT/ansible_module_os_server.py\", line 387, in _exit_hostvars\n    hostvars = meta.get_hostvars_from_server(cloud, server)\n  File \"/var/lib/jenkins/workspace/Setup-Nodepool/.venv/local/lib/python2.7/site-packages/shade/meta.py\", line 465, in get_hostvars_from_server\n  File \"/var/lib/jenkins/workspace/Setup-Nodepool/.venv/local/lib/python2.7/site-packages/shade/openstackcloud.py\", line 4242, in get_image_name\n  File \"/var/lib/jenkins/workspace/Setup-Nodepool/.venv/local/lib/python2.7/site-packages/shade/openstackcloud.py\", line 4233, in get_image_exclude\n  File \"/var/lib/jenkins/workspace/Setup-Nodepool/.venv/local/lib/python2.7/site-packages/shade/openstackcloud.py\", line 1553, in search_images\n  File \"/var/lib/jenkins/workspace/Setup-Nodepool/.venv/local/lib/python2.7/site-packages/shade/openstackcloud.py\", line 2140, in list_images\n  File \"/var/lib/jenkins/workspace/Setup-Nodepool/.venv/local/lib/python2.7/site-packages/keystoneauth1/adapter.py\", line 304, in get\n  File \"/var/lib/jenkins/workspace/Setup-Nodepool/.venv/local/lib/python2.7/site-packages/shade/_adapter.py\", line 154, in request\n  File \"/var/lib/jenkins/workspace/Setup-Nodepool/.venv/local/lib/python2.7/site-packages/shade/task_manager.py\", line 260, in submit_task\n  File \"/var/lib/jenkins/workspace/Setup-Nodepool/.venv/local/lib/python2.7/site-packages/shade/task_manager.py\", line 271, in run_task\n  File \"/var/lib/jenkins/workspace/Setup-Nodepool/.venv/local/lib/python2.7/site-packages/shade/task_manager.py\", line 285, in _run_task\n  File \"/var/lib/jenkins/workspace/Setup-Nodepool/.venv/local/lib/python2.7/site-packages/shade/task_manager.py\", line 99, in wait\n  File \"/var/lib/jenkins/workspace/Setup-Nodepool/.venv/local/lib/python2.7/site-packages/shade/task_manager.py\", line 110, in run\n  File \"/var/lib/jenkins/workspace/Setup-Nodepool/.venv/local/lib/python2.7/site-packages/shade/_adapter.py\", line 152, in main\n  File \"/var/lib/jenkins/workspace/Setup-Nodepool/.venv/local/lib/python2.7/site-packages/keystoneauth1/adapter.py\", line 189, in request\n  File \"/var/lib/jenkins/workspace/Setup-Nodepool/.venv/local/lib/python2.7/site-packages/keystoneauth1/session.py\", line 698, in request\n  File \"/var/lib/jenkins/workspace/Setup-Nodepool/.venv/local/lib/python2.7/site-packages/keystoneauth1/session.py\", line 760, in _send_request\nkeystoneauth1.exceptions.connection.SSLError: SSL exception connecting to https://iad.images.api.rackspacecloud.com/v2/images: [Errno 2] No such file or directory\n", "module_stdout": "", "msg": "MODULE FAILURE", "rc": 0}
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: IOError: [Errno 2] No such file or directory: '/var/lib/jenkins/workspace/Setup-Nodepool/.venv/local/lib/python2.7/site-packages/ansible/modules/cloud/openstack/os_server.py'
fatal: [localhost]: FAILED! => {"failed": true, "msg": "Unexpected failure during module execution.", "stdout": ""}
	to retry, use: --limit @/var/lib/jenkins/workspace/Setup-Nodepool/rpc-gating/playbooks/setup_openstack_instances.retry
```

hughsaunders:
Second attempt succeeded: https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/Setup-Nodepool/detail/Setup-Nodepool/20/pipeline/ ✅ 

@odyssey4me  and @mattt416 have also both seen the os_server module failure, but the conditions for reproducing it have not been discovered yet.

hughsaunders:
The os_server issue is not confined to this job, so while it needs fixing, it doesn't block this PR.

scripts/ansible_v2_3_2_0_1_contrib_inventory_openstack.py
odyssey4me: Card added: https://rpc-openstack.atlassian.net/browse/RE-1195

playbooks/setup_nodepool.yml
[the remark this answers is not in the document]
odyssey4me: Added a card for this in https://rpc-openstack.atlassian.net/browse/RE-1196 - please add comments for anything else that comes to mind might be good to document.

PR #548, opened 2017-11-23
RE-787 Add PR build skipping for matching files

odyssey4me:
This patch introduces the ability in a standard job to define a skip
regex. If all the paths of all the files modified by the pull request
match the regex, the build is ended and marked a success. This allows
builds to be skipped when they aren't impacted by the pull request
changes.

The regex support is provided by Python's re module with the verbose
flag enabled [1], this allow for more readable multi-line regex to be
written without the whitespace impacted the expression. When defining
`skip_pattern` in YAML use `>-` so that the string does not include new
lines.

This change also adds a regex for rpc-openstack pull requests to fix the
functionality lost when those jobs were moved from using rpc_aio.yml to
standard_job_premerge.yml.

This facility can be applied to the project, or to an individual axis.
The skip regex is also entirely optional, by default no build is
skipped.

`common.is_doc_update_pr`, the original implementation of this
functionality has been left untouched. It can be removed once all
old-style jobs using it have been removed.

[1] https://docs.python.org/2/library/re.html#re.VERBOSE

Co-Authored-By: git-harry <git-harry@live.co.uk>

Issue: [RE-787](https://rpc-openstack.atlassian.net/browse/RE-787)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -516,6 +516,42 @@ def docker_cache_workaround(){
    sh "touch -t 201704100000 *.txt"
 }
 
+/* Used to check whether a pull request only includes changes
+ * to files that match the skip regex. If the regex is an
+ * empty string, that is treated as matching nothing.
+ */
+Boolean skip_build_check(String regex) {
+  print "Skipping pattern: '$regex'"
+  if (regex != "") {
+    def rc = sh(
+      script: """#!/bin/bash
+        set -xeu
+        cd ${env.WORKSPACE}/${env.RE_JOB_REPO_NAME}
+        git status
+        git show --stat=400,400 | awk '/\\|/{print \$1}' \
+            |python -c 'import re, sys; all(re.search(r"${regex}", line, re.VERBOSE) for line in sys.stdin) and sys.exit(0) or sys.exit(1)'
+        """,
+        returnStatus: true
+    )
+    if (rc==0) {
+      print "All change files match skip pattern. Skipping..."
+      return true
+    } else if(rc==1) {
+      print "One or more change files not matched by skip pattern. Continuing..."
+      return false
+    } else if(rc==128) {
+      throw new Exception("Directory is not a git repo, cannot compile changes.")
+    }
+  } else {
+    return false
+  } // if
+}
+
+/* DEPRECATED FUNCTION
+ * This function should be removed once it is no longer in use. It remains
+ * while still required by old-style jobs and has been replaced in standard
+ * jobs by `skip_build_check`.
+ */
 def is_doc_update_pr(String git_dir) {
   if (env.ghprbPullId != null) {
     dir(git_dir) {
--- a/rpc_jobs/rpc_openstack.yml
+++ b/rpc_jobs/rpc_openstack.yml
@@ -5,6 +5,13 @@
     series: "master"
     branches:
       - "master"
+    skip_pattern: >-
+      \.md$
+      | \.rst$
+      | ^releasenotes/
+      | ^gating/generate_release_notes/
+      | ^gating/post_merge_test/
+      | ^gating/update_dependencies/
     image:
       - xenial:
           IMAGE: "Ubuntu 16.04.2 LTS prepared for RPC deployment"
@@ -25,6 +32,13 @@
     series: "newton"
     branches:
       - "newton.*"
+    skip_pattern: >-
+      \.md$
+      | \.rst$
+      | ^releasenotes/
+      | ^gating/generate_release_notes/
+      | ^gating/post_merge_test/
+      | ^gating/update_dependencies/
     image:
       - trusty:
           IMAGE: "Ubuntu 14.04.5 LTS prepared for RPC deployment"
@@ -47,6 +61,13 @@
     series: "mitaka"
     branches:
       - "mitaka.*"
+    skip_pattern: >-
+      \.md$
+      | \.rst$
+      | ^releasenotes/
+      | ^gating/generate_release_notes/
+      | ^gating/post_merge_test/
+      | ^gating/update_dependencies/
     image:
       - trusty:
           IMAGE: "Ubuntu 14.04.5 LTS prepared for RPC deployment"
@@ -66,6 +87,13 @@
     series: "liberty"
     branches:
       - "liberty.*"
+    skip_pattern: >-
+      \.md$
+      | \.rst$
+      | ^releasenotes/
+      | ^gating/generate_release_notes/
+      | ^gating/post_merge_test/
+      | ^gating/update_dependencies/
     image:
       - trusty:
           IMAGE: "Ubuntu 14.04.5 LTS prepared for RPC deployment"
@@ -85,6 +113,13 @@
     series: "kilo"
     branches:
       - "kilo.*"
+    skip_pattern: >-
+      \.md$
+      | \.rst$
+      | ^releasenotes/
+      | ^gating/generate_release_notes/
+      | ^gating/post_merge_test/
+      | ^gating/update_dependencies/
     image:
       - trusty:
           IMAGE: "Ubuntu 14.04.5 LTS prepared for RPC deployment"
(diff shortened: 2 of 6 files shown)


rpc_jobs/standard_job_premerge.yml
```diff
@@ -26,6 +26,7 @@
     FLAVOR: "performance1-1"
     IMAGE: "Ubuntu 16.04 LTS (Xenial Xerus) (PVHVM)"
     TRIGGER_PR_PHRASE_ONLY: false
+    skip_list: []
```
mattt416: What if we changed the default to ...

```
    skip_list:
      - '.*md\$'
      - '.*rst\$'
      - '^releasenotes/'
      - '^gating/generate_release_notes/'
      - '^gating/post_merge/'
      - '^gating/update_dependencies/'
```

... and then allowed projects to override if necessary?

rpc_jobs/standard_job_premerge.yml
```diff
@@ -82,6 +83,24 @@
...
+      // in the skip list. If that is the case, we can skip
+      // the rest of the job execution and return success.
+      common.shared_slave() {{
+        common.withRequestedCredentials(credentials) {{
+          stage('Check PR against skip-list') {{
+            print("Triggered by PR: ${{env.ghprbPullLink}}")
+            common.clone_with_pr_refs(
+              "${{env.WORKSPACE}}/${{env.RE_JOB_REPO_NAME}}",
+            )
+            if (common.skip_list_test("{skip_list}")) {{
```
hughsaunders: common.skip_list_test expects a list of strings, but this looks like it will present the whole list as a single string?

rpc_jobs/standard_job_premerge.yml
mancdaz: +1

git-harry: I think we need to be careful here. We cannot know for sure that test code in `gating/post_merge_test/` is not referenced by some other testing. In the case of rpc-openstack `gating/post_merge_test/` is a symlink and so this is fine but that might not be how all projects structure this code and it would be better to skip too little than too much.

pipeline_steps/common.groovy
```diff
@@ -516,33 +516,48 @@ def docker_cache_workaround(){
...
 
-def is_doc_update_pr(String git_dir) {
-  if (env.ghprbPullId != null) {
-    dir(git_dir) {
+/* Used to check whether a PR only includes changes
+ * which are defined in the skip list.
+ * Expects a list of regex items to provide to egrep
+ * and returns a boolean.
+ */
+def skip_list_test(String[] args) {
```
git-harry: `def` can be changed to `Boolean` to both document and enforce that it returns a boolean value.

rpc_jobs/standard_job_premerge.yml
odyssey4me: Yes, I specifically only implemented this skip list for rpc-openstack because that's what always was there. I don't think we can assume that other projects want to skip the same things. Also bear in mind that most projects only have one premerge project defined, so it's not much to manage. RPC-O has many because it's maintaining so many branches. Eventually that will change, so having those repeated bits is a small price.

rpc_jobs/standard_job_premerge.yml
odyssey4me: Is there a better way to make it pass the things correctly? Or perhaps to change the function to receive the things JJB is able to pass. I tried making JJB use an object here and it croaked.

pipeline_steps/common.groovy
odyssey4me: Nice idea - will do.

rpc_jobs/rpc_openstack.yml
```diff
@@ -5,6 +5,13 @@
     series: "master"
     branches:
       - "master"
+    skip_list:
+      - '.*md\$'
+      - '.*rst\$'
+      - '^releasenotes/'
+      - '^gating/generate_release_notes/'
+      - '^gating/post_merge/'
```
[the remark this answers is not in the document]
odyssey4me: Ah, thanks for picking that up! Will fix.

git-harry:
recheck_cit_gh_re_unit_tests

git-harry:
recheck_cit_gh_re_unit_tests

git-harry:
recheck_cit_gh_re_unit_tests

git-harry:
recheck_cit_gh_re_unit_tests

git-harry:
recheck_cit_all

git-harry:
recheck_cit_all

git-harry:
recheck_cit_gh-re_unit_tests

git-harry:
recheck_cit_all

git-harry:
recheck_cit_all

git-harry:
recheck_cit_all

git-harry:
recheck_cit_all

git-harry:
recheck_cit_all

git-harry:
recheck_cit_all

rpc_jobs/unit/re_unit_tests.yml
```diff
@@ -1,3 +1,41 @@
+- job:
+    name: 'RPC-Gating-Unit-Tests-gh'
```
hughsaunders: I assume you intend to remove this before merge?

git-harry: Yes, that's just for testing. It's not yet ready for review. I'd rather see the unit test job modified separately so as not to confuse things.

git-harry:
@odyssey4me @hughsaunders @mattt416 this is ready for review.

rpc_jobs/standard_job_premerge.yml
```diff
@@ -84,80 +84,100 @@
...
 
       String credentials = "{credentials}"
+      String skip_pattern = "{skip_pattern}"
 
-      common.standard_job_slave(env.SLAVE_TYPE) {{
-        // Set the default environment variables used
-        // by the artifact and test result collection.
-        env.RE_HOOK_ARTIFACT_DIR="${{env.WORKSPACE}}/artifacts"
-        env.RE_HOOK_RESULT_DIR="${{env.WORKSPACE}}/results"
+      Boolean run_test
```
mattt416: I think this defaults `run_test` to null, which is false.  Would it be safer to explicitly default it to true?

rpc_jobs/unit/skip_build_check.yml
```diff
@@ -0,0 +1,76 @@
...
+      library "rpc-gating@${RPC_GATING_BRANCH}"
+      common.shared_slave{
+
+        env.RE_JOB_REPO_NAME = "test_repository"
+
+        stage("Setup repository") {
+          sh """#!/bin/bash -xeu
+            mkdir ${env.RE_JOB_REPO_NAME}
+            cd ${env.RE_JOB_REPO_NAME}
+            git init .
```
mattt416: Not a huge saving, but I think `common.shared_slave` already clones rpc-gating so you could probably just commit some dummy files there for testing?

mattt416: ^^^ is probably a bad idea, prefer this.

rpc_jobs/standard_job_premerge.yml
git-harry: That's a fair point, the stage "should" always run but Jenkins is not one to be trusted. I've made the following change
```diff
diff --git a/rpc_jobs/standard_job_premerge.yml b/rpc_jobs/standard_job_premerge.yml
index da02428..55b56d8 100644
--- a/rpc_jobs/standard_job_premerge.yml
+++ b/rpc_jobs/standard_job_premerge.yml
@@ -86,7 +86,7 @@
       String credentials = "{credentials}"
       String skip_pattern = "{skip_pattern}"
 
-      Boolean run_test
+      Boolean run_test = true
       // We need to checkout the repo on the CIT Slave so that
       // we can check whether the patch only includes changes
       // in files matching the skip pattern. If that is the case,
```

PR #551, opened 2017-11-24
RE-1151 Add playbook / jobs to deploy repo servers

mattt416:
This commit leverages the work done in [1] to continuously deploy
repo servers for artifact storage.

In order to do this, setup_openstack_instances.yml had to be
modified to provision block storage, which is required for storage of
artifacts.  This primarily involves:

- adjusting the initial unlock to only happen when the instance is
  marked for deletion (state=absent), or when there are volumes defined
  in instance_list
- adding the necessary tasks to deploy and attach the specified volumes
- adding an inventory refresh to provide an updated inventory for when
  we need to query inventory to determine if an instance needs to be
  locked
- finally, we check if the instance is in the inventory to determine if
  it needs to be locked

To keep things simple for the first iteration, this PR does
not handle detaching / deleting volumes on a running instance as a fair
bit of work would need to take place to ensure the disks are not in
use.  That said, volumes with a state of `absent` on an instance with
state `absent` will result in those volumes being deleted after the
instance is removed.

NOTE: To address a failing lint job, we've had to skip playbooks/roles
      in check_python() in lint.sh.  We have external ansible roles
      being pulled in here and shouldn't run a python lint against this
      dir.

TODO:
  - lock down any necessary ports w/ ufw
  - look at combining setup_openstack_instances.yml,
    drop_ssh_auth_keys.yml, and slave_security.yml into the same
    venvPlaybook execution in both rpc_jobs/setup_nodepool.yml and
    rpc_jobs/repo_server.yml (this will avoid an extra inventory
    lookup since the inventory is refreshed at the end of
    playbooks/setup_openstack_instances.yml)

[1] https://github.com/rcbops/rpc-gating/pull/542/

Issue: [RE-1151](https://rpc-openstack.atlassian.net/browse/RE-1151)

### DIFF
--- a/playbooks/repo_server.yml
+++ b/playbooks/repo_server.yml
@@ -1,0 +1,54 @@
+---
+- hosts: repo_all
+  user: root
+  gather_facts: no
+  tasks:
+    - name: Gather facts
+      setup:
+        gather_subset: "!facter,!ohai"
+
+    - set_fact:
+        repo_server_volumes: "{{ item['volumes'] }}"
+      when: "{{ item['name'] == inventory_hostname }}"
+      with_items: "{{ instance_list | default([]) }}"
+
+    - name: Create vg0 volume group
+      lvg:
+        vg: vg0
+        pvs: "{{ repo_server_volumes | map(attribute='device') | join(',') }}"
+        state: present
+
+    - name: Create repo logical volume
+      lvol:
+        lv: repo
+        vg: vg0
+        size: 100%PVS
+        shrink: no
+        state: present
+
+    - name: Create filesystem on /dev/vg0/repo
+      filesystem:
+        fstype: ext4
+        force: no
+        dev: /dev/vg0/repo
+        resizefs: yes # suicide?
+
+    - name: Mount /dev/vg0/repo
+      mount:
+        path: /var/www/repo
+        src: /dev/vg0/repo
+        fstype: ext4
+        state: mounted
+
+- hosts: repo_all
+  user: root
+  gather_facts: no
+  vars:
+    is_metal: true
+    openstack_release: testing
+  pre_tasks:
+    - name: Gather facts
+      setup:
+        gather_subset: "!facter,!ohai"
+  roles:
+    - role: "repo_server"
--- a/playbooks/repo_server.yml
+++ b/playbooks/repo_server.yml
@@ -7,9 +7,10 @@
       setup:
         gather_subset: "!facter,!ohai"
 
-    - set_fact:
+    - name: "Create fact for instance's volumes"
+      set_fact:
         repo_server_volumes: "{{ item['volumes'] }}"
-      when: "{{ item['name'] == inventory_hostname }}"
+      when: "item['name'] == inventory_hostname"
       with_items: "{{ instance_list | default([]) }}"
 
     - name: Create vg0 volume group
--- a/playbooks/setup_openstack_instances.yml
+++ b/playbooks/setup_openstack_instances.yml
@@ -8,6 +8,22 @@
 #     region: DFW
 #     metadata:
 #       group: mygroup
+#
+# Additionally, you can optionally specify volumes
+# to be created and attached:
+#
+# instance_list:
+#   - name: instance1
+#     flavor: general1-1
+#     region: DFW
+#     volumes:
+#       - name: vol1
+#         device: /dev/xvde
+#         type: SATA
+#         size: 1024
+#     metadata:
+#       group: mygroup
+#
 
 - name: Create/delete OpenStack Instances
   hosts: localhost
@@ -23,7 +39,7 @@
       # deployment tooling.
       build_config: core
   tasks:
-    - name: Unlock instances scheduled to be changed
+    - name: Unlock instances scheduled to be deleted or where volumes may need to be added
       os_server_actions:
         action: unlock
         server: "{{ instance['name'] | lower }}"
@@ -31,8 +47,7 @@
         region_name: "{{ instance['region'] | default('IAD') }}"
         wait: yes
         timeout: 900
-      when:
-        - "instance['name'] in groups['all']"
+      when: "{{ instance['name'] in groups['all'] and (instance['state'] | default('present') == 'absent' or (instance['state'] | default('present') == 'present' and 'volumes' in instance)) }}"
       with_items: "{{ instance_list | default([]) }}"
       loop_control:
         loop_var: instance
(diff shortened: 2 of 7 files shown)


mattt416:

mattt416:
recheck_cit_xenial_lint_test

mattt416:
Hmm, the `xenial_lint_test` is failing on:

```
./playbooks/roles/apt_package_pinning/doc/source/conf.py:72:80: E501 line too long (80 > 79 characters)
./playbooks/roles/apt_package_pinning/releasenotes/source/conf.py:68:80: E501 line too long (80 > 79 characters)
./playbooks/roles/repo_server/doc/source/conf.py:72:80: E501 line too long (80 > 79 characters)
./playbooks/roles/repo_server/releasenotes/source/conf.py:68:80: E501 line too long (80 > 79 characters)
Python syntax fail
```

Do we skip `playbooks/roles`, or adjust our flake rules to match what OSA's using?

odyssey4me:
@mattt416 Our lint tests should not be doing tests against external deps.

mattt416:
@odyssey4me I'll update to skip playbooks/roles, but hopefully we don't end up adding custom roles there that end up not getting (python) linted.  :(

repo_server/instances.yml
```diff
@@ -0,0 +1,32 @@
+---
+instance_list:
```
hughsaunders: I realise state has been defaulted so isn't strictly necessary here, but could you add it in for documentation/example purposes?. Then if we need to remove these instances, its obvious where to set absent.

mattt416: Fair enough ... I've updated repo_server/instances.yml and the commented instances outline in playbooks/setup_openstack_instances.yml.

playbooks/repo_server.yml
```diff
@@ -0,0 +1,54 @@
...
+        src: /dev/vg0/repo
+        fstype: ext4
+        state: mounted
+
+- hosts: repo_all
+  user: root
+  gather_facts: no
+  vars:
+    is_metal: true
+    openstack_release: testing
```
evrardjp: Is there a reason to override openstack_release without the ability to override with group vars?
IIRC, openstack_release should be taken from variables...

playbooks/setup_openstack_instances.yml
```diff
@@ -23,16 +42,15 @@
...
       os_server_actions:
         action: unlock
         server: "{{ instance['name'] | lower }}"
         cloud: "{{ cloud_name | default('public_cloud') }}"
         region_name: "{{ instance['region'] | default('IAD') }}"
         wait: yes
         timeout: 900
-      when:
-        - "instance['name'] in groups['all']"
+      when: "{{ instance['name'] in groups['all'] and (instance['state'] | default('present') == 'absent' or (instance['state'] | default('present') == 'present' and 'volumes' in instance)) }}"
```
evrardjp: nit: You could make this more readable by writing
`
when:
- instance['name'] in groups['all']
- (instance['state'] | default('present') == 'absent' or (instance['state'] | default('present') == 'present' and 'volumes' in instance))
`

playbooks/setup_openstack_instances.yml
```diff
@@ -58,16 +76,74 @@
...
+        display_name: "{{ item[1]['name'] }}"
+        volume_type: "{{ item[1]['type'] }}"
+        size: "{{ item[1]['size'] }}"
+        state: absent
+        cloud: "{{ cloud_name | default('public_cloud') }}"
+        region_name: "{{ item[0]['region'] }}"
+        wait: yes
+        timeout: 900
+      when:
+        - "{{ item[0]['state'] | default('present') == 'absent' and item[1]['state'] | default('present') == 'absent' }}"
```
evrardjp: if you put that in two lines it improves readablity.

playbooks/setup_openstack_instances.yml
```diff
@@ -58,16 +76,74 @@
...
+        region_name: "{{ item[0]['region'] }}"
+        wait: yes
+        timeout: 900
+      when:
+        - "{{ item[0]['state'] | default('present') != 'absent' and item[1]['state'] | default('present') != 'absent' }}"
+      with_subelements:
+        - "{{ instance_list | default([]) }}"
+        - volumes
+        - skip_missing: True
+
```
evrardjp: Why not doing all those operations based on present / absent state on a higher level, in a include for example, and simplify the conditions?

Alternatively, instead of doing it in a serial way, you could do it in parallel, by generating a list of hostnames based on instance_list, and then use a connection local to do your openstack module things. That would be faster for many elements.

playbooks/repo_server.yml
mattt416: The reason why I set it here is because we're not actually deploying a full openstack-ansible, so we have no legitimate openstack_release to set.  I'm only setting the var for the purpose of allowing repo_server to deploy without throwing an error due to missing vars.

playbooks/repo_server.yml
```diff
@@ -0,0 +1,54 @@
...
+  user: root
+  gather_facts: no
+  tasks:
+    - name: Gather facts
+      setup:
+        gather_subset: "!facter,!ohai"
+
+    - set_fact:
+        repo_server_volumes: "{{ item['volumes'] }}"
+      when: "{{ item['name'] == inventory_hostname }}"
```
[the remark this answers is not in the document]
mattt416: Makes sense, updating.

playbooks/repo_server.yml
```diff
@@ -0,0 +1,54 @@
+---
+- hosts: repo_all
+  user: root
+  gather_facts: no
+  tasks:
+    - name: Gather facts
+      setup:
+        gather_subset: "!facter,!ohai"
+
+    - set_fact:
```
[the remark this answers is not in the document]
mattt416: Makes sense, adding.

playbooks/setup_openstack_instances.yml
mattt416: Your comment got a bit garbled, but if I understand I can remove the jinja `{{` from the first condition but then need to leave it in the second since we're using jinja filters?

playbooks/setup_openstack_instances.yml
mattt416: Agreed, updating.

playbooks/setup_openstack_instances.yml
mattt416: I'll ping you tomorrow @evrardjp to discuss this point a little better.  I'd like the logic to be clearer if possible.

PR #554, opened 2017-11-29
 RE-1163 Add rpc-openstack master tox jobs

git-harry:
Issue: [RE-1163](https://rpc-openstack.atlassian.net/browse/RE-1163)

### DIFF
--- a/rpc_jobs/rpc_openstack.yml
+++ b/rpc_jobs/rpc_openstack.yml
@@ -19,6 +19,27 @@
       - 'PR_{repo_name}-{series}-{image}-{scenario}-{action}'
 
 - project:
+    name: "rpc-openstack-master-tox-premerge"
+    repo_name: "rpc-openstack"
+    repo_url: "https://github.com/rcbops/rpc-openstack"
+    series: "master"
+    branches:
+      - "master"
+    image:
+      - container:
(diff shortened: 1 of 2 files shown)


git-harry:

rpc_jobs/rpc_openstack.yml
```diff
@@ -18,6 +18,27 @@
...
+      - "master"
+    image:
+      - container:
+          SLAVE_TYPE: "container"
+    scenario:
+      - "ansible-lint"
+      - "ansible-syntax"
+      - "bashate"
+      - "pep8"
+      - "releasenotes"
```
hughsaunders: What does this do, and do we need to have this as well as the [release PR job](https://github.com/rcbops/rpc-gating/blob/master/rpc_jobs/standard_job_premerge_release.yml)?

git-harry: https://github.com/rcbops/rpc-openstack/blob/master/tox.ini#L58 - this looks to be doing something different.

This change is simply moving the existing tests. I'd be inclined to leave any modification the tests to a separate task.

hughsaunders: I'm so glad we have two totally separate ways of generating release notes. That can be addressed in a different issue.

PR #556, opened 2017-11-30
RO-3147 Fix ansible warnings

major:
This patches fixes a warning from the command module when
using `tar`.

    [WARNING]: Consider using unarchive module rather than running tar

Issue: [RO-3147](https://rpc-openstack.atlassian.net/browse/RO-3147)


mattt416:
@major So potentially dumb question ... why can we not use the unarchive module here?

PR #561, opened 2017-12-01
RE-1150 Remove old RPC-O specific code

odyssey4me:
This removes jobs that have either been replaced or have not
been working for some time and their related code. Most of
these have been replaced with standard jobs, with the complexity
of how to handle the job execution living in the test repo
instead of the gating repo.

Issue: [RE-1150](https://rpc-openstack.atlassian.net/browse/RE-1150)


pipeline_steps/aio_prepare.groovy
```diff
@@ -1,43 +0,0 @@
-def prepare(){
-  common.conditionalStage(
-    stage_name: "Prepare Deployment",
```
mattt416: This stage is still referenced in `rpc_jobs/irr_role_test.yml` and `docs/job_structure.md`.

mattt416:
Some other files which I think we can nuke:

* pipeline_steps/horizon.groovy
* pipeline_steps/kibana.groovy
* pipeline_steps/maas.groovy
* playbooks/multi_node_aio_maas_cleanup.yml
* playbooks/multi_node_aio_maas_entities.yml
* playbooks/vars/onmetal.yml

Also, looks like there's some stuff in the following which can be cleaned up:

* playbooks/vars/all.yml

pipeline_steps/common.groovy
```diff
@@ -92,85 +92,27 @@ def venvGalaxy(String[] args){
  *  args: list of string args to pass to ansible-playbook
  */
 def venvPlaybook(Map args){
-  withEnv(get_deploy_script_env()){
```
git-harry: This change is not specific to removing RPCO related code. `venvPlaybook` is used a lot in this repository and so I think this shouldn't be changed by this patch.

mattt416:
I think @git-harry's suggestion makes sense.  Ripping this all out wholesale is probably going to result in something breaking.

pipeline_steps/common.groovy
odyssey4me: get_deploy_script_env is only used for the old RPC-O code, because it's only related to the setting of RPC-O deploy vars - that's why this change is included.

odyssey4me:
The trouble with trying to break this into pieces is that they are all deeply intertwined. I originally tried to do it in pieces, but the first few steps removed jobs, and nothing else. If you feel it would be advantageous to really split it down and give a reason for each, that's fine - but it'll take several days to get through all that.

mattt416:
Thoughts @hughsaunders ?  We're down a reviewer atm, so it's going to take us longer to get things through the board.  Shall we do one large patchset and deal w/ the potential fallout, or break it up into numerous changes to minimise damage?

hughsaunders:
Kicked off an RPCO PM build to check this doesn't impact it: https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-newton-xenial-swift-deploy/83/ ▶️

odyssey4me:
Closing this PR to leave it as a reference. I'll open another one which will be revised.

PR #565, opened 2017-12-05
RE-1200 override MaaS Verify retries

hughsaunders:
At the end of the RE-1200 deploy, a large spike in rabbitmq messages
was observed that built up over ~40 minutes before quickly clearing.
This caused the verification stage to fail. An immediate reverify
succeeded as the spike had subsided.

To avoid this situation in future, we extend maas retries to 15 minutes,
and retry the whole verification phase 3 times. This will allow for the
duration of a similar spike, and time for the alarm status to change in
MaaS.

Issue: [RE-1200](https://rpc-openstack.atlassian.net/browse/RE-1200)

### DIFF
--- a/rpc_jobs/phobos.yml
+++ b/rpc_jobs/phobos.yml
@@ -56,7 +56,8 @@
       }
       void verify(){
         sshagent (credentials:['rpc-jenkins-svc-github-ssh-key']){
-          sh """#!/bin/bash
+          retry(3){
+            sh """#!/bin/bash
       # Deliberate outdent to match bash's heredoc requirements.
       sudo -E bash <<EOF
               cd rpc-eng-ops/phobos-re-automation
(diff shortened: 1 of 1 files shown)


hughsaunders:
Makes use of: https://github.com/rcbops/rpc-maas/pull/403
However these PRs can safely be merged in either order.

rpc_jobs/phobos.yml
```diff
@@ -192,3 +198,4 @@
           """
         }
       } // phobos node
+      } // timestamps
```
mattt416: Can you indent the use_node block?

mancdaz: I @mattt416 I added the indent - it seems to have thrown off the formatting when viewing the commit diff. I'm pretty sure it's all the same apart from the timestamp wrapper, but just double check for me...

PR #568, opened 2017-12-11
RE-1211 Reduce num-to-keep for PM jobs

mattt416:
Since these jobs run once a day, it should be fine to retain two weeks
worth of logs for troubleshooting, investigating, etc.

Issue: [RE-1211](https://rpc-openstack.atlassian.net/browse/RE-1211)


mattt416:
@odyssey4me @antonym will this reduction cause any issues?

odyssey4me:
@mattt416 No objection from me, but I don't know whether @claco and team are relying on a longer history?

antonym:
I'm good with two weeks.  If we need to increase it, we can always make a case for it then.

andymcc:
LGTM

PR #569, opened 2017-12-12
RE-35 Merge instance_prep into allocate_pubcloud

mattt416:
Previously, instance_prep ran before we did any ansible tasks on a
server.  However, additional tasks were added to allocate_pubcloud that
run before instance_prep is executed, which results in jobs failing
when python does not exist in the base image.

This commit merges instance_prep with allocate_pubcloud.

Issue: [RE-35](https://rpc-openstack.atlassian.net/browse/RE-35)


mattt416:
Testing [here](https://rpc.jenkins.cit.rackspace.net/view/Upgrades/job/PM_rpc-upgrades-master-trusty_mnaio-swift-liberty_to_newton_leap/73/), and it looks to have gotten past the error that @antonym originally reported.

PR #571, opened 2017-12-12
RLM-820 Reduce timeout during startup

antonym:
Reduce the time on startup to 1 minute instead of 5
to see if this causes any issues.

Issue: [RLM-820](https://rpc-openstack.atlassian.net/browse/RLM-820)


antonym:
recheck_all

antonym:
recheck_cit_xenial_lint_test

antonym:
Alright, will close out then.

PR #575, opened 2017-12-19
RLM-316 Add PM deploy tests for artifact 'loose' mode

odyssey4me:
The 'loose' mode of using artifacts has been introduced
to help with initial deployments where the packages on
the hosts may be newer than those in the apt artifacts
repo. This patch adds two more tests which facilitate
verifying that this mode works on a day to day basis.

The existing tests are not changed so that the jobs
retain their history and so that the merge trigger
jobs also do not need to change.

The PR to implement the changes being tested by this
job are in https://github.com/rcbops/rpc-openstack/pull/2739

Issue: [RLM-316](https://rpc-openstack.atlassian.net/browse/RLM-316)

### DIFF
--- a/rpc_jobs/rpc_openstack.yml
+++ b/rpc_jobs/rpc_openstack.yml
@@ -174,6 +174,10 @@
           IMAGE: "Ubuntu 16.04.2 LTS prepared for RPC deployment"
       - trusty:
           IMAGE: "Ubuntu 14.04.5 LTS prepared for RPC deployment"
+      - xenial_loose:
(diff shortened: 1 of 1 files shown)


odyssey4me:
@claco Strict mode is the default, and the 'xenial' job, for example, still uses the fixed start point images. The 'xenial_loose' job will build from the latest pubcloud image which is updated around every two weeks. We can make it to the upgrade if you like, but that will be done in https://github.com/rcbops/rpc-openstack/pull/2739 not this patch.

claco:
@odyssey4me Thanks. Works for me. I'd think us forcing upgrade vs waiting on 2 week updates by the images team would surface issues sooner, which is a good thing.

PR #577, opened 2018-01-02
RE-1191 Control Log Storage

hughsaunders:
This PR adds two controls for log storage:

1. Ensure that no console log may exceed 200MB - the job will be killed if that limit is exceeded
1. Ensure that all jobs have a retention policy so logs are not stored indefinitely. A lint rule is added to enforce this. 

Issue: [RE-1191](https://rpc-openstack.atlassian.net/browse/RE-1191)

### DIFF
--- a/lint.sh
+++ b/lint.sh
@@ -25,6 +25,7 @@ create_jjb_ini(){
   echo -e "[jenkins]\nurl=http://127.0.0.1:8080" > lint_jjb.ini
 }
 
+# Check JJB for syntax
 check_jjb(){
   which jenkins-jobs >/dev/null \
     || { echo "jenkins-jobs unavailble, please install jenkins-job-builder from pip"
@@ -87,18 +88,21 @@ check_bash(){
   done < <(find ${fargs[@]} -iname \*.sh)
 }
 
-check_naming_standards() {
+check_jjb_lint() {
+  # Check JJB for internal standards:
+  #   * naming conventions
+  #   * retention policy
   # Limits lint to only gating unique files
   # Excludes webhooktranslator as it has additional packaging
-  # and it's own tox
+  # and its own tox.
   # Excludes NonCPS.groovy as this filename is required, but
   # not matching rpc-gating conventions
   dirs_to_lint="pipeline_steps,rpc_jobs,scripts"
   exclude_files="NonCPS.groovy"
-  python scripts/lint_naming_conventions.py \
+  python scripts/lint_jjb.py \
     --dirs ${dirs_to_lint} --exclude ${exclude_files} \
-    && echo "Naming conventions: OK" \
-    || { echo "Naming conventions: FAIL"; rc=1; }
+    && echo "JJB Lint: OK" \
+    || { echo "JJB Lint: FAIL"; rc=1; }
 }
 
 check_python(){
@@ -166,7 +170,7 @@ check_groovy
 check_ansible
 check_bash
 check_python
-check_naming_standards
+check_jjb_lint
 check_webhooktranslator
 
 if [[ $rc == 0 ]]
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -825,7 +825,6 @@ void withRequestedCredentials(String list_of_cred_ids, Closure body){
   }
 }
 
-
 Cause getRootCause(Cause cause){
     if (cause.class.toString().contains("UpstreamCause")) {
          for (upCause in cause.upstreamCauses) {
@@ -858,6 +857,22 @@ void setTriggerVars(){
       env.RE_JOB_TRIGGER="OTHER"
   }
   print ("Trigger: ${env.RE_JOB_TRIGGER} (${env.RE_JOB_TRIGGER_DETAIL})")
+}
+
+// add wrappers that should be used for all jobs.
+// max log size is in MB
+void globalWraps(Closure body){
+  // global timeout is long, so individual jobs can set shorter timeouts and
+  // still have to cleanup, archive atefacts etc.
+  timeout(time: 10, unit: 'HOURS'){
+    wrap([$class: 'LogfilesizecheckerWrapper', 'maxLogSize': 200, 'failBuild': true, 'setOwn': true]) {
+      timestamps{
+        print("common.globalWraps pre body")
+        body()
+        print("common.globalWraps post body")
+      } // timestamps
+    } // log size limiter
+  } // timeouts
 }
 
 return this
--- a/rpc_jobs/periodic_cleanup.yml
+++ b/rpc_jobs/periodic_cleanup.yml
@@ -25,6 +25,7 @@
           days-to-keep: 3
     dsl: |
       library "rpc-gating@${RPC_GATING_BRANCH}"
+
       // Get list of jenkins slaves
       @NonCPS
       def getLongRunningNodes() {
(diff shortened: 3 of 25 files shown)


scripts/lint_naming_conventions.py
```diff
@@ -89,6 +93,22 @@ def parse_file_name(in_dir, in_file):
     return 0
 
 
+# Ensure all jobs have a retention policy
+def check_retention(job, in_file):
```
git-harry: This function doesn't fail on `build-discarder: {}`, is that intentional? Based on [1] I'd have thought this would want to require either `days-to-keep` or `num-to-keep` is set. Is it worth putting limits on what these values can be?

[1] https://docs.openstack.org/infra/jenkins-job-builder/properties.html#properties.build-discarder

scripts/lint_naming_conventions.py
```diff
@@ -89,6 +93,22 @@ def parse_file_name(in_dir, in_file):
     return 0
 
 
+# Ensure all jobs have a retention policy
```
git-harry: If you are going to add the functionality to this file, I think you should rename it and modify any references to it being specific to naming conventions because it and its use in lint.sh is a bit misleading.

pipeline_steps/common.groovy
```diff
@@ -904,4 +906,20 @@ void setTriggerVars(){
   print ("Trigger: ${env.RE_JOB_TRIGGER} (${env.RE_JOB_TRIGGER_DETAIL})")
 }
 
+// add wrappers that should be used for all jobs.
+// max log size is in MB
+void globalWraps(Closure body){
+  // global timeout is long, so individual jobs can set shorter timeouts and
+  // and still have to cleanup, archive atefacts etc.
```
mattt416: Nit: s/atefacts/artifacts/

pipeline_steps/common.groovy
```diff
@@ -904,4 +906,20 @@ void setTriggerVars(){
   print ("Trigger: ${env.RE_JOB_TRIGGER} (${env.RE_JOB_TRIGGER_DETAIL})")
 }
 
+// add wrappers that should be used for all jobs.
+// max log size is in MB
+void globalWraps(Closure body){
```
mattt416: Would it make sense to add another parameter for the timeout value, so that individual jobs can override on a case-by-case basis without having to add inner timeouts specific for the job?  If we do, I'd personally like to see the default (for when none specified) be far lower, like 1 hour.  This change would also allow us to remove some existing timeouts that are littered through jobs and common utils.

rpc_jobs/periodic_cleanup.yml
```diff
@@ -40,64 +38,68 @@
...
-      for (int i=0; i<nodes.size(); i++){
-        nodeName = nodes[i]
-        parallel_steps['node_'+nodeName] = {
-          common.use_node(nodeName){
-            stage("Cleanup "+nodeName){
-              dir("rpc-gating") {
-                sh "scripts/workspace_cleanup.sh"
-                sh "scripts/tmp_cleanup.sh"
-                sh "scripts/docker_cleanup.sh"
+      library "rpc-gating@${RPC_GATING_BRANCH}"
```
git-harry: Why was this line moved? There is a pattern in place for having this as the first line, unless there is a good reason to move this I think it should be left where it is.

pipeline_steps/common.groovy
```diff
@@ -729,7 +729,10 @@ void use_node(String label=null, body){
                          env.RPC_GATING_BRANCH)
       install_ansible()
       print "${env.NODE_NAME} preparation complete, now ready for use."
-      body()
+      // Ensure log size is not excessive
```
git-harry: Why is this needed here, why does `globalWraps` not suffice?

pipeline_steps/common.groovy
```diff
@@ -904,4 +906,20 @@ void setTriggerVars(){
   print ("Trigger: ${env.RE_JOB_TRIGGER} (${env.RE_JOB_TRIGGER_DETAIL})")
 }
 
+// add wrappers that should be used for all jobs.
+// max log size is in MB
+void globalWraps(Closure body){
+  // global timeout is long, so individual jobs can set shorter timeouts and
+  // and still have to cleanup, archive atefacts etc.
+  timeout(time: 10, unit: 'HOURS'){
+    wrap([$class: 'LogfilesizecheckerWrapper', 'maxLogSize': 200, 'failBuild': true, 'setOwn': true]) {
```
git-harry: What does `'setOwn'` do?

scripts/lint_naming_conventions.py
mattt416: Yep, I think we should check specific values ... with at minimum value > 0 being acceptable.

pipeline_steps/common.groovy
git-harry: I think this is a good suggestion. I'd prefer to see it done as a separate patch. `globalWraps` allows for additional layers to be added painlessly. The timestamp and log size pieces don't duplicate anything existing. I'd suggest fixing this code without including any timeout changes then creating a separate pull request to address all the occurrences of timeout.

rpc_jobs/sync_credential_store.yml
```diff
@@ -2,6 +2,9 @@
     name: 'Sync-Credential-Store'
     project-type: workflow
     concurrent: false
+    properties:
+      - build-discarder:
+          days-to-keep: 360
```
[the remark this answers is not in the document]
hughsaunders: Because it hardly ever runs, doesn't log much, and the results could be useful when tracking down why a credentials does/doesn't exist.

pipeline_steps/common.groovy
hughsaunders: Ooops, that was my first attempt before I added globalwraps. Will remove.

pipeline_steps/common.groovy
hughsaunders: AFAICT it makes maxLogSize active, otherwise a value from Jenkins global config is used. I'd prefer to keep stuff out of Jenkins global config as its not in SCM.

rpc_jobs/periodic_cleanup.yml
hughsaunders: This was to match the [phobos](https://github.com/rcbops/rpc-gating/blob/master/rpc_jobs/phobos.yml#L91) job where the functions are defined first, then the library statement is still the first line to get executed. 

I'm happy to put the library line back at the top, but should probably change that job as well.

pipeline_steps/common.groovy

pipeline_steps/common.groovy

rpc_jobs/periodic_cleanup.yml
hughsaunders: Done

scripts/lint_naming_conventions.py
hughsaunders: Done

hughsaunders:
Comments have been addressed, now ready for review.

pipeline_steps/common.groovy
hughsaunders: OK, Will do this as a separate patch.

hughsaunders: https://rpc-openstack.atlassian.net/browse/RE-1263

scripts/lint_jjb.py
```diff
@@ -1,18 +1,21 @@
...
+#        https://github.com/rcbops/rpc-gating#naming-conventions
+#   * Retention Policy
 
 import argparse
 import os
 import re
 import sys
 import yaml
 
+import jmespath
```
mattt416: Do we need to specify this in requirements/constraints?

lint.sh
```diff
@@ -87,15 +88,20 @@ check_bash(){
...
   # not matching rpc-gating conventions
+  # Also runs other checks that require parsing JJB yaml including:
+  #   * Ensure retention policy is set for all jobs.
   dirs_to_lint="pipeline_steps,rpc_jobs,scripts"
   exclude_files="NonCPS.groovy"
-  python scripts/lint_naming_conventions.py \
+  python scripts/lint_jjb.py \
     --dirs ${dirs_to_lint} --exclude ${exclude_files} \
     && echo "Naming conventions: OK" \
     || { echo "Naming conventions: FAIL"; rc=1; }
```
mattt416: Probably worth updating these strings here.

scripts/lint_jjb.py
hughsaunders: Already in both, as used in other scripts.

lint.sh
hughsaunders: Done

hughsaunders:
Rebased and addressed comments, ready for re-review.

scripts/lint_jjb.py
```diff
@@ -89,6 +96,67 @@ def parse_file_name(in_dir, in_file):
...
+def check_retention(job, in_file):
+    error_message = ("{f}/{j}: Valid build-discarder config not found."
+                     " Please ensure days-to-keep (0 < x <= 365) or"
+                     " num-to-keep (0 < x <= 100) is specified."
+                     " Properties: {p}\n"
+                     .format(j=job['name'], f=in_file,
+                             p=job.get('properties')))
+
+    error_counter = 0
+
```
git-harry: A for-loop might be nicer here.

```python
checks = (
    ('days-to-keep', 0, 365),
    ('num-to-keep', 0, 100),
)

for (retention_type, min_retention, max_retention) in checks:
    try:
        check_retention_value(job, retention_type, min_retention, max_retention)
    except (TypeError, IndexError):
        return_value = 1
    except ValueError as e:
        return_value = 1
        print e
    else:
        return_value = 0
        break
else:
    sys.stderr.write(error_message)

return return_value
```

scripts/lint_jjb.py
```diff
@@ -89,6 +96,67 @@ def parse_file_name(in_dir, in_file):
...
+        print e
+        error_counter += 1
+
+    if (error_counter > 1):
+        sys.stderr.write(error_message)
+        return 1
+
+    return 0
+
+
```
git-harry: `out` defined on line 175 (I can't comment on that part of the file) should be updated so that it doesn't mislead on failure.

scripts/lint_jjb.py
```diff
@@ -1,18 +1,21 @@
 #!/usr/bin/env python
 
-# Script used to review jenkins jobs and report job names
-# not following conventions
-# https://github.com/rcbops/rpc-gating#naming-conventions
+# Script used to review jenkins jobs for compliance with internal conventions:
+#   * Naming convetions:
```
git-harry: conventions

lint.sh
```diff
@@ -87,18 +88,23 @@ check_bash(){
   done < <(find ${fargs[@]} -iname \*.sh)
 }
 
-check_naming_standards() {
+# Check JJB for internal standards:
+#   * naming conventions
+#   * retention policy
```
git-harry: The above three lines look like they should be moved into the function. At the moment there are comments about the function either side the start of its definition.

lint.sh
```diff
@@ -87,18 +88,23 @@ check_bash(){
...
+check_jjb_lint() {
   # Limits lint to only gating unique files
   # Excludes webhooktranslator as it has additional packaging
-  # and it's own tox
+  # and it's own tox.
   # Excludes NonCPS.groovy as this filename is required, but
   # not matching rpc-gating conventions
+  # Also runs other checks that require parsing JJB yaml including:
+  #   * Ensure retention policy is set for all jobs.
```
git-harry: These lines read like they were added before `check_naming_standards` was changed to `check_jjb_lint` and so should probably be removed.

lint.sh
hughsaunders: Done

lint.sh
```diff
@@ -87,18 +88,23 @@ check_bash(){
...
 
-check_naming_standards() {
+# Check JJB for internal standards:
+#   * naming conventions
+#   * retention policy
+check_jjb_lint() {
   # Limits lint to only gating unique files
   # Excludes webhooktranslator as it has additional packaging
-  # and it's own tox
+  # and it's own tox.
```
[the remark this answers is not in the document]
hughsaunders: Done

lint.sh
hughsaunders: Done

pipeline_steps/common.groovy
```diff
@@ -866,7 +862,7 @@ void setTriggerVars(){
 // max log size is in MB
 void globalWraps(Closure body){
   // global timeout is long, so individual jobs can set shorter timeouts and
-  // and still have to cleanup, archive atefacts etc.
+  // and still have to cleanup, archive artifacts etc.
```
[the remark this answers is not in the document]

scripts/lint_jjb.py
hughsaunders: Done

scripts/lint_jjb.py
hughsaunders: Much better thanks, shortcuts the second check if the first check is successful.

scripts/lint_jjb.py
hughsaunders: Done, also updated readme with retention policy requirement.

pipeline_steps/common.groovy
```diff
@@ -685,10 +685,6 @@ void use_node(String label=null, body){
                          env.RPC_GATING_BRANCH)
       install_ansible()
       print "${env.NODE_NAME} preparation complete, now ready for use."
-      // Ensure log size is not excessive
-      wrap([$class: 'LogfilesizecheckerWrapper', 'maxLogSize': 200, 'failBuild': false, 'setOwn': true]) {
-        body()
-      }
```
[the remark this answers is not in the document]
hughsaunders: And it shouldn't have removed the body call ⚠️ 
Done.

PR #578, opened 2018-01-02
RE-1225 Exclude webhook-proxy from cleanup

mattt416:
The periodic cleanup job is failing because it's attempting to delete
an instance called `webhook-proxy`, which is locked and cannot be
deleted.

We already have an instance called `WEBHOOK-PROXY` and a periodic
cleanup exclusion for this, so this commit simply adds an exclusion for
`webhook-proxy` also.

Issue: [RE-1225](https://rpc-openstack.atlassian.net/browse/RE-1225)

### DIFF
--- a/rpc_jobs/periodic_cleanup.yml
+++ b/rpc_jobs/periodic_cleanup.yml
@@ -9,7 +9,7 @@
             Hours. Instances older than this will be removed.
       - string:
           name: "PROTECTED_PREFIX"
-          default: "long-running-slave|influx-|WEBHOOK-PROXY|nodepool|repo-server-"
+          default: "long-running-slave|influx-|WEBHOOK-PROXY|webhook-proxy|nodepool|repo-server-"
           description: |
             Instances that match this prefix regex will not be cleaned up.
       - string:


rpc_jobs/periodic_cleanup.yml
```diff
@@ -9,7 +9,7 @@
             Hours. Instances older than this will be removed.
       - string:
           name: "PROTECTED_PREFIX"
-          default: "long-running-slave|influx-|WEBHOOK-PROXY|nodepool|repo-server-"
+          default: "long-running-slave|influx-|WEBHOOK-PROXY|webhook-proxy|nodepool|repo-server-"
```
git-harry: Where did `"webhook-proxy"` come from? Are there any called `"WEBHOOK-PROXY"`? Looking at the code it is never defined in lower case
```
$ git grep -i webhook-proxy
pipeline_steps/webhooks.groovy:  String instance_name = "WEBHOOK-PROXY"
rpc_jobs/periodic_cleanup.yml:          default: "long-running-slave|influx-|WEBHOOK-PROXY|nodepool|repo-server-"
rpc_jobs/webhook.yml:          INSTANCE_NAME: "WEBHOOK-PROXY"
```
If this is only a temporary fix that is fine (assuming this instance is supposed to exist, although wouldn't running the job manually suffice, but I think we need to understand how this instance was created and ensure we are keeping our naming consistent. Regarding the use of instance locking, that seems to be duplicating/making-redundant the use of this protected string, maybe we should consider reworking this code to remove `"PROTECTED_PREFIX"` or enable this code to delete locked instances.

mattt416: I don't know @git-harry -- the instance was created on 2017-12-21 when most of us were out.  This is really just a stop-gap until @hughsaunders can respond.

I think you are right in that the periodic cleanup tooling needs updating to handle the situation where a locked instance exists and a deletion on it is attempted, but I still think there is value having both locked instances and `PROTECTED_PREFIX`.  The former guards against accidental deletion outside of the scope of the cleanup job, while the latter ensures we're not wasting time trying to delete instances that we know are permanent.

hughsaunders: Its not defined in lower case anywhere, because I wanted to carry on using WEBHOOK-PROXY, but [setup_openstack_instances](https://github.com/rcbops/rpc-gating/blob/390ceefe903147c924eba25917b0cbc50824ba1f/playbooks/setup_openstack_instances.yml#L65) insists on lowercasing the instance name, which caused the new instance to be created.

hughsaunders: This is where the change that introduced the new instance, its not merged yet: https://github.com/rcbops/rpc-gating/compare/RE-1131#diff-d8c0ac5cddecba3f2bad8f8e531dcf8cR3

hughsaunders: The rax api doesn't seem to return lock status with instance info :( https://developer.rackspace.com/docs/cloud-servers/v2/api-reference/svr-basic-operations/#show-server-details

So maybe cathing exceptions may be the best way to determine if an instance was locked :(

mattt416: Created a separate commit [here](https://github.com/rcbops/rpc-gating/pull/580) for that, @hughsaunders .

hughsaunders: Set metadata fails for a locked instance, so this could be a good test of lock status.

PR #580, opened 2018-01-02
RE-1225 Don't error when deleting locked instances

mattt416:
Issue: [RE-1225](https://rpc-openstack.atlassian.net/browse/RE-1225)

### DIFF
--- a/scripts/periodic_cleanup.py
+++ b/scripts/periodic_cleanup.py
@@ -123,7 +123,22 @@ class Cleanup:
             if errored or age > max_age:
                 _indp("Deleting {name} Errored: {error} Age: {age}".format(
                     name=server.name, error=errored, age=age))
-                self.conn.compute.delete_server(server.id)
+                try:
+                    self.conn.compute.delete_server(server.id)
+                except openstack.exceptions.HttpException as e:
+                    details = "Instance {id} is locked".format(
+                        id=server.id
+                    )
+                    if e.details == details:
+                        _indp(
+                            "Failed to delete server: {id},"
+                            " error: {err}".format(
+                                id=server.id,
+                                err=e.details
+                            )
+                        )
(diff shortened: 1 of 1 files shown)


mattt416:
This hasn't been tested yet -- waiting for all gates to pass first.

mattt416:
Tested [here](https://rpc.jenkins.cit.rackspace.net/job/Periodic-Cleanup/6420/).

hughsaunders:
Alternative:
```
    @log
    def cleanup_instances(self):
        """Cleanup Instances.

        Delete instances if they are in an error state or are over the
        age defined in INSTANCE_AGE_LIMIT.

        Instances that don't match PROTECTED_PREFIX are cleaned up
        """
        current_time = datetime.datetime.now(tzutc())
        max_age = datetime.timedelta(hours=self.age_limit)

        for server in self.unprotected_servers:
            created_time = dateutil.parser.parse(server.created_at)
            age = current_time - created_time
            errored = server.status == "ERROR"
            locked = False
            try:
                server.set_metadata(self.conn.session,
                                    periodic_cleanup="scanned")
            except openstack.exceptions.HttpException as e:
                if "is locked" in e.details:
                    locked = True
                else:
                    raise e
            if (errored or age > max_age) and not locked:
                _indp("Deleting {name} Errored: {error} Age: {age}".format(
                    name=server.name, error=errored, age=age))
                self.conn.compute.delete_server(server.id
```

mattt416:
Closing this PR.  When we're using nodepool, we can update periodic_cleanup.sh to only clean up instances w/ the nodepool prefix, and have the job fail if it finds any instances that are unlocked but fall outside of the nodepool prefix.  Additionally, we're going to create a separate card to implement immediately which will create a card when builds of this periodic cleanup job fail.

PR #581, opened 2018-01-03
RE-1227 WIP DO NOT MERGE

mattt416:


mattt416:
recheck_cit_re_unit_tests

mattt416:
recheck_cit_unit_tests

mattt416:
Most changes here are a result of re-indenting as code is moved inside the try/catch block :(  Can I get some eyes on this before I test these changes?

job_dsl/dep_update.groovy
```diff
@@ -3,56 +3,70 @@
 // Alternative is to use parameters with JJB vars as the defaults.
 library "rpc-gating@${RPC_GATING_BRANCH}"
 common.shared_slave(){
```
git-harry: Why not include this in the try block?

mattt416:
@git-harry yeah was kind of thinking the same, and perhaps this could fit nicely into common.globalWraps()?

git-harry:
@mattt416 sounds reasonable to me.

mattt416:
recheck_cit_unit_tests

mattt416:
recheck_cit_xenial_lint_test

mattt416:
recheck_cit_unit_tests

job_dsl/dep_update.groovy
mattt416: shared_slave(), internal_slave(), etc. all call use_node() which does its own error handling, including cleanup.  The cleanup will remove the venv which we need to create the JIRA issue. We could update the catch to pull down the venv if it doesn't exist, but that's just adding potentially more and more time to job's execution time.

mattt416:
@git-harry squashed down commits, and have created a dummy locked instance to test this periodic_cleanup against tomorrow.  Please review if you're able to and I'll provide links to test as they're available.

rpc_jobs/jjb_setup.yml
```diff
@@ -40,43 +40,52 @@
...
+                  jenkins-jobs --conf jenkins_jobs.ini \
+                               --user $JENKINS_USER \
+                               --password $JENKINS_API_PASSWORD \
+                               --ignore-cache \
+                               \$JJB_ARGS update \$UPDATE_ARGS $JOBS
+                """
+              }
             }
           }
+        } catch(e) {
```
git-harry: This job is frequently run manually when testing a new/modified job, we probably don't want to log the failures from those jobs. I think the if block starting on line 69 is an attempt to identify the builds that were started automatically, maybe the same logic could be used here?

mattt416: Good catch, updated to avoid creating cards when we're running against a specific branch or job.

mattt416:
Periodic-Cleanup change tested [here](https://rpc.jenkins.cit.rackspace.net/job/Periodic-Cleanup/6586/).

mattt416:
recheck_cit_unit_tests - single use slave failure

mattt416:
Tested Jenkins-Job-Builder [here](https://rpc.jenkins.cit.rackspace.net/job/Jenkins-Job-Builder/251/) -- I can't think of a way to make it fail without adjusting the job itself, but I wanted to make sure we didn't have a regression at least (which we did originally, because the JJB conf file was indented incorrectly).

mattt416:
Tested a (read-only) Dep-Update_rpc-openstack-newton-PM run [here](https://rpc.jenkins.cit.rackspace.net/job/Dep-Update_rpc-openstack-newton-PM/15/console).

rpc_jobs/jjb_setup.yml
hughsaunders: L69 is not about detecting builds that were started automatically. It's for detecting builds that use the master branch and are not limited to a subset of jobs, as those are the conditions required for safely removing old jobs.

hughsaunders:
I'll merge this and rebase 574.

PR #582, opened 2018-01-03
RE-1191 Remove usages of the ansicolor plugin

hughsaunders:
This plugin uses Jenkins console notes to store color information,
this is hugely inefficient as it involves storing an encoded java
object in every line.

An alternative colourisation approach is to colourise on the client side
using javascript. This has already been implemented, so we can now
remove the server side implementation.

Issue: [RE-1191](https://rpc-openstack.atlassian.net/browse/RE-1191)


hughsaunders:
@mattt416 yes it has, @bighappyface updated the cloud files container that TES jenkins uses.

PR #583, opened 2018-01-04
RE-1228 Patch openstack inventory script

mattt416:
The openstack.py bundled w/ ansible 2.3.2.0 does not set a hostvar for
ansible_host, and ansible 2.0 deprecated ansible_ssh_host in favour of
ansible_host.  Using the current inventory script causes problems with
stable/pike's version of repo_server since it uses ansible_host and not
ansible_ssh_host.

Issue: [RE-1228](https://rpc-openstack.atlassian.net/browse/RE-1228)


mattt416:
The previous test was w/ the updated inventory, [this](https://rpc.jenkins.cit.rackspace.net/job/Setup-Repo-Servers/38/console) one is with the setting of the fact.

mattt416:
Tested this [here](https://rpc.jenkins.cit.rackspace.net/job/Setup-Repo-Servers/40/console).

PR #591, opened 2018-01-12
RE-1270 Remove PIP from setup_jenkins_slave.yml

mattt416:
As far as I can tell, there's no reason for this to happen any more.

This came to light testing the new rpc-openstack MNAIO jobs -- when run
through jenkins (where we install pip w/ rpc-gating's constraints file)
ansible failed to run with the following error:

```
ContextualVersionConflict: (cryptography 1.2.3 (/usr/lib/python2.7/dist-packages), Requirement.parse('cryptography>=1.5'), set(['paramiko']))
```

When testing manually (without installing pip w/ constraints), a newer
version of setuptools is installed and doesn't cause ansible to throw
that error.  The version of cryptography is still problematic but
ansible seems to be able to handle it internally and works as expected.

Issue: [RE-1270](https://rpc-openstack.atlassian.net/browse/RE-1270)

### DIFF
--- a/playbooks/setup_jenkins_slave.yml
+++ b/playbooks/setup_jenkins_slave.yml
@@ -78,58 +78,6 @@
         line: 'jenkins ALL=(ALL) NOPASSWD: ALL'
       when: "{{ allow_jenkins_sudo }}"
 
-    - name: Copy constraints file over to cloud server
-      copy:
-        src: "{{ lookup('env', 'WORKSPACE') ~ '/rpc-gating/constraints.txt' }}"
-        dest: "/opt/rpc_gating_constraints.txt"
-
-    - block:
-        - name: Get Modern PIP
-          get_url:
-            url: "https://bootstrap.pypa.io/get-pip.py"
-            dest: "/opt/get-pip.py"
-            force: "yes"
-          register: get_pip
-          until: get_pip | success
-          retries: 5
-          delay: 2
-          tags:
-            - pip-install-script
-
-      rescue:
-        - name: Get Modern PIP using fallback URL
-          get_url:
-            url: "https://raw.githubusercontent.com/pypa/get-pip/master/get-pip.py"
-            dest: "/opt/get-pip.py"
-            force: "yes"
-          when: get_pip | failed
-          register: get_pip_fallback
-          until: get_pip_fallback | success
-          retries: 5
-          delay: 2
-          tags:
-            - pip-install-script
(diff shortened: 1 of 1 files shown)


mattt416:
@hughsaunders @odyssey4me @git-harry can you all keep me honest here?

mattt416:
The [following](https://rpc.jenkins.cit.rackspace.net/view/RPC-O/job/PM_rpc-openstack-master-xenial_mnaio-swift-deploy/6/console) was spun up with this change, and bypasses the ansible error that we were seeing in this [build](https://rpc.jenkins.cit.rackspace.net/view/RPC-O/job/PM_rpc-openstack-master-xenial_mnaio-swift-deploy/4/console).

odyssey4me:
@mattt416 We do, at the very least, need virtualenv there. As the slaves setup can either be Xenial or CentOS we'll need to be careful with what we do here but @hughsaunders will have to enlighten a bit more about where exactly we need what.

I can say with confidence that pip 7.1+ supports the ``--constraints`` CLI option and that CentOS does not meet that required version and that's why we normally use get-pip.

In OSA we did a check for the presence of virtualenv at a specific version. If it does not meet it, only then do we install pip using get-pip. See https://github.com/openstack/openstack-ansible/blob/master/scripts/bootstrap-ansible.sh#L117-L133

git-harry:
The code appears to be used by a number of other jobs, such as the Webhook-Setup and Long-Running-Slave jobs. Take a look at https://rpc.jenkins.cit.rackspace.net/job/Long-Running-Slave/24/consoleFull for an example of these tasks being run. I'd want to see all the other types of affected job tested before this is approved.

To me this sounds like a failure of OSA to properly detect that the version installed does not meet its needs. Maybe OSA should be forcing the install of a specific version if validating installed requirements is not possible.

mattt416:
The Long-Running-Slave job ran successfully [here](https://rpc.jenkins.cit.rackspace.net/job/Long-Running-Slave/26/).

mattt416:
recheck_cit_unit_tests

mattt416:
recheck_cit_unit_tests

mattt416:
recheck_cit_unit_tests

mattt416:
recheck_cit_unit_tests

mattt416:
recheck_cit_unit_tests

mattt416:
recheck_cit_unit_tests

PR #600, opened 2018-01-18
[CEPHSTORA-53] Add keystone/rgw integration test

andymcc:
Issue: [CEPHSTORA-53](https://rpc-openstack.atlassian.net/browse/CEPHSTORA-53)


andymcc:
@mattt416 sounds good - would need one PR on the rpc-ceph side to fix up the reference but its not a hard one.

PR #605, opened 2018-01-19
RE-1277 Move logstash role to standard template

mattt416:
This commit creates a new file for rpc-role-logstash and uses the
standard pre and post gating templates.

A separate commit will be created to remote this role from
irr_role_test.yml.

Issue: [RE-1277](https://rpc-openstack.atlassian.net/browse/RE-1277)


mattt416:
Please allow https://github.com/rcbops/rpc-role-logstash/pull/24 to merge before this merges.

PR #612, opened 2018-01-24
RE-1276 Add rpc-maas projects using standard jobs

git-harry:
These new jobs are designed to replace the IRR jobs for rpc-maas. Once
these jobs have been added the equivalent IRR jobs will be removed in a
separate commit.

Issue: [RE-1276](https://rpc-openstack.atlassian.net/browse/RE-1276)

### DIFF
--- a/rpc_jobs/rpc_maas.yml
+++ b/rpc_jobs/rpc_maas.yml
@@ -11,10 +11,12 @@
     scenario:
       - master:
           FLAVOR: "performance2-15"
+          TRIGGER_PR_PHRASE_ONLY: true
       - pike:
           FLAVOR: "general1-8"
       - ocata:
           FLAVOR: "general1-8"
+          TRIGGER_PR_PHRASE_ONLY: true
       - newton:
           FLAVOR: "general1-8"
       - ceph:
@@ -38,7 +40,8 @@
       - newton
       - mitaka
       - liberty
-      - kilo
+      - kilo:
+          TRIGGER_PR_PHRASE_ONLY: true
     action:
       - deploy:
           FLAVOR: "general1-8"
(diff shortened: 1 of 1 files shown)


rpc_jobs/rpc_maas.yml
```diff
@@ -0,0 +1,43 @@
...
+    series: "master"
+    branches:
+      - "master.*"
+    image:
+      - xenial:
+          IMAGE: "Ubuntu 16.04 LTS (Xenial Xerus) (PVHVM)"
+    scenario:
+      - master
+      - pike
+      - ocata
```
odyssey4me: In an effort to save some resources, I suggest adding ``TRIGGER_PR_PHRASE_ONLY: true`` to ocata, pike & master as they are not required PR's in the repo.

rpc_jobs/rpc_maas.yml
```diff
@@ -0,0 +1,43 @@
...
+      - newton
+      - ceph
+    action:
+      - deploy:
+          FLAVOR: "performance2-15"
```
odyssey4me: The original jobs use "general1-8" for all scenario's except for master which uses "performance2-15". Can this be adjusted to do the same in order to save some resourcing.

git-harry: Done.

Good spot, I don't remember doing this intentionally.

rpc_jobs/rpc_maas.yml
git-harry: I've done this for master, ocata and kilo. I think pike should be required given the current state of rpc-openstack so I'll make that so.

git-harry:
@mattt416 I've added a couple of requests for reviews.

@cloudnull / @tonytan4ever this change is to move from using IRR to the standard jobs, mostly it is a like for like swap however given that a few of the jobs are not required by rpc-maas I've made them require a manual trigger. Please will one of you give this a review so that someone working on that repository has verified the change.

PR #615, opened 2018-01-26
RE-1169 Standardise ops-fabric-chatbot jobs

mattt416:
NOTE: The junit change in common.groovy is necessary because at present
no job output is being parsed.  It looks like this pipeline helper only
accepts a relative path, so we cd into the directory and then pass in
'*.xml'.

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -326,7 +326,9 @@ def rpco_archive_artifacts(String build_type = "AIO"){
 def archive_artifacts(){
   stage('Compress and Publish Artifacts'){
     if (env.RE_HOOK_RESULT_DIR != null){
-      junit allowEmptyResults: true, testResults: "${env.RE_HOOK_RESULT_DIR}/*.xml"
+      dir(env.RE_HOOK_RESULT_DIR) {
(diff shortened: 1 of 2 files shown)


mattt416:
recheck_cit_xenial_lint_test

mattt416:
recheck_cit_unit_tests

mattt416:
recheck_xenial_lint_test

git-harry:
recheck_xenial_functional_test - build data no longer exists

pipeline_steps/common.groovy
```diff
@@ -326,7 +326,9 @@ def rpco_archive_artifacts(String build_type = "AIO"){
 def archive_artifacts(){
   stage('Compress and Publish Artifacts'){
     if (env.RE_HOOK_RESULT_DIR != null){
-      junit allowEmptyResults: true, testResults: "${env.RE_HOOK_RESULT_DIR}/*.xml"
+      dir(env.RE_HOOK_RESULT_DIR) {
+        junit allowEmptyResults: true, testResults: '*.xml'
+      }
```
git-harry: There was an issue using `dir` within a Docker slave but based on the notes and the logs that appears to have been fixed.

PR #624, opened 2018-01-30
RLM-1308 Prevent Jira Tickets on Kilo/Mitaka

antonym:
Prevent Jira Tickets on Kilo/Mitaka builds.

Issue: [RLM-1308](https://rpc-openstack.atlassian.net/browse/RLM-1308)

### DIFF
--- a/rpc_jobs/rpc_upgrades.yml
+++ b/rpc_jobs/rpc_upgrades.yml
@@ -54,7 +54,6 @@
     scenario:
       - "swift"
     action:
-      - "mitaka_to_newton_major"
       - "liberty_to_newton_leap"
       - "r12.2.8_to_newton_leap"
       - "r12.2.5_to_newton_leap"
@@ -80,6 +79,21 @@
         action: "kilo_to_newton_leap"
 
 - project:
+    name: "rpc-upgrades-aio-newton-head-pm-no-jira"
+    repo_name: "rpc-upgrades"
+    repo_url: "https://github.com/rcbops/rpc-upgrades"
+    image:
+      - trusty_aio:
+          FLAVOR: "performance2-15"
+          IMAGE: "Ubuntu 14.04.5 LTS prepared for RPC deployment"
+    scenario:
+      - "swift"
+    action:
+      - "mitaka_to_newton_major"
+    jobs:
+      - 'PM_{repo_name}-{branch}-{image}-{scenario}-{action}'
+
+- project:
     name: "rpc-upgrades-aio-newton-release-pm"
     repo_name: "rpc-upgrades"
     repo_url: "https://github.com/rcbops/rpc-upgrades"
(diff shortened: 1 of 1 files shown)


hughsaunders:
@antonym Why do you want to run the jobs if you don't want to be notified about failures? Why not remove the jobs? If a job falls over in a forest and noone is around to hear it, was there any point in running it?

antonym:
@hughsaunders Kilo and Mitaka isn't our primary focus right now but we still want to continue moving forward on them and tune as we have spare cycles.  We plan on flipping on notifications when we're ready to turn our full focus onto them. @claco

mattt416:
I have to agree w/ @hughsaunders here ... it feels a bit pointless spinning up 14 separate performance2-15 instances every single day and having them run for upwards of 4 hours each when they're not actively being monitored.  Do we want to consider removing these jobs and adding them back in when these specific upgrades are the focus of your work?  /cc @claco

mattt416:
recheck_all

claco:
I guess my opinion is one of job ownership. If you've given teams the option to disable it under the auspices of reducing noise during dev, and you want teams to be responsible for their jobs, then it's the job owners decision ultimately, not the infrastructure providers responsibility to worry if we footgun ourselves. In this case, we've added the jobs to evaluate where things stand so we can break down the work needed from there.

If it really does only add comments to 1 existing job, it's not as bad as it was, but the JIRA noise is distracting for me. I'll let others weigh in on that bit.

git-harry:
@claco I agree, we have explicitly pushed responsibility onto the project owners with regards to what are appropriate jobs but we are still responsible for the overall platform. I would agree with the others that there seems little value in creating all these jobs if they not being monitored, especially given the fact that use of these resources isn't free and capacity is finite. If the idea is to work on this in a limited fashion, along the lines of what @mattt416 said, I would suggest removing them and adding them incrementally. As each job is confirmed to work add a new one. This way it is easier to keep on to of the failures and reduced alerts and resources are saved until they are actually needed.

mattt416:
Correct -- my concern is only from a cost/infrastructure standpoint.  I think we need to be mindful of our usage, especially as more and more teams onboard to using the platform.  14 15.5 GB instances, running for 4 hours each every single day is substantial usage when the results aren't necessarily being tracked.

antonym:
Do we have the ability to specify how often the jobs run?  We might not need a daily run of some of these jobs but might get away with a larger interval?  That would probably enable us to save resources.

odyssey4me:
@antonym Byt the looks of https://github.com/rcbops/rpc-gating/blob/master/rpc_jobs/standard_job_postmerge.yml#L67 it's hard-coded in the job right now, but I think that's a very good idea.

antonym:
Yeah, I'm thinking we could probably reduce some of the upgrade jobs to even weekly, just as long as we can get a good history of what's passing and failing so when support wants to run a scenario, they can look at the upgrades dashboard and see what the history is of that scenario and ensure it's passing.

antonym:
Opened: https://rpc-openstack.atlassian.net/browse/RE-1313 to see if we can get support added for altering the frequency of jobs.

If we can tune that, that'll let us reduce the noise and resources being used to run rpc-upgrades jobs.  Will close this out in favor of that option.

PR #625, opened 2018-01-31
RE-1164 Add rpc-maas jobs to replace Travis CI

git-harry:
This change adds the jobs necessary to stop using Travis CI with
rpc-maas and is part of the work to move all projects to using the
standard jobs.

Issue: [RE-1164](https://rpc-openstack.atlassian.net/browse/RE-1164)


git-harry:
These jobs have been built and tested with https://github.com/rcbops/rpc-maas/pull/432

PR #626, opened 2018-01-31
[RO-3870] Change the PR job to non-artifacted

cloudnull:
The default PR test needs to be non-artifacted. This change updates the
PR image properties so that non-artifacted jobs can run as the default.

Issue: [RO-3870](https://rpc-openstack.atlassian.net/browse/RO-3870)

### DIFF
--- a/rpc_jobs/rpc_openstack.yml
+++ b/rpc_jobs/rpc_openstack.yml
@@ -13,8 +13,8 @@
       | ^gating/post_merge_test/
       | ^gating/update_dependencies/
     image:
-      - xenial:
-          IMAGE: "Ubuntu 16.04.2 LTS prepared for RPC deployment"
+      - xenial_no_artifacts:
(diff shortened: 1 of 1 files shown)


rpc_jobs/rpc_openstack.yml
```diff
@@ -174,7 +174,6 @@
     branch: "master"
     jira_project_key: "RO"
     image:
-      - xenial_mnaio
```
prometheanfire: why was this removed?

odyssey4me: Strict artifact builds using MNAIO may or may not work depending on when upstream packages are released versus the last successful build because the MNAIO builds deploy the base OS using the latest upstream packages, not the artifact repo. As such, this test is a bit pointless until we get the base OS deployment working with the artifacts.

prometheanfire: ok

PR #627, opened 2018-01-31
RE-1169 Specify container base image

mattt416:
In most cases, it's irrelevant which base image we use for creating a
container, however an ops-fabric test requires an ubuntu:17.10
container and subsequent tests call python3.6, which is not available
in the standard ubuntu base container image we use for all tests.

This review introduces the parameter `SLAVE_CONTAINER_BASE_IMAGE`,
allowing projects to specify which Ubuntu-flavored container base
image is used. Additionally, we duplicate the Dockerfile to
Dockerfile.standard_job because when the ARG is in the default
Dockerfile all jobs that use it which do not explicitly pass in
BASE_IMAGE will fail. This means this specific job will be impossible
to gate unless we manually update the docker call in Build-Gating-Venv.

NOTE: We remove the systemd-detect-virt call in
rpc_jobs/unit/single_use_slave.yml since systemd does not appear to be
installed on ubuntu:17.10.  Instead, we check for the existence of the
/.dockerenv file to determine if we're inside a container. This may not
be the most reliable long-term test since Docker may remove this file,
however for now it seems to work.

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -937,8 +937,12 @@ void standard_job_slave(String slave_type, Closure body){
         body()
       }
     } else if (slave_type == "container"){
+      String image_name = env.BUILD_TAG.toLowerCase()
+      // NOTE(mattt): This is necessary to initially gate this change as we
+      // have no parameter for SLAVE_CONTAINER_BASE_IMAGE defined yet
+      String base_image = env.SLAVE_CONTAINER_BASE_IMAGE ? env.SLAVE_CONTAINER_BASE_IMAGE : "ubuntu"
       dir("rpc-gating"){
-        container = docker.build env.BUILD_TAG.toLowerCase()
+        container = docker.build(image_name, "--build-arg BASE_IMAGE=${base_image} -f Dockerfile.standard_job .")
       }
       container.inside {
         configure_git()
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -940,7 +940,7 @@ void standard_job_slave(String slave_type, Closure body){
       String image_name = env.BUILD_TAG.toLowerCase()
       // NOTE(mattt): This is necessary to initially gate this change as we
       // have no parameter for SLAVE_CONTAINER_BASE_IMAGE defined yet
-      String base_image = env.SLAVE_CONTAINER_BASE_IMAGE ? env.SLAVE_CONTAINER_BASE_IMAGE : "ubuntu"
+      String base_image = env.SLAVE_CONTAINER_BASE_IMAGE ? env.SLAVE_CONTAINER_BASE_IMAGE : "ubuntu:16.04"
       dir("rpc-gating"){
         container = docker.build(image_name, "--build-arg BASE_IMAGE=${base_image} -f Dockerfile.standard_job .")
       }
--- a/rpc_jobs/unit/single_use_slave.yml
+++ b/rpc_jobs/unit/single_use_slave.yml
@@ -16,6 +16,7 @@
       - rpc_gating_params
       - standard_job_params:
           SLAVE_TYPE: "instance"
+          SLAVE_CONTAINER_BASE_IMAGE: "ubuntu"
       - string:
           name: STAGES
           default: "Allocate Resources, Connect Slave, Cleanup, Destroy Slave"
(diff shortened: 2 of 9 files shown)


mattt416:
recheck_cit_xenial_lint_test

mattt416:
recheck_cit_xenial_lint_test

mattt416:
recheck_cit_xenial_lint_test

mattt416:
recheck_all

mattt416:
TODO: Add a test for this after #633 merges.

mattt416:
recheck_cit_unit_tests

mattt416:
This should be good to review now.  Please consider if we want to add this functionality in addition to the "bring your own Dockerfile" in https://github.com/rcbops/rpc-gating/pull/632.  Thanks!

mattt416:
recheck_cit_unit_tests

rpc_jobs/unit/single_use_slave.yml
```diff
@@ -31,22 +32,39 @@
...
+              assert inside_container == "no"
+            } else {
+              assert inside_container == "yes"
+            } // if
+          } // stage
+
+          stage("Ensure container distro matches env.SLAVE_CONTAINER_BASE_IMAGE") {
+            if (env.SLAVE_TYPE == "container") {
+              if (env.SLAVE_CONTAINER_BASE_IMAGE == "ubuntu") {
+                assert distro == "16.04"
```
hughsaunders: We should probably use xenial rather than ubuntu as the docker image tag, then this won't break when the next lts appears.

mattt416: Done!

mattt416:
recheck_all

pipeline_steps/common.groovy
```diff
@@ -937,8 +937,12 @@ void standard_job_slave(String slave_type, Closure body){
         body()
       }
     } else if (slave_type == "container"){
+      String image_name = env.BUILD_TAG.toLowerCase()
+      // NOTE(mattt): This is necessary to initially gate this change as we
+      // have no parameter for SLAVE_CONTAINER_BASE_IMAGE defined yet
+      String base_image = env.SLAVE_CONTAINER_BASE_IMAGE ? env.SLAVE_CONTAINER_BASE_IMAGE : "ubuntu:16.04"
       dir("rpc-gating"){
-        container = docker.build env.BUILD_TAG.toLowerCase()
+        container = docker.build(image_name, "--build-arg BASE_IMAGE=${base_image} -f Dockerfile.standard_job .")
```
git-harry: How about we just combine the two approaches and creating something a bit more general, for example if we define the following:
SLAVE_CONTAINER_DOCKERFILE_REPO: ["RE", "PROJECT"]
SLAVE_CONTAINER_DOCKERFILE_PATH: file path to dockerfile relative to repo
SLAVE_CONTAINER_DOCKERFILE_BUILD_ARGS

this code could look something like:
```

if (env.SLAVE_CONTAINER_DOCKERFILE_REPO == "RE") {
    dockerfile_repo_dir = "${env.WORKSPACE}/rpc-gating"
} elif (env.SLAVE_CONTAINER_DOCKERFILE_REPO == "PROJECT") {
    // If pr clone
        common.clone_with_pr_refs("${{env.WORKSPACE}}/${{env.RE_JOB_REPO_NAME}}")
    // else clone branch
        ...
    dockerfile_repo_dir = "${{env.WORKSPACE}}/${{env.RE_JOB_REPO_NAME}}"
} else {
    BOOM!
}
dir(dockerfile_repo_dir) {
    container = docker.build(image_name, "${SLAVE_CONTAINER_DOCKERFILE_BUILD_ARGS} -f ${SLAVE_CONTAINER_DOCKERFILE_PATH} .")
}
```

mattt416: This seems like an entirely reasonable suggestion to me and I can't think of any reason why this wouldn't work.  @hughsaunders this look good to you before I test this out?

hughsaunders: 👍

mattt416: @git-harry is there any reason why we can't just merge `SLAVE_CONTAINER_DOCKERFILE_PATH` into `SLAVE_CONTAINER_DOCKERFILE_BUILD_ARGS`?  For example:

```
SLAVE_CONTAINER_DOCKERFILE_BUILD_ARGS="--build-arg BASE_IMAGE=ubuntu:17.10 -f Dockerfile"
```

git-harry: @mattt416 I guess it depends how much control we want to retain. In my mind `SLAVE_CONTAINER_DOCKERFILE_BUILD_ARGS` would actually be a list of args that are then used to build part of the command, something vaguely like:
```
build_args = ""
for arg in SLAVE_CONTAINER_DOCKERFILE_BUILD_ARGS:
    quoted_arg = \\ do groovy equivalent of https://docs.python.org/dev/library/shlex.html#shlex.quote
    build_args += " --build-arg ${quoted_arg}"
```

I'm not sure if there are any options that `docker build` takes that we would explicitly want to exclude but my natural inclination would be to manage what can be set to prevent the unexpected. If this can only be set in the project then maybe reviews of rpc-gating are a sufficiently strong control.

mattt416:
Closing in favour of https://github.com/rcbops/rpc-gating/pull/645.

PR #629, opened 2018-02-01
RE-1306 Detect mainline=rc and fail out

hughsaunders:
This patch also makes the default command more explicit so that
its easier to predict the effect changing each variable will have.

Issue: [RE-1306](https://rpc-openstack.atlassian.net/browse/RE-1306)

PR #633, opened 2018-02-02
RE-1169 Add unit test for container slaves

hughsaunders:
RE-1169 requires upgrading the jenkins docker workflow plugin, to ensure
that the container standard slave type is still functioning after
this change and future changes, an additional unit test is added.

### DIFF
--- a/rpc_jobs/unit/re_unit_tests.yml
+++ b/rpc_jobs/unit/re_unit_tests.yml
@@ -49,15 +49,38 @@
             ]
           )
         },
-        "Single Use Slave": {
+        "Standard Slave Instance": {
           build(
-            job: "RE-unit-test-single-use-slave",
+            job: "RE-unit-test-slave-types",
             wait: true,
             parameters: [
               [
                 $class: 'StringParameterValue',
                 name: 'RPC_GATING_BRANCH',
                 value: env.RPC_GATING_BRANCH
+              ],
+              [
+                $class: 'StringParameterValue',
+                name: 'SLAVE_TYPE',
+                value: 'instance'
+              ]
+            ]
+          )
+        },
+        "Standard Slave Container": {
+          build(
+            job: "RE-unit-test-slave-types",
+            wait: true,
+            parameters: [
+              [
+                $class: 'StringParameterValue',
+                name: 'RPC_GATING_BRANCH',
+                value: env.RPC_GATING_BRANCH
+              ],
+              [
+                $class: 'StringParameterValue',
+                name: 'SLAVE_TYPE',
+                value: 'container'
               ]
             ]
           )
--- a/rpc_jobs/unit/single_use_slave.yml
+++ b/rpc_jobs/unit/single_use_slave.yml
@@ -1,5 +1,5 @@
 - job:
-    name: RE-unit-test-single-use-slave
+    name: RE-unit-test-slave-types
     project-type: workflow
     concurrent: true
     properties:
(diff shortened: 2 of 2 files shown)


hughsaunders:
recheck_unit_tests

hughsaunders:
recheck_unit_tests

hughsaunders:
recheck_xenial_lint_test recheck_xenial_functional_test

lint.sh
```diff
@@ -17,7 +17,7 @@ install(){
     virtualenv $venv
   fi
   . $venv/bin/activate
-  pip install -c constraints.txt -r test-requirements.txt >/dev/null
+  pip install -c constraints.txt -r test-requirements.txt -r requirements.txt >/dev/null
```
mattt416: Hey @hughsaunders , why is this change necessary all of a sudden?

rpc_jobs/unit/single_use_slave.yml
```diff
@@ -14,6 +14,10 @@
           REGIONS: "DFW ORD"
           FALLBACK_REGIONS: "IAD"
       - rpc_gating_params
+      - string:
+          name: SLAVE_TYPE
+          default: container
```
mattt416: You could use the `standard_job_params` params here instead.  The default there is `instance`.

rpc_jobs/unit/re_unit_tests.yml
```diff
@@ -49,15 +49,38 @@
...
               [
                 $class: 'StringParameterValue',
                 name: 'RPC_GATING_BRANCH',
                 value: env.RPC_GATING_BRANCH
+              ],
+              [
+                $class: 'StringParameterValue',
+                name: 'SLAVE_TYPE',
+                value: 'instance'
+              ]
```
mattt416: Ideally we'd want to unset this and allow it to use the default ("instance"), but we'd then need to add something into rpc_jobs/unit/single_use_slave.yml to actually validate we've gotten an instance, and not a container.  Know of any straight-forward way to do that?

lint.sh
hughsaunders: I think its usage of jmespath, I can't remember where it is though.

rpc_jobs/unit/single_use_slave.yml
hughsaunders: Ah good plan.

rpc_jobs/unit/re_unit_tests.yml
hughsaunders: systemd-detect-virt?

mattt416: Brilliant :)

rpc_jobs/unit/single_use_slave.yml
mattt416: Adjusted this.

rpc_jobs/unit/re_unit_tests.yml
mattt416: Updated to do some actual assertions against the output of this.

lint.sh
git-harry: So what is the distinction between requirements and test-requirements?

mattt416: Deferring to @hughsaunders on this one, since he made that specific change.

mattt416: Note this same change is in https://github.com/rcbops/rpc-gating/pull/629/ ... so I don't think this is needed by this specific PR?

git-harry: Why is it here then? Looking at the lint jobs, they are not failing.

mattt416:
recheck_cit_unit_tests

mattt416:

PR #636, opened 2018-02-06
Fix warning for jinja2 brackets

major:
This patch removes the jinja2 brackets from the when argument to
avoid Ansible warnings.

Issue: [RE-1210](https://rpc-openstack.atlassian.net/browse/RE-1210)

### DIFF
--- a/playbooks/setup_jenkins_slave.yml
+++ b/playbooks/setup_jenkins_slave.yml
@@ -76,7 +76,7 @@
         state: present
         regexp: 'jenkins'
         line: 'jenkins ALL=(ALL) NOPASSWD: ALL'
-      when: "{{ allow_jenkins_sudo }}"
+      when: allow_jenkins_sudo | bool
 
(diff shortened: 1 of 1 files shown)


mattt416:
recheck_all

odyssey4me:
recheck_all

hughsaunders:
recheck_all failed on venv build

odyssey4me:
I'm guessing the test are all failing due to the use of a fork here? Should we force merge it?

major:
Would y'all prefer that I not use forks in rpc-gating?

odyssey4me:
/me admits he didn't look at the test results, and walks away in shame...

hughsaunders:
@major The failure was unrelated to your patch. I think I fixed the problem with fork PRs to rpc-gating, but the re-run of these jobs will be a good test of that 😉

major:
recheck_all

PR #637, opened 2018-02-06
RO-3856 Remove legacy RPC-O archive collection

odyssey4me:
With the changes implemented in RPC-O to remove the use
of container artifacts, the job executing the container
builds is failing as it cannot find the appropriate log
folders on the host after the script executes. The script
is currently set to just exit immediately, and therefore
has nothing to collect.

In keeping with the rest of the jobs moving all log
collection to a standard location, then the job picking
up any artifacts if they're there, we remove the legacy
method of collection in this path to get those jobs moving
again and will add the standard log collection method
into the appropriate scripts as necessary. For now there
is actually little value in collecting anything other
than the console log anyway.

Issue: [RO-3856](https://rpc-openstack.atlassian.net/browse/RO-3856)


hughsaunders:
@git-harry mentioned something about using a log collection playbook for mnaio in standup yesterday, is this related? (See RO-3857)

odyssey4me:
@hughsaunders Any log collection should be done using in-repo scripts/playbooks and making use of the appropriate hooks rather than using stuff in rpc-gating so I think this should be removed regardless.

git-harry:
@odyssey4me, as long is there is a card tracking adding log collection to the artefacts jobs, I'm okay with it being removed.

odyssey4me:
@git-harry As noted in the commit message, the only valuable log collection for the artifacts jobs at this point is the console log. If other log collection is required for whatever reason, I'll raise a card for it and implement it then.

git-harry:
@odyssey4me ah okay, I missed that. If the console captures everything required I see no issues. This does highlight that the artefacts job template exists and has not been replaced by a standard job of some form.

odyssey4me:
@git-harry Yeah, I don't think this can make use of standard jobs until the apt artifact build method is changed up to be able to use single-use slaves and execute tests in parallel. For now we're stuck with a special job for this. I have some ideas for improvements which would make this work, but it's not a priority at this stage.

PR #638, opened 2018-02-08
RE-1328 Fix job artifact upload to use the RE hook

odyssey4me:
Currently the job artifact upload assumes the directory
where the artifacts live, rather than making use of the
implemented hook to find out where it is.

This patch make the uploadToSwift function generic,
requiring only a path to find the stuff to upload. The
archive_artifacts function is then changed to make it
take parameters for the paths for the results/artifacts,
but to fall back to a set of default paths if they are
not provided.

The standard jobs are then changed to provide the correct
paths to the function so that the hooks are properly
implemented.

Issue: [RE-1294](https://rpc-openstack.atlassian.net/browse/RE-1294)

Issue: [RE-1328](https://rpc-openstack.atlassian.net/browse/RE-1328)

### DIFF
--- a/pipeline_steps/pubcloud.groovy
+++ b/pipeline_steps/pubcloud.groovy
@@ -237,8 +237,8 @@ def runonpubcloud(Map args=[:], Closure body){
 }
 
 def uploadToSwift(Map args){
-  if(fileExists("${WORKSPACE}/artifacts")){
-    print("WORKSPACE/artifacts directory found, Preparing to upload artifacts.")
+  if (fileExists(args.path)) {
+    print("Directory ${args.path} found. Uploading contents.")
     withCredentials(common.get_cloud_creds()) {
       clouds_cfg = common.writeCloudsCfg(
         username: env.PUBCLOUD_USERNAME,
@@ -248,14 +248,17 @@ def uploadToSwift(Map args){
         common.venvPlaybook(
           playbooks: ["rpc-gating/playbooks/upload_to_swift.yml"],
           vars: [
+            archive_name: args.archive_name,
+            artifacts_dir: args.path,
             container: args.container,
-            description_file: args.description_file
+            description_file: args.description_file,
+            report_dir: args.report_dir
           ]
         ) // venvPlaybook
       } // withEnv
     } // withCredentials
   } else {
-    print("WORKSPACE/artifacts not found, skipping artifact upload")
+    print("Directory ${args.path} not found. Skipping upload.")
   }
 }
 
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -293,25 +293,33 @@ def String gen_instance_name(String prefix="AUTO"){
   return instance_name
 }
 
-def archive_artifacts(){
+def archive_artifacts(Map args = [:]){
   stage('Compress and Publish Artifacts'){
-    if (env.RE_HOOK_RESULT_DIR != null){
-      dir(env.RE_HOOK_RESULT_DIR) {
-        junit allowEmptyResults: true, testResults: '*.xml'
-      }
+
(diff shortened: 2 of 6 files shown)


odyssey4me:
recheck_unit_tests

pipeline_steps/pubcloud.groovy
```diff
@@ -237,8 +237,11 @@ def runonpubcloud(Map args=[:], Closure body){
 }
 
 def uploadToSwift(Map args){
-  if(fileExists("${WORKSPACE}/artifacts")){
-    print("WORKSPACE/artifacts directory found, Preparing to upload artifacts.")
+  if (env.RE_HOOK_ARTIFACT_DIR == null) {
```
mattt416: How about `if (env.RE_HOOK_ARTIFACT_DIR) {` to cater for situations where `env.RE_HOOK_ARTIFACT_DIR` is an empty string?

pipeline_steps/pubcloud.groovy
```diff
@@ -237,8 +237,11 @@ def runonpubcloud(Map args=[:], Closure body){
 }
 
 def uploadToSwift(Map args){
```
git-harry: I would prefer to either have the function take an artefacts directory argument or the function to fail if `env.RE_HOOK_ARTIFACT_DIR` isn't set. I think having a default here creates a potential bug.

odyssey4me:
I've registered https://rpc-openstack.atlassian.net/browse/RE-1328 to continue working on this, rather than the previous card which got me to the point of seeing this particular issue.

hughsaunders:
The environment variable publishes to users of the hooks (scripts) where artifacts should be stored. 
The location is hardcoded (relative to workspace). https://github.com/rcbops/rpc gating/blob/master/rpc_jobs/standard_job_postmerge.yml#L99. Because it's hard coded its safe to use in other parts of the code.

Hook definitions are for users of the platform, not the platform itself. Hooks are also limited to standard jobs which is why this function didn't rely on `env.RE_HOOK_ARTIFACT_DIR` previously.

I agree with @git-harry, make the function take a dir argument, and the standard jobs can then supply `env.RE_HOOK_ARTIFACT_DIR`. Others users of the function can specify their own value, probably `${workspace}/artifacts`. It should still skip if the dir doesn't exist as it's the job's responsibility to create and populate that dir.

odyssey4me:
recheck_unit_tests

odyssey4me:
recheck_unit_tests

odyssey4me:
recheck_unit_tests

odyssey4me:
Unit tests passed in https://rpc.jenkins.cit.rackspace.net/job/RPC-Gating-Unit-Tests/615 which required the artefact unit test job to be rebuilt from this fork.

pipeline_steps/pubcloud.groovy
```diff
@@ -237,8 +237,8 @@ def runonpubcloud(Map args=[:], Closure body){
...
-  if(fileExists("${WORKSPACE}/artifacts")){
-    print("WORKSPACE/artifacts directory found, Preparing to upload artifacts.")
+  if (fileExists(args.path)) {
+    print("Directory ${args.path} found. Uploading contents.")
```
hughsaunders: Whats args.path for, couldn't args.artifact_dir be used instead?

odyssey4me: I was trying to move towards ``uploadToSwift`` be something more generically usable for more than just the artifact upload purpose. I will admit to thinking that perhaps the change of var name makes it just a little harder to follow in the current way this is used though. I can change it up if you'd like.

hughsaunders: @odyssey4me My question isn't so much the name, but rather are both required?

odyssey4me: @hughsaunders ``args.artifact_dir`` is not a valid parameter for the function, unless I rename ``args.path``.

PR #639, opened 2018-02-08
RLM-1368 Move release jobs to r14.current

antonym:
In order to avoid regnerating jobs for monthly releases,
leave the job name generic for the major version and then
determine the latest tag from RPC-O.

Depends on https://github.com/rcbops/rpc-upgrades/pull/126

Issue: [RLM-1368](https://rpc-openstack.atlassian.net/browse/RLM-1368)


antonym:
recheck_all

PR #644, opened 2018-02-13
Jenkins docker service

johnduarte:
This adds a docker configuration to deploy a locally available
Jenkins instance. The jobs for rpc-gating can then deployed to this
instance using jenkins-jobs.

This is a three step process managed by the user and is documented in
the `jenkins_sandbox/README.md` file.

  * Build the Jenkins docker image
  * Run a docker container using the built image
  * Exec a `jenkins-job` command to deploy the rpc-gating jobs to the
  container.

The job deploy can be repeated as necessary while working on jjb code in
order to verify that the resulting job reflects the developers
intentions.

Issue: [ASC-101](https://rpc-openstack.atlassian.net/browse/ASC-101)


jenkins_sandbox/blueocean_jenkins_plugins.txt
```diff
@@ -0,0 +1,102 @@
+matrix-project
```
mattt416: What's this file for here @johnduarte ?

docker-jenkins-sandbox
```diff
@@ -0,0 +1,13 @@
...
+ENV JAVA_OPTS="-Djenkins.install.runSetupWizard=false"
+COPY jenkins_sandbox/plugins.txt /usr/share/jenkins/ref/plugins.txt
+RUN /usr/local/bin/install-plugins.sh < /usr/share/jenkins/ref/plugins.txt
+
+# install jenkins-job-builder
+USER root
+RUN apt-get update
+RUN apt-get install -y python python-dev python-pip
+RUN pip install --upgrade pip
+RUN pip install jenkins-job-builder
```
mattt416: Do you think it'd be worthwhile installing jenkins-job-builder using [constraints.txt](https://github.com/rcbops/rpc-gating/blob/master/constraints.txt) so the version used inside here is the same as we're using in "prod"?

hughsaunders:
This PR may be useful as part of https://rpc-openstack.atlassian.net/browse/RE-1542 which is part of the [Repeatable Jenkins Deploy epic](https://rpc-openstack.atlassian.net/browse/RE-1539).

jenkins_sandbox/plugins.txt
```diff
@@ -0,0 +1,177 @@
+jira-steps
```
hughsaunders: These are all going to need pinned versions. We can generate that from the current master with some groovy.

PR #646, opened 2018-02-15
(ASC-143) Add mnaio system test job

johnduarte:
This commit adds 'system' as a scenario to
`rpc-openstack-master-mnaio-postmerge` project and a filter to ensure
that it only creates a single jobs for
`PM_rpc-openstack-master-xenial_mnaio_no_artifacts-system-deploy`.

This job is will be the proof of concept job for running system level
testing on rpc-openstack. Hooks for taking advantage of the `system`
scenario have not yet been implemented in the `rpc-openstack` repo.

Issue: [ASC-143](https://rpc-openstack.atlassian.net/browse/ASC-143)

### DIFF
--- a/rpc_jobs/rpc_openstack.yml
+++ b/rpc_jobs/rpc_openstack.yml
@@ -179,11 +179,21 @@
       - "ironic"
       - "swift"
     action:
+      - system:
+          IMAGE: "OnMetal - Ubuntu 16.04 LTS (Xenial Xerus)"
+          FLAVOR: "onmetal-io1"
+          REGIONS: "IAD"
+          FALLBACK_REGIONS: ""
       - deploy:
           IMAGE: "OnMetal - Ubuntu 16.04 LTS (Xenial Xerus)"
           FLAVOR: "onmetal-io1"
           REGIONS: "IAD"
           FALLBACK_REGIONS: ""
+    exclude:
+      - action: system
+        image: xenial_mnaio_loose_artifacts
+      - action: system
+        scenario: ironic
     jobs:
       - 'PM_{repo_name}-{branch}-{image}-{scenario}-{action}'
 


odyssey4me:
@johnduarte The 'scenario' axis is meant to be used for applying different configurations, while the 'action' axis is meant for things like 'deploy' or 'perftest' or 'upgrade'. The use of 'system' as a scenario axis does not really give an idea of what the configuration is going to be.

Executing a new set of testing scripts for MNAIO builds doesn't actually need a new test, it just needs to use the already available hooks. Can you give us some insight into where you're hoping to lead with this and we can better advise the forward course.

johnduarte:
Thanks @odyssey4me . At the moment, we are interested in getting a separate experimental up to work out the issues with a new test suite. The final goal is to have the MNAIO deploy execute a separate test runner for the 'system' integration tests. The details for firing this off are still being worked out.
Given what I have seen of the rpc-openstack jobs, that is type of execution path is controlled by one of the axis variables. If there is a better method for controlling such execution, please let me know. The thought behind this PR was that execution would be added to the rpc-openstack post-merge/run step in the following manner:
```
...
if [[ "${RE_JOB_SCENARIO}" == "system" ]]; then
    pushd path/to/new/test/runner
    runner converge
    runner verify
    popd
fi
...
```

For the final job, I agree that is not necessary to re-deploy MNAIO. There are two items I could use a better understanding of.
1. Best way to utilize available hooks for executing test scripts.
2. How to segregate this test job from existing test jobs run on top of MNAIO in order to accomplish the following.
  a. Maintain clean system state between test suites
  b. Allow for discrete testing results by test effort. For example a failure within TEMPEST and SYSTEM suites should be reported independent of each other.

hughsaunders:
@johnduarte 
1. This is mostly handled by the answers to the other questions. You seem to have a good understanding of how the jjb axes interact with the repo scripts. Maybe your use case lines up slightly more with the action axis rather than scenario, but either would work. 

2 a. This is already handled by your current PR - it will generate a new job, and you can modify the rpco scripts to ensure that only your tests run so you have a clean env. If after testing it's too inefficient to have multiple MNAIOs built per merge then we'll have to look at partitioning the resources within the mnaio in order to separate tests. For example if your tests are mostly using openstack resources then a domain could be created for each top level test runner.

2 b. A job can only have one result, but we do have a mechanism for collecting junit tests, which would allow a user to determine whether a tempest or system test failed, assuming the system test report can be converted to junit. However tempest results are not making it to Jenkins currently so that needs investigation. (@git-harry is working on fixing log and test result collection in RO-3857)

odyssey4me:
Yep, I think using the 'action' axis would be best - then in the rpc-openstack repo you'd need a corresponding reaction to the RE_JOB_ACTION environment variable to change the behaviour. This will allow the system tests to be executed against multiple scenario configurations, rather than requiring the duplication of the configuration implementation again.

johnduarte:
If I change this this to use the 'action' axis how do I invoke the MNAIO 'deploy' on the appropriate public cloud infrastructure?

odyssey4me:
@johnduarte Currently the 'deploy' action is not even used - it just assumes that it is going to deploy, unless it happens to be a tox test: https://github.com/rcbops/rpc-openstack/blob/master/gating/pre_merge_test/run#L30-L36

In your case I don't see why there'd need to be any reaction to the new action axis until after the deploy. So for a MNAIO you'd add something into https://github.com/rcbops/rpc-openstack/blob/master/gating/pre_merge_test/run_deploy_mnaio.sh which does whatever you want to do based on the new action.

johnduarte:
Thanks @odyssey4me . I have updated this PR to associate 'system' with the `action` axis and amended the filter accordingly to restrict it to a single job.

git-harry:
Is the intention to manage the test code in rpc-openstack? If not, my understanding was that the rpc-openstack images, currently only available for AIOs, are the intended end goal to facilitate additional layers of testing. If that is the case, maybe it is better to do something like https://github.com/rcbops/rpc-gating/blob/master/rpc_jobs/hardening.yml, the hardening repo calls the rpc-openstack gating scripts to set up the environment which is similar to how images would work.

odyssey4me:
@git-harry You make a good point, although I suspect the job itself makes no sense without the corresponding implementation details in the repo.

However, one could find a similar implementation for the rpc-upgrades repo. It consumes rpc-openstack and runs other things on it: https://github.com/rcbops/rpc-upgrades/blob/master/gating/pre_merge_test/run

johnduarte:
restest this please

johnduarte:
@rpc-jenkins-svc retest this please

rpc_jobs/rpc_openstack.yml
```diff
@@ -179,11 +179,17 @@
       - "ironic"
       - "swift"
     action:
+      - system
```
odyssey4me: As this stands right now, the 'system' action has no image settings, so it inherits the defaults which are hopelessly too low for the MNAIO test. You'll need to copy over the IMAGE, FLAVOR, REGIONS, FALL_BACK_REGIONS settings from the 'deploy' action.

johnduarte:
Thanks @odyssey4me . I have amended the commit to include the same image settings for the `system` action as the `deploy` action.

johnduarte:
https://github.com/rcbops/rpc-openstack-system-tests/pull/2 has been merged.

PR #649, opened 2018-02-21
RI-190: Update Designate Test to include Pike

leader716:
Issue: [RI-190](https://rpc-openstack.atlassian.net/browse/RI-190)


rpc_jobs/rpc_designate.yml
```diff
@@ -1,5 +1,5 @@
  - project:
-    name: "rpc-designate-pre-merge"
+    name: "rpc-designate-master-xenial-pre-merge"
```
mattt416: Hey @leader716, I'd leave this change out as you may end up adding more images down the line.

PR #650, opened 2018-02-21
RE-1169 Docker follow-up changes

mattt416:
In #645, we had set some defaults or handle situations where
parameters being introduced were not yet available on live jobs.

1. In common.groovy, we can now explicitly throw an exception if
   env.SLAVE_CONTAINER_DOCKERFILE_REPO is not set to either "PROJECT"
   or "RE".  The previous structure would allow "RE" to be assumed
   (the default) if env.SLAVE_CONTAINER_DOCKERFILE_REPO hadn't been
   set.
2. In container.groovy, we can remove the default of "./Dockerfile" on
   env.SLAVE_CONTAINER_DOCKERFILE_PATH since a default of
   "./Dockerfile" will be set at the parameter level. We also change
   from using `environ.get` so we can throw a KeyError if one of the
   env variables do not exist.

NOTE: Please allow #645 to merge before this merges.

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -926,8 +926,12 @@ void standard_job_slave(String slave_type, Closure body){
         }
 
         dockerfile_repo_dir += env.RE_JOB_REPO_NAME
-      } else {
+      } else if (env.SLAVE_CONTAINER_DOCKERFILE_REPO == "RE") {
         dockerfile_repo_dir += "rpc-gating"
+      } else {
+        throw new Exception(
+          "SLAVE_CONTAINER_DOCKERFILE_REPO '${env.SLAVE_CONTAINER_DOCKERFILE_REPO}' is not supported."
+        )
       }
 
       dir(dockerfile_repo_dir){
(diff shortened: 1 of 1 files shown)


git-harry:
recheck_unit_tests

pipeline_steps/common.groovy
```diff
@@ -1143,7 +1147,7 @@ from shlex import quote
 
 # We don't want to set a default for build_args as these do not always apply
 build_args = environ.get('SLAVE_CONTAINER_DOCKERFILE_BUILD_ARGS')
-dockerfile = environ.get('SLAVE_CONTAINER_DOCKERFILE_PATH', './Dockerfile')
+dockerfile = environ.get('SLAVE_CONTAINER_DOCKERFILE_PATH')
```
git-harry: What happens if the env var isn't set? What does a build arg of `-f ''` do? It might be worth putting in something more explicit here.

mattt416: Passing in `-f ''` seems to just assume a filename of `Dockerfile`, so I may as well just leave the default here to avoid confusion.

git-harry: I still don't think you want the default there, why would this function be called without `SLAVE_CONTAINER_DOCKERFILE_PATH` being set? If it happened it would be a bug, I'd rather we find out about it with an explicit failure than guess the name of the dockerfile.

mattt416: If we want to be explicit and make specifying `SLAVE_CONTAINER_DOCKERFILE_PATH` an absolute requirement then I can fail if that hasn't been set.  However, would the same not apply to `SLAVE_CONTAINER_DOCKERFILE_BUILD_ARGS`?

git-harry: That one is slightly different because there is no requirement for additional args but we could distinguish been the env var not being set and no args being specified null/''.

We've been caught out by defaults enough times before that I would personally always prefer have as few defaults nestled in the code as possible and keep that stuff in the job definitions/default file.

PR #651, opened 2018-02-22
RE-1329 Compress Logs

hughsaunders:
This commit ensures console logs are zipped when a job terminates.
This will affect all future jobs, current jobs are currentbly being
compressed by a bash loop. This relies on a plugin which has already
been installed.

Issue: [RE-1329](https://rpc-openstack.atlassian.net/browse/RE-1329)

### DIFF
--- a/rpc_jobs/periodic_cleanup.yml
+++ b/rpc_jobs/periodic_cleanup.yml
@@ -117,4 +117,23 @@
             throw e
           }
         } // internal_slave
+        stage("Master Log Compression"){
+          node("master"){
+            // Note that already compressed logs will be skipped as they
+            // get renamed .gz .
+            // This only affects the stage logs used for blue ocean,
+            // console logs are compressed using a plugin activated via
+            // common.globalWraps.
+            // Logs newer than 1 day are skipped to prevent compressing
+            // the logs from running builds.
+            sh """#!/bin/bash
+              cd /var/lib/jenkins/jobs
+              find . -regex '.+/builds/[0-9]+/[0-9]+\\.log' -mtime +1 \\
+                |while read stagelog; do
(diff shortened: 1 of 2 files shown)


hughsaunders:
Created job with DSL
```groovy

properties([compressBuildLog()])
print("test compression plugin")
```
 https://rpc.jenkins.cit.rackspace.net/job/scratchpipeline/830/console
Also verified via cli that the console is compressed.

hughsaunders:
Marked as WIP as this needs a new periodic to compress the blue ocean logs.

hughsaunders:
Introduced second commit which adds a section to periodic cleanup that compresses stage logs.
Ready for review.

rpc_jobs/periodic_cleanup.yml
```diff
@@ -117,4 +117,23 @@
...
+            // Note that already compressed logs will be skipped as they
+            // get renamed .gz .
+            // This only affects the stage logs used for blue ocean,
+            // console logs are compressed using a plugin activated via
+            // common.globalWraps.
+            // Logs newer than 1 day are skipped to prevent compressing
+            // the logs of jobs that are currently running.
+            sh """#!/bin/bash
+              cd /var/lib/jenkins/jobs
+              find . -name \\*.log -mtime +1 \\
```
git-harry: It looks like there are other things, e.g. workspaces, in the job directories in addition to log files. While unlikely, it seem like there is the potential for this to modify a live build. What are your thoughts on something a bit more targetted, e.g. `find . -regex '.+/builds/[0-9]+/[0-9]+\.log' -mtime +1`

hughsaunders:
Updated to include the tighter regex suggested by @git-harry.

PR #654, opened 2018-02-26
RE-1332 Remove artifact jobs from pike.*

mattt416:
@chundu and @mendi have requested the removal of these jobs since pike
is not shipping with any artifacts.

Issue: [RE-1332](https://rpc-openstack.atlassian.net/browse/RE-1332)


mattt416:
@odyssey4me made the requested adjustment.

PR #656, opened 2018-03-01
RE-1310 Release Improvements

hughsaunders:
Two improvements to assist with future releases:

* Ensure that create tag happens before release notes are generated so that the new tag can be compared against the previous tag when generating release notes.

Issue: [RE-1310](https://rpc-openstack.atlassian.net/browse/RE-1310)

### DIFF
--- a/rpc_jobs/release.yml
+++ b/rpc_jobs/release.yml
@@ -53,6 +53,9 @@
                 --repo "${REPO}" \
                 clone \
                     --ref "${RC_BRANCH}" \
+                publish_tag \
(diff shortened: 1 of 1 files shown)


hughsaunders:
@mattt416 you are right that shouldn't be there. At one point I was going to refactor out all of the context usage, but decided against that. Will remove the extra required=True.

hughsaunders:
Updated, ready for review.

hughsaunders:
Removed the rename, now only contains the reordering commit.

PR #657, opened 2018-03-01
RE-36 Adjustable boot timeout & new snapshot jobs

mattt416:
Booting from a saved image is significantly slower than from a base
image, and as such we need to make the timeout configurable.

The default remains 900 seconds, but we can now override this via
BOOT_TIMEOUT.

Additionally, we add some new rpc-designate periodic jobs which use
snapshot images. These jobs are purely for testing and we will look
at adding process and moving all rpc-designate's jobs over to these
images once they are running reliably.

Lastly, we update a comment in the rpc-designate-post-merge project
which was made mention of the pre-merge template and not the
post-merge template.

Issue: [RE-36](https://rpc-openstack.atlassian.net/browse/RE-36)

### DIFF
--- a/playbooks/allocate_pubcloud.yml
+++ b/playbooks/allocate_pubcloud.yml
@@ -9,6 +9,7 @@
     inventory_path: "{{ lookup('env', 'WORKSPACE') }}/rpc-gating/playbooks/inventory"
     user_data_path: "{{ lookup('env', 'WORKSPACE') }}/rpc-gating/scripts/user_data_pubcloud.sh"
     cloud_name: "public_cloud"
+    boot_timeout: "{{ lookup('env', 'BOOT_TIMEOUT') | default(900, true) }}"
   tasks:
 
     # Region should be a CSV string.
@@ -48,7 +49,7 @@
         meta:
           build_config: core
         wait: yes
-        timeout: 900
+        timeout: "{{ boot_timeout }}"
       register: result
       with_items: "{{ regions_shuf + fallback_regions_shuff }}"
       when:
--- a/rpc_jobs/rpc_designate.yml
+++ b/rpc_jobs/rpc_designate.yml
@@ -50,6 +50,36 @@
 
     jira_project_key: "RI"
 
-    # Link to the standard pre-merge-template
+    # Link to the standard post-merge-template
+    jobs:
+      - 'PM_{repo_name}-{branch}-{image}-{scenario}-{action}'
+
+ - project:
+    name: "rpc-designate-post-merge-snapshot"
+
+    repo_name: "rpc-designate"
+    # URL of the rpc-designate repository
+    repo_url: "https://github.com/rcbops/rpc-designate"
+
+    # Use image for RPC-O install
+    image:
+      - xenial_snapshot:
+          REGIONS: "DFW"
+          FALLBACK_REGIONS: ""
+          FLAVOR: "7"
+          BOOT_TIMEOUT: 1500
+
+    scenario:
+      - newton:
+          IMAGE: "rpc-r14.9.0-xenial-swift"
+      - pike:
+          IMAGE: "rpc-r16.2.0-xenial_no_artifacts-swift"
+    action:
+      - "deploy"
+
+    # Sending these failures to RE while we stabilise these images
+    jira_project_key: "RE"
+
+    # Link to the standard post-merge-template
     jobs:
       - 'PM_{repo_name}-{branch}-{image}-{scenario}-{action}'
--- a/rpc_jobs/standard_job_premerge.yml
+++ b/rpc_jobs/standard_job_premerge.yml
@@ -26,6 +26,7 @@
     FLAVOR: "performance1-1"
     IMAGE: "Ubuntu 16.04 LTS (Xenial Xerus) (PVHVM)"
     TRIGGER_PR_PHRASE_ONLY: false
+    BOOT_TIMEOUT: 900
     skip_pattern: ""
     properties:
       - build-discarder:
@@ -84,6 +85,10 @@
       }}
       library "rpc-gating@${{RPC_GATING_BRANCH}}"
       common.globalWraps(){{
+
+        // BOOT_TIMEOUT is used inside an ansible playbook when creating cloud
+        // instances. Builds from snapshots need a significantly longer timeout.
+        env.BOOT_TIMEOUT = "{BOOT_TIMEOUT}"
 
         // Pass details about the job parameters through
         // to the target environment so that scripts can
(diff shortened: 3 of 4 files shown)


mattt416:
recheck_cit_xenial_lint_test

mattt416:
recheck_cit_unit_tests

mattt416:
This will need to merge _after_ https://github.com/rcbops/rpc-openstack/pull/2865 merges otherwise our saved images will get wiped when this runs.

mattt416:
recheck_all

mattt416:
recheck_cit_re_unit_tests

mattt416:
recheck_cit_unit_tests

mattt416:
recheck_cit_unit_tests

playbooks/allocate_pubcloud.yml
```diff
@@ -9,6 +9,7 @@
     inventory_path: "{{ lookup('env', 'WORKSPACE') }}/rpc-gating/playbooks/inventory"
     user_data_path: "{{ lookup('env', 'WORKSPACE') }}/rpc-gating/scripts/user_data_pubcloud.sh"
     cloud_name: "public_cloud"
+    boot_timeout: "{{ lookup('env', 'BOOT_TIMEOUT') | default(900, true) }}"
```
git-harry: What does `true` do? I couldn't find a reference to a second argument in the Ansible documentation for `default`.

rpc_jobs/rpc_designate.yml
```diff
@@ -41,10 +41,17 @@
       - xenial:
           FLAVOR: "performance2-15"
           IMAGE: "Ubuntu 16.04.2 LTS prepared for RPC deployment"
+      - xenial_snapshot:
```
git-harry: The commit does not reference these changes.

rpc_jobs/rpc_designate.yml
```diff
@@ -41,10 +41,17 @@
...
+          REGIONS: "DFW"
+          FALLBACK_REGIONS: ""
+          FLAVOR: "7"
+          BOOT_TIMEOUT: 1500
 
     scenario:
-      - "newton"
-      - "pike"
+      - newton:
+          IMAGE: "rpc-r14.9.0-xenial-swift"
```
git-harry: Who is responsible for incrementing these with each new release and are they aware?

rpc_jobs/rpc_designate.yml
```diff
@@ -41,10 +41,17 @@
       - xenial:
```
git-harry: What happens when `image` is `xenial` and `scenario` is `newton`, which `IMAGE` wins?

rpc_jobs/standard_job_premerge.yml
```diff
@@ -94,6 +95,7 @@
         env.RE_JOB_ACTION = "{action}"
         env.RE_JOB_FLAVOR = "{FLAVOR}"
         env.RE_JOB_REPO_NAME = "{repo_name}"
+        env.BOOT_TIMEOUT = "{BOOT_TIMEOUT}"
```
git-harry: This variable doesn't match with the comment although it does get passed through. Looking more, not all the RE_ variables are collected here anyway so maybe it doesn't matter.

rpc_jobs/rpc_designate.yml
mattt416: Good point -- think I'll just create a new project for these snapshot images instead.

rpc_jobs/rpc_designate.yml
mattt416: I haven't really gotten there yet -- I just want to add some jobs to start seeing how they run on a continual basis.

rpc_jobs/standard_job_premerge.yml
mattt416: Can you clarify which comment you're referring to?  The only reason why I have to export it here is because it gets used subsequently by ansible.

rpc_jobs/rpc_designate.yml
mattt416: Will rewrite commit msg!

rpc_jobs/standard_job_premerge.yml
git-harry:
```
         // Pass details about the job parameters through
         // to the target environment so that scripts can
         // use them to adapt behaviour.
```
It's tacked onto the end of a block for which the comment is indicating they are there for the benefit of gating scripts which this is not.

playbooks/allocate_pubcloud.yml
mattt416: From http://jinja.pocoo.org/docs/dev/templates/#default:

```This will output the value of my_variable if the variable was defined, otherwise 'my_variable is not defined'. If you want to use default with variables that evaluate to false you have to set the second parameter to true```

If I don't add the true, `boot_timeout` defaults to `""`:

```root@aio1:/opt/openstack-ansible/playbooks# cat test.yml
- hosts: localhost
  connection: local
  gather_facts: False
  vars:
    boot_timeout: "{{ lookup('env', 'BOOT_TIMEOUT') | default(900) }}"
  tasks:
    - name: Create primary region list
      debug: var=boot_timeout
root@aio1:/opt/openstack-ansible/playbooks# openstack-ansible test.yml
Variable files: "-e @/etc/openstack_deploy/user_secrets.yml -e @/etc/openstack_deploy/user_variables.yml "
 [WARNING]: Unable to parse /etc/openstack_deploy/inventory.ini as an inventory source


PLAY [localhost] *************************************************************************************************************************************************************************************************************************************************************

TASK [Create primary region list] ********************************************************************************************************************************************************************************************************************************************
Tuesday 06 March 2018  14:58:28 +0000 (0:00:00.226)       0:00:00.226 *********
ok: [localhost] => {
    "boot_timeout": ""
}

PLAY RECAP *******************************************************************************************************************************************************************************************************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=0

Tuesday 06 March 2018  14:58:28 +0000 (0:00:00.043)       0:00:00.270 *********
===============================================================================
Create primary region list -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 0.04s

EXIT NOTICE [Playbook execution success] **************************************
===============================================================================
root@aio1:/opt/openstack-ansible/playbooks#```

Comment (git-harry) [reply playbooks/allocate_pubcloud.yml]: thanks

File: rpc_jobs/rpc_designate.yml [hunk 6]
```diff
@@ -53,3 +53,33 @@
...
+          IMAGE: "rpc-r14.9.0-xenial-swift"
+      - pike:
+          IMAGE: "rpc-r16.2.0-xenial_no_artifacts-swift"
+    action:
+      - "deploy"
+
+    # Sending these failures to RE while we stabilise these images
+    jira_project_key: "RE"
+
+    # Link to the standard pre-merge-template
```
git-harry: This comment is incorrect.

mattt416: Fixed, and updated the incorrect reference from which I copied.

mattt416:
recheck_cit_xenial_lint_test

PR #663, opened 2018-03-12
RE-1359 Adjust gate jobs for Queens/Pike

odyssey4me:
RPC-O master will now become Queens, and Pike will be branched.
This caters for the following changes:

1. Adjust the artifact build jobs to only run against Newton
   and Pike branches. If any Queens artifact work is done then
   the jobs can be adjusted later. Queens will start with no
   artifact support.

2. Adjust the RPC-O jobs for master to only run without artifacts.

3. Adjust all current master/master-rc jobs that are Pike orientated
   to run against the pike/pike-rc branches.

4. There will be no jobs for a master-rc branch until there is a
   Queens RC, so we just rename all the master-rc branch to pike-rc.

Issue: [RE-1359](https://rpc-openstack.atlassian.net/browse/RE-1359)

### DIFF
--- a/rpc_jobs/rpc_openstack.yml
+++ b/rpc_jobs/rpc_openstack.yml
@@ -47,6 +47,54 @@
       - 'PR_{repo_name}-{series}-{image}-{scenario}-{action}'
 
 - project:
+    name: "rpc-openstack-pike-premerge"
+    repo_name: "rpc-openstack"
+    repo_url: "https://github.com/rcbops/rpc-openstack"
+    series: "pike"
+    branches:
+      - "pike.*"
+    skip_pattern: |
+      \.md$
+      | \.rst$
+      | ^releasenotes/
+      | ^gating/generate_release_notes/
+      | ^gating/post_merge_test/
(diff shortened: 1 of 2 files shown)


rpc_jobs/rpc_openstack.yml
```diff
@@ -167,10 +262,10 @@
       - 'PM-push-trigger_{repo_name}-{branch}-{image}-{scenario}-{action}'
 
 - project:
-    name: "rpc-openstack-master-mnaio-postmerge"
+    name: "rpc-openstack-pike-mnaio-postmerge"
```
cloudnull: I think it might make sense to add the pike job and keep the master one, however i've no strong opinions on this; if we need it we could always add it back.

odyssey4me: The master job starts at line 214.

odyssey4me:
@mattt416 There's a PR for master which won't pass without these job changes: https://github.com/rcbops/rpc-openstack/pull/2872

PR #664, opened 2018-03-13
RE-1360 Remove all Newton/Ironic jobs

odyssey4me:
Apparently Ironic on Newton is not supported by RPC-O,
so the jobs are being removed.

Issue: [RE-1360](https://rpc-openstack.atlassian.net/browse/RE-1360)


mattt416:
shocking

PR #667, opened 2018-03-14
RE-1359 Add missing RPC-O Pike PR release test

odyssey4me:
The test to validate that a patch will not break the
release scripts is missing.

Issue: [RE-1359](https://rpc-openstack.atlassian.net/browse/RE-1359)


mattt416:

PR #669, opened 2018-03-19
RE-1130 Add jenkins-slave & Ubuntu Trusty configuration

odyssey4me:
In order to ensure that every nodepool instance is prepared for
jenkins to use it, some preparation of the instance image is
required. This patch adds an diskimage-builder element to do
the preparation, and to configure nodepool to use it.

Also, to ensure that Trusty nodes are available, the configuration
is added to build Trusty images and prepare Trusty nodes.

Issue: [RE-1130](https://rpc-openstack.atlassian.net/browse/RE-1130)

### DIFF
--- a/nodepool/elements/jenkins-slave/extra-data.d/20-jenkins-user
+++ b/nodepool/elements/jenkins-slave/extra-data.d/20-jenkins-user
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+if [ ${DIB_DEBUG_TRACE:-0} -gt 0 ]; then
+    set -x
+fi
+set -eu
+set -o pipefail
+
+# the location to fetch the public keys from
+SSH_PUBLIC_KEYS_URL="https://raw.githubusercontent.com/rcbops/rpc-gating/master/keys/rcb.keys"
+
+# fetch the public keys and put them inside the chroot
+curl --connect-timeout 5 \
+     --retry 3 \
+     ${SSH_PUBLIC_KEYS_URL} > ${TMP_HOOKS_PATH}/ssh-public-keys
--- a/nodepool/elements/jenkins-slave/install.d/20-jenkins-slave
+++ b/nodepool/elements/jenkins-slave/install.d/20-jenkins-slave
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+if [ ${DIB_DEBUG_TRACE:-0} -gt 0 ]; then
+    set -x
+fi
+set -eu
+set -o pipefail
+
+##
+## Add jenkins user, group.
+##
+JENKINS_HOME="/var/lib/jenkins"
+groupadd jenkins
+useradd --gid jenkins \
+        --shell /bin/bash \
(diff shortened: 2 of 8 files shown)


nodepool/elements/jenkins-slave/extra-data.d/20-jenkins-user
```diff
@@ -0,0 +1,15 @@
...
+set -eu
+set -o pipefail
+
+# the location to fetch the public keys from
+SSH_PUBLIC_KEYS_URL="https://raw.githubusercontent.com/rcbops/rpc-gating/master/keys/rcb.keys"
+
+# fetch the public keys and put them inside the chroot
+curl --connect-timeout 5 \
+     --retry 3 \
+     ${SSH_PUBLIC_KEYS_URL} > ${TMP_HOOKS_PATH}/ssh-public-keys
```
dealako: We assume TMP_HOOKS_PATH is defined in the env (??)

nodepool/elements/jenkins-slave/install.d/20-jenkins-slave
```diff
@@ -0,0 +1,61 @@
...
+        --home-dir ${JENKINS_HOME} \
+        --create-home jenkins
+
+##
+## Configure SSH
+##
+# this was copied from outside the chroot by extras.d
+_pub_key=/tmp/in_target.d/ssh-public-keys
+if [ ! -f $_pub_key ]; then
+    die "Can not find public keys!"
```
dealako: Is the `die` function defined/imported anywhere? Unlink python, I think in bash you need to define it each time or in a common script??

nodepool/elements/jenkins-slave/extra-data.d/20-jenkins-user
odyssey4me: ``TMP_HOOKS_PATH`` is a standard path provided by diskimage-builder. It provides a hook path for the extra-data to get something from the host into the chroot. See https://github.com/openstack/diskimage-builder/blob/221c44f24cf398247c9b94e822965b5f6f048f11/doc/source/developer/developing_elements.rst#phase-subdirectories

nodepool/elements/jenkins-slave/install.d/20-jenkins-slave
odyssey4me: The ``die`` function is provided by diskimage-builder. See https://github.com/openstack/diskimage-builder/blob/958ea8a3372157fe62c3a2bb1fcd55a9cadedab1/diskimage_builder/lib/die#L19

nodepool/elements/jenkins-slave/extra-data.d/20-jenkins-user
dealako: 👍

nodepool/elements/jenkins-slave/install.d/20-jenkins-slave
dealako: 👍 cool, thanks for the link.

PR #670, opened 2018-03-20
RE-1372 Provide full ansible role source details

odyssey4me:
Ansible Galaxy is failing to download the 'mongrelion.docker'
role. It would appear that the author has done his first tagged
release and that ansible-galaxy does not allow a role download
that uses a SHA for a version when there is a tagged release.

To ensure that we're able to continue using the role at the
previous version, and to also provide enough information for
ansible-galaxy to fall back when the API service fails, this
patch adds the git source information for each role used.


evrardjp:
The commit message is misleading: If I'm not mistaken, you won't be using API of of galaxy server, just merely using ansible-galaxy CLI. Impact is low for our case: we don't care about the download counts...

evrardjp:
👍

mattt416:
recheck_cit_unit_tests

odyssey4me:
recheck_cit_issue_link
recheck_unit_tests

odyssey4me:
recheck_issue_link

PR #672, opened 2018-03-20
RE-1328 Fix job artifact upload to use the RE hook

odyssey4me:
Currently the job artifact upload assumes the directory
where the artifacts live, rather than making use of the
implemented hook to find out where it is.

This patch make the uploadToSwift function generic,
requiring only a path to find the stuff to upload. The
archive_artifacts function is then changed to make it
take parameters for the paths for the results/artifacts,
but to fall back to a set of default paths if they are
not provided.

The standard jobs are then changed to provide the correct
paths to the function so that the hooks are properly
implemented.

Issue: [RE-1328](https://rpc-openstack.atlassian.net/browse/RE-1328)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -293,25 +293,33 @@ def String gen_instance_name(String prefix="AUTO"){
   return instance_name
 }
 
-def archive_artifacts(){
+def archive_artifacts(Map args = [:]){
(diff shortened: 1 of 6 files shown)


odyssey4me:
Note that the unit test ``Artefact Publication`` will require a job rebuild for the test to pass.

PR #673, opened 2018-03-20
RE-1362 Update nodepool implementation

odyssey4me:
This patch updates the nodepool implementation to the current
versions/SHA's available.

1. 'nodepool_git_dest' and 'diskimage_builder_git_dest' from
   the respective ansible roles now use the value we had as
   a default, so the overrides in our vars are no longer
   necessary.

2. All role and service versions have been updated.

This implementation has been verified to work, including
upgrading it in a like test environment.

Issue: [RE-1362](https://rpc-openstack.atlassian.net/browse/RE-1362)

### DIFF
--- a/nodepool/role_requirements.yml
+++ b/nodepool/role_requirements.yml
@@ -1,16 +1,16 @@
 - name: AnsibleShipyard.zookeeper
   scm: git
   src: https://github.com/AnsibleShipyard/ansible-zookeeper
-  version: 769c5e974a50291e8556b789e57f809d87063262
+  version: "v0.21.0"
 - name: openstack.nodepool
   scm: git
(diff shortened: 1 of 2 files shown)


odyssey4me:
The diskimage-builder SHA used includes a bug which will be resolved when https://review.openstack.org/554771 merges. Holding this patch back until then.

odyssey4me:
The issue fixed by https://review.openstack.org/554771 doesn't seem to be something that will affect us in the short term, so I think we should just go ahead with this version which definitely works for Ubuntu in my testing.

odyssey4me:
I've updated the diskimage-builder release to a version that includes https://review.openstack.org/554771.

mattt416:
SHAmazing

PR #679, opened 2018-03-22
RI-226 Add new project for octavia's pike jobs

mattt416:
Issue: [RI-226](https://rpc-openstack.atlassian.net/browse/RI-226)

### DIFF
--- a/rpc_jobs/rpc_octavia.yml
+++ b/rpc_jobs/rpc_octavia.yml
@@ -1,17 +1,52 @@
  - project:
-    name: "rpc-octavia-pre-merge"
+    name: "rpc-octavia-master-pre-merge"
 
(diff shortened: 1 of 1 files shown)


rpc_jobs/rpc_octavia.yml
```diff
@@ -43,7 +78,7 @@
     # currently we only do master and pike. Once RPC-O switches
     # to Queens we will retire rpc-octavia and use the
     # upstream installer
-    branches:
+    branch:
```
odyssey4me: This needs to be a single value and there needs to be another PM job for the other branch. 'branch' is not a list, but a string.

odyssey4me: Never mind. I'm confused. Branch is an axis.

PR #684, opened 2018-04-05
RE-1397 Add ArtifactBuilder to protected prefix to avoid clean up

odyssey4me:
The artifact builder is a long-lived node, so it should never
be cleaned up.

Issue: [RE-1397](https://rpc-openstack.atlassian.net/browse/RE-1397)

### DIFF
--- a/rpc_jobs/periodic_cleanup.yml
+++ b/rpc_jobs/periodic_cleanup.yml
@@ -9,7 +9,7 @@
             Hours. Instances older than this will be removed.
       - string:
           name: "PROTECTED_PREFIX"
-          default: "long-running-slave|influx-|WEBHOOK-PROXY|webhook-proxy|nodepool|repo-server-"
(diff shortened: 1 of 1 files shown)


rpc_jobs/periodic_cleanup.yml
```diff
@@ -9,7 +9,7 @@
             Hours. Instances older than this will be removed.
       - string:
           name: "PROTECTED_PREFIX"
-          default: "long-running-slave|influx-|WEBHOOK-PROXY|webhook-proxy|nodepool|repo-server-"
+          default: "long-running-slave|influx-|WEBHOOK-PROXY|webhook-proxy|nodepool|repo-server-|ArtifactBuilder"
```
git-harry: I thought the hostnames needed to be lowercase otherwise it causes issues with some tool or another. Is this not affected in the same way as webhook-proxy?

odyssey4me: This is just a jenkins name, not a host name. The actual host name is quite different and is lower case.

PR #689, opened 2018-04-11
RE-1412 Add bzip package install to nodepool elements

odyssey4me:
Without this package, the rpc-gating venv tarball cannot
be extracted.

We also sort the packages installed alphabetically to make
them easier to find.

Issue: [RE-1412](https://rpc-openstack.atlassian.net/browse/RE-1412)

### DIFF
--- a/nodepool/elements/jenkins-slave/package-installs.yaml
+++ b/nodepool/elements/jenkins-slave/package-installs.yaml
@@ -1,3 +1,4 @@
-git-core:
-default-jre-headless:
+bzip2:
 curl:
+default-jre-headless:
+git-core:


odyssey4me:
recheck_unit_tests

odyssey4me:
recheck_unit_tests

odyssey4me:
recheck_unit_tests

odyssey4me:
@mattt416 Yes we do. https://github.com/rcbops/rpc-gating/blob/master/nodepool/nodepool-config.yml#L113-L135

mattt416:
@odyssey4me will this setup work then, considering trusty's default-jre-headless is 1.7, which is not compatible w/ jenkins?

odyssey4me:
recheck_unit_tests

odyssey4me:
@mattt416 Hmm, good point. I guess we'll need to submit a patch to update that. Can you register a card for it and push up a patch?

PR #694, opened 2018-04-11
RE-1190 Remove leading remote name from RC_BRANCH

odyssey4me:
The library being used adds the remote prefix (origin)
to the ref already, so it's unnecessary to use it here.


odyssey4me:
@hughsaunders I did that previously, but that breaks osa_differ. So either this needs to change or osa_differ needs to change.

PR #695, opened 2018-04-11
RI-254 Remove unnecessary PM jobs and adjust timer

mattt416:

1. Removes the ironic PM AIO jobs for master, pike, and pike-rc since
   we have equivalent PM MNAIO jobs for these. At present we have no
   immediate use for ironic snapshot images, so there's no point
   leaving the ironic PM AIO jobs around for this.
2. Sets a @monthly CRON value for the PM swift AIO jobs on master, pike
   and pike-rc so that these jobs are effectively only run when the
   branches are pushed to (again, we have equivalent PM MNAIO jobs for
   these).
3. Similar to #2, we do this for the PM xenial_loose_artifacts newton
   and newton-rc AIO jobs since we have this combintation running as
   PM MNAIO jobs.

Issue: [RI-254](https://rpc-openstack.atlassian.net/browse/RI-254)

### DIFF
--- a/rpc_jobs/rpc_openstack.yml
+++ b/rpc_jobs/rpc_openstack.yml
@@ -193,11 +193,13 @@
     repo_url: "https://github.com/rcbops/rpc-openstack"
     branch: "master"
     jira_project_key: "RO"
+    # We have an MNAIO equivalent of this PM job, but we need to keep
+    # this for the push triggering for snapshot building.
+    CRON: "@monthly"
     image:
       - xenial_no_artifacts:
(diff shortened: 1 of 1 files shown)


mattt416:
Let's get confirmation from RI team on these changes before this merges.

mattt416:
recheck_all

mattt416:
recheck_cit_unit_tests

mendeni:
What's the next step(s) for this PR?
Shall we merge with failed **re_unit_tests** or try to resolve?

mattt416:
recheck_cit_unit_tests

mattt416:
@mendeni Let's just let this job pass and then we can merge.

PR #698, opened 2018-04-12
RE-1418 Fix issue link job when using remote forks

odyssey4me:
The issue link job currently does not work properly when
executing from a fork. While the current mechanism is ok
it can be simplified to use the other existing clone
function which includes refs, and the function to find
the JIRA issues can then use the pr merge ref.

This simplifies the system, negating the need to add more
remotes and ensuring that everything required is in place.

Issue: [RE-1418](https://rpc-openstack.atlassian.net/browse/RE-1418)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -473,7 +473,7 @@ void clone_repo(String directory, String ssh_key, String repo, String ref, Strin
       git remote add origin "${repo}"
       # Don't quote refspec as it should be separate args to git.
       # only log errors
-      git fetch --tags origin ${refspec} > /dev/null
+      git fetch --quiet --tags origin ${refspec}
       git checkout ${ref}
       git submodule update --init
     """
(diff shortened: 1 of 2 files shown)


pipeline_steps/common.groovy
```diff
@@ -618,7 +618,7 @@ def get_jira_issue_key(String repo_path="rpc-openstack"){
     returnStdout: true,
     script: """#!/bin/bash -e
       cd ${repo_path}
-      git log --pretty=%B origin/${ghprbTargetBranch}..origin/${ghprbSourceBranch}""")
+      git log --pretty=%B origin/${ghprbTargetBranch}..origin/pr/${ghprbPullId}/merge}""")
```
hughsaunders: extra `}` after merge

odyssey4me: Great catch. I fell like a banana now. Thanks!

PR #699, opened 2018-04-13
RE-1418 Fix issue link job when using remote forks

odyssey4me:
The issue link job currently does not work properly when
executing from a fork. While the current mechanism is ok
it can be simplified to use the other existing clone
function which includes refs, and the function to find
the JIRA issues can then use the pr merge ref.

This simplifies the system, negating the need to add more
remotes and ensuring that everything required is in place.


odyssey4me:
recheck_issue_link

PR #701, opened 2018-04-13
RE-1384 Add component dependency update support

git-harry:
This change adds support for testing component dependency updates. By
default this feature is not enabled but can be specified in the project
definition with `component_dependencies_update: "true"`. Currently this
feature is still being tested and should not be used outside of that
test work.

This change also allows for the use of update_dependencies/run updates
to be disabled with `third_party_dependencies_update: "false"`.

Issue: [RE-1384](https://rpc-openstack.atlassian.net/browse/RE-1384)

### DIFF
--- a/scripts/dep_update.sh
+++ b/scripts/dep_update.sh
@@ -4,21 +4,37 @@
 # Create pr if required
 # Run post hook
 run_hook(){
-  if [[ -x "$1" ]]
+  if [[ ${THIRD_PARTY_DEPENDENCIES_UPDATE} == "true" ]]
   then
-    $1 || {
-      echo "Hook script failed: $1"
-      exit 1
-    }
-  else
-    if [[ "$1" =~ .*/run ]]
+    if [[ -x "$1" ]]
     then
-      echo "Run hook $1 not found."
-      exit 1
+      $1 || {
+        echo "Hook script failed: $1"
+        exit 1
+      }
+    else
+      if [[ "$1" =~ .*/run ]]
+      then
+        echo "Run hook $1 not found."
+        exit 1
+      fi
     fi
   fi
 }
 cd ${WORKSPACE}/repo
+
+start_sha=$(git rev-parse --verify HEAD)
+
+if [[ "${COMPONENT_DEPENDENCIES_UPDATE}" == 'true' ]]; then
+  apt-get update
+  apt-get install -y python3-pip
+  pip3 install 'git+https://github.com/mattt416/rpc-metadata#rpc_component&subdirectory=rpc_component'
+  component --releases-repo 'https://github.com/mattt416/rpc-metadata' dependency update-requirements
+  if [[ ${start_sha} == $(git rev-parse --verify HEAD) ]]; then
+    echo "No component dependency updates found."
+  fi
+fi
+
 run_hook "gating/update_dependencies/pre"
 run_hook "gating/update_dependencies/run"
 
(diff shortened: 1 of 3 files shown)


scripts/dep_update.sh
```diff
@@ -27,43 +43,47 @@ if [[ "${READ_ONLY_TEST:-true}" == "true" ]]; then
...
+# split URL into array separated by '/'s
+IFS='/ ' read -r -a url_split <<< "${REPO_URL}"
+# owner/org is whatever is between the second and third / in the url
+owner="${url_split[3]}"
+# repo is anything after owner/
+repo=${REPO_URL#https://github.com/${owner}/}
+ssh_url="git@github.com:${owner}/${repo}"
 
-  # Get / Create Jira issue
+if [[ -n "$(git status -s)" ]] || [[ ${start_sha} != $(git rev-parse --verify HEAD) ]]; then
```
hughsaunders: I like this, I hadn't considered that the script might commit its own changes.

scripts/dep_update.sh
```diff
@@ -4,21 +4,37 @@
...
   fi
 }
 cd ${WORKSPACE}/repo
+
+start_sha=$(git rev-parse --verify HEAD)
+
+if [[ "${COMPONENT_DEPENDENCIES_UPDATE}" == 'true' ]]; then
+  apt-get update
+  apt-get install -y python3-pip
+  pip3 install 'git+https://github.com/mattt416/rpc-metadata#rpc_component&subdirectory=rpc_component'
```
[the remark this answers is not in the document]
git-harry: Putting it on PyPI is not of any benefit at the moment, once we have a demo workflow set up and have finalised a first version we can look at distribution. For now PyPi would simply make testing harder.

scripts/dep_update.sh
```diff
@@ -76,8 +96,10 @@ else
     create_pr \
       --source-branch "${pr_branch}"\
       --target-branch "${BRANCH}" \
-      --title "${title}" \
-      --body "${message}"
+      --title "${issue} Update ${BRANCH} dependencies" \
+      --body "Automated dependency update pull request, see individual commits for details."
+else
+  echo "No pull request created, update scripts did discover any dependency changes."
```
mattt416: s/update scripts did/update scripts didn't/ ?

scripts/dep_update.sh
```diff
@@ -27,43 +43,47 @@ if [[ "${READ_ONLY_TEST:-true}" == "true" ]]; then
...
+  echo "Repo is clean, prep script made no changes to be committed."
+else
+  echo "Repo is dirty, committing changes."
+  title="${issue} Update ${BRANCH} third-party dependencies"
+  message="This change is the result of running gating/update_dependencies/*
+  See update_dependencies in https://rpc-openstack.atlassian.net/wiki/spaces/RE/pages/19005457/RE+for+Projects for more details."
   git commit -a -m "${title}" -m "${message}"
+fi
 
+if [[ ${start_sha} != $(git rev-parse --verify HEAD) ]]; then
```
mattt416: Why a separate if block here?  If we commit in the if block above, then this will always be true, right?

git-harry: The component dependency update creates its own commit separately (lines 28-36). The separation is so that a pull request is made regardless of where the change comes from.

scripts/dep_update.sh
git-harry: fixed

mattt416:
recheck_all

PR #702, opened 2018-04-13
RE-1420 Transition rpc-gating tests to use nodepool

odyssey4me:
To increase the usage of nodepool, we transition some low-impact
tests to it.

Given that the 'functional' pre-merge test is duplicated in the
unit tests, we remove that.

Issue: [RE-1420](https://rpc-openstack.atlassian.net/browse/RE-1420)

### DIFF
--- a/rpc_jobs/standard_job_premerge.yml
+++ b/rpc_jobs/standard_job_premerge.yml
@@ -9,9 +9,9 @@
     branches:
       - "master"
     image:
-      - "xenial"
+      - "xenial":
+          SLAVE_TYPE: "nodepool-ubuntu-xenial"
     scenario:
-      - "functional"
       - "lint"
     action:
       - "test"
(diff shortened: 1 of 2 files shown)


odyssey4me:
recheck_unit_tests
recheck_xenial_lint_test

rpc_jobs/standard_job_premerge.yml
```diff
@@ -11,10 +11,10 @@
     image:
       - "xenial"
     scenario:
-      - "functional"
       - "lint"
     action:
-      - "test"
+      - "test":
+          SLAVE_TYPE: "nodepool-ubuntu-xenial"
```
mattt416: Would it be cleaner to place this under the image axis so it directly corresponds to the xenial image?  That way if we add another image we don't have to move it later.

odyssey4me: Sure, can do that if you like.

odyssey4me: Done.

odyssey4me:
recheck_xenial_lint_test

PR #704, opened 2018-04-13
RE-1415 Only install openjdk-8-jre-headless

mattt416:
Since #703 has since merged (and only installs openjdk-8-jre-headless
on both trusty and xenial), this commit has been created to update
setup_jenkins_slave.yml to do the same thing.  This should ensure
that we have consistent nodes irrespective of nodepool or legacy boot
method.

NOTE: We also add a task to set set the default JAVA_HOME to the
installed jdk, otherwise the groovy syntax check fails as groovy looks
for java under /usr/lib/jvm/default-java/bin/java. This doesn't
exist as we no longer install default-jre-headless or run
update-java-alternatives to point the default to
java-1.8.0-openjdk-amd64.

Issue: [RE-1415](https://rpc-openstack.atlassian.net/browse/RE-1415)

### DIFF
--- a/playbooks/setup_jenkins_slave.yml
+++ b/playbooks/setup_jenkins_slave.yml
@@ -58,24 +58,12 @@
       delay: 2
       with_items:
         - git-core
-        - default-jre-headless
-
-    - name: Install openjdk-8-jre-headless
-      apt:
-        pkg: "{{ item }}"
-        state: installed
-        update_cache: yes
-      register: install_packages
-      until: install_packages | success
-      retries: 5
-      delay: 2
-      with_items:
         - openjdk-8-jre-headless
-      when: ansible_distribution_release | lower == 'trusty'
 
-    - name: Set default java to java-1.8.0-openjdk-amd64
-      command: update-java-alternatives -s java-1.8.0-openjdk-amd64
-      when: ansible_distribution_release | lower == 'trusty'
+    - name: Set default JAVA_HOME environment variable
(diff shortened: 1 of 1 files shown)


mattt416:
Testing this [here](https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-newton-trusty-swift-deploy/214/).

mattt416:
Now I remember why I did this, because the cit_xenial_lint_test fails with:

```
10:31:49 groovy: JAVA_HOME is not defined correctly, can not execute: /usr/lib/jvm/default-java/bin/java
10:31:49 Groovy syntax fail
```

odyssey4me:
recheck_xenial_lint_test

odyssey4me:
Interesting. I guess the difference is that in this case, jenkins is already logged into the node - whereas with the nodepool setup, jenkins only logs into the node when it needs it. With nodepool then it receives the right env vars when logging in.

mattt416:
recheck_cit_unit_tests

odyssey4me:
Perhaps we should set the playbook to add the JAVA_HOME env var to /etc/environment, similar to https://github.com/rcbops/rpc-gating/pull/707

odyssey4me:
recheck_xenial_lint_test

odyssey4me:
recheck_xenial_lint_test

mattt416:
@odyssey4me Added a task to update /etc/environment.

PR #707, opened 2018-04-17
RE-1415 Add JAVA_HOME env var to default environment

odyssey4me:
In order to ensure that java applications work properly
we set the environment variable as a default. To ensure
that we use whatever the current location is, we use a
find instead of a static path.

Issue: [RE-1415](https://rpc-openstack.atlassian.net/browse/RE-1415)


odyssey4me:
recheck_unit_tests

PR #711, opened 2018-04-18
RE-1425 RPC Gating Container Options for venv

dealako:
- Updated the build venv logic to perform building, arching AND SSH/copying all within the container.

Issue: [RE-1425](https://rpc-openstack.atlassian.net/browse/RE-1425)


dealako:
@hughsaunders can you confirm that updating the `env.REPO_USER_KEY` and `env.REPO_HOST_PUBKEY` values **DO NOT** have any side effects?

dealako:
@git-harry Looking for examples where this is done _within_ a container which inclues a completely different path for the SSH key(s).  Don't see any.  Specifically, how to specify a static value for `credentials:[blah]`?

git-harry:
@dealako take a look at the diff in my comment, that shows how it can be done. Existing examples are a little less obvious but I believe one would be when `env.SLAVE_TYPE=container` then `stdJob` > `standard_job_slave` > `clone_with_pr_refs` > `clone_external_repo` > `clone_repo` uses `sshagent` within a container.

PR #712, opened 2018-04-18
RE-1222 Disable Build Status Comments in JIRA Issues

dealako:
- Removed call to create build status JIRA comment.

### DIFF
--- a/rpc_jobs/standard_job_premerge.yml
+++ b/rpc_jobs/standard_job_premerge.yml
@@ -191,8 +191,6 @@
               currentBuild.result="FAILURE"
               throw e
             }} finally {{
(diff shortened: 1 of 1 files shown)


odyssey4me:
recheck_all

mattt416:
recheck_all

odyssey4me:
@mattt416 The artifact build job - sure, although it should still happen if the trigger is a PM job, not a PR job.

git-harry:
@odyssey4me the artefact job does not create/comment-on an issue if it was triggered by cron, it only happens for pull requests.

git-harry:
The merge conflict is the result of https://github.com/rcbops/rpc-gating/pull/718. I took this change into account as part of that patch to ensure it wasn't reintroduced. Given that it has merged first, this change is no longer relevant unless it is going to address rpc_artifact_build.

PR #720, opened 2018-04-24
RE-1424 Phobos VPN Service

hughsaunders:
This service adds a job to create a background vpn service that runs
on the Jenkis nodes within the Rackspace firewall.

This is for two purposes:
1. Give the master access to phobos outside the context of a job
so that phobos can be used as a region

2. Give the slave access to phobos, so jobs can use it without
starting their own vpn instance.


hughsaunders:
recheck_issue_link

odyssey4me:
recheck_all

odyssey4me:
recheck_cit_all

odyssey4me:
recheck_cit_all

odyssey4me:
recheck_issue_link

PR #725, opened 2018-04-26
RE-1434 Expose failures compressing logs

git-harry:
The script to compress logs in the periodic_cleanup only registers a
failure in the while loop if it happens on the last pass. This change
ensure that any failures are reported by the first build that detects
them and are consistent between builds.

Issue: [RE-1434](https://rpc-openstack.atlassian.net/browse/RE-1434)

### DIFF
--- a/rpc_jobs/periodic_cleanup.yml
+++ b/rpc_jobs/periodic_cleanup.yml
@@ -125,11 +125,15 @@
             // the logs from running builds.
             sh """#!/bin/bash
               cd /var/lib/jenkins/jobs
-              find . -regex '.+/builds/[0-9]+/\\([0-9]+\\.\\)?log' -mtime +1 \\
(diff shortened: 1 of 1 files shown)


git-harry:
recheck_all

git-harry:
recheck_unit_tests

git-harry:
recheck_unit_tests

git-harry:
The unit test failure looks to be caused by https://github.com/rcbops/rpc-gating/pull/749 and can be ignored.

hughsaunders:
recheck re_unit_tests

hughsaunders:
recheck_unit_tests

hughsaunders:
I was thinking the unit test job is fixed, but it won't work with master at the moment. So happy to merge with the failure with another approval.

PR #731, opened 2018-04-26
RE-1355 Fix dep_update JIRA label

git-harry:
This change fixes the label string to reference the correct variable.
Without this change the label was taking the form `some-repo_` when it
should be `some-repo_a-branch`.

Issue: [RE-1355](https://rpc-openstack.atlassian.net/browse/RE-1355)


git-harry:
recheck_all

PR #744, opened 2018-04-26
RE-1449 Lint XML before running junit step

mattt416:
Currently, if bad XML ends up in the result dir and jenkins parses it
with the junit step, then jenkins will mark the build as UNSTABLE.

To avoid builds failing because of this situation, we first run xmllint
against the XML files and then only run the junit step if no errors
were found.

We also install libxml2-utils (which provides xmllint) in Docker
containers, pubcloud single use slaves, and nodepool instances.

Issue: [RE-1449](https://rpc-openstack.atlassian.net/browse/RE-1449)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -302,7 +302,25 @@ def archive_artifacts(Map args = [:]){
     results_dir = args.get("results_dir", "${env.WORKSPACE}/results")
 
     dir(results_dir) {
-      junit allowEmptyResults: true, testResults: "*.xml"
+      // The junit step doesn't like parsing non-XML (resulting in builds being
+      // marked as UNSTABLE). If xmllint is installed we verify that all XML
+      // input is valid and if not we skip the junit step.
+      Integer xmlStatus = sh(
+        returnStatus: true,
+        script: """#!/bin/bash -xe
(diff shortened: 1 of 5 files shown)


mattt416:
Here's a [test](https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-product-1-master-xenial-functional-test/24/) where this patch is tested against [bad XML](https://github.com/mattt416/rpc-product-1/commit/a0be7b25301434fbbfae492abb313b63a5c2a225).

pipeline_steps/common.groovy
```diff
@@ -302,7 +302,25 @@ def archive_artifacts(Map args = [:]){
...
+          if ls *.xml && [[ -f /usr/bin/xmllint ]]; then
+            /usr/bin/xmllint *.xml
+          else
+            exit 0
+          fi
+        """
+      )
+
+      if (xmlStatus == 0) {
+        junit allowEmptyResults: true, testResults: "*.xml"
```
odyssey4me: This will unfortunately skip the import for all xml files if only one is invalid. Ideally we should check each individual file and execute the import for each one that is OK - only skipping the file that is broken.

This way we can still ingest the ARA results, even if the tempest results are broken.

PR #771, opened 2018-04-27
[WIP] RE-1432 Build Summary

hughsaunders:
This commit adds a job to generate a summary of build failures.
The job runs in a docker container which is built by an
additional job which is also added by this commit.

The summary page is pushed to a unindexed dir on rpc repo.

visualisation of the modified lint filename regex:
screen shot 2018-04-27 at 13 55 07


hughsaunders:
The problem is that the logs are compressed (as Matt warned me about!)

odyssey4me:
recheck_unit_tests

constraints.txt
```diff
@@ -19,6 +19,7 @@ funcsigs==1.0.2
 gitdb2==2.0.2
 github3.py==0.9.6
 GitPython==2.1.5
+humanize==0.5.1
```
hughsaunders: I should probably remove this, as I created separate constraints and requirements files for build-summary

requirements.txt
```diff
@@ -4,10 +4,12 @@ click
 croniter
 github3.py
 gitpython
+humanize
```
hughsaunders: I should probably remove this, as I created separate constraints and requirements files for build-summary

hughsaunders:
Pushed a new version to remove `30 Days` from some of the heads, as retention period varies by job. Also removed unnecessary additions to root requirements/constraints.

Thanks for reviews, will merge once test jobs have run through.

PR #793, opened 2018-05-01
RE-1449 Lint XML before running junit step

mattt416:
Currently, if bad XML ends up in the result dir and jenkins parses it
with the junit step, then jenkins will mark the build as UNSTABLE.

To avoid builds failing because of this situation, we first run xmllint
against the XML files and then only run the junit step if no errors
were found.

We also install libxml2-utils (which provides xmllint) in Docker
containers, pubcloud single use slaves, and nodepool instances.

Issue: [RE-1449](https://rpc-openstack.atlassian.net/browse/RE-1449)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -320,7 +320,7 @@ def archive_artifacts(Map args = [:]){
           Integer xmlStatus = sh(
             returnStatus: true,
             script: """#!/bin/bash -xe
-              /usr/bin/xmllint ${file} || mv ${file} ${file}-broken
+              /usr/bin/xmllint ${file} > /dev/null || mv ${file} ${file}-broken
             """
           )
         }
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -302,7 +302,25 @@ def archive_artifacts(Map args = [:]){
     results_dir = args.get("results_dir", "${env.WORKSPACE}/results")
 
     dir(results_dir) {
-      junit allowEmptyResults: true, testResults: "*.xml"
+      // The junit step doesn't like parsing non-XML (resulting in builds being
+      // marked as UNSTABLE). If xmllint is installed we verify that all XML
+      // input is valid and if not we skip the junit step.
+      Integer xmlStatus = sh(
+        returnStatus: true,
+        script: """#!/bin/bash -xe
+          if ls *.xml && [[ -f /usr/bin/xmllint ]]; then
+            /usr/bin/xmllint *.xml
+          else
+            exit 0
+          fi
+        """
+      )
+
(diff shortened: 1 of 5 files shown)


pipeline_steps/common.groovy
```diff
@@ -302,7 +302,25 @@ def archive_artifacts(Map args = [:]){
...
+          if ls *.xml && [[ -f /usr/bin/xmllint ]]; then
+            /usr/bin/xmllint *.xml
+          else
+            exit 0
+          fi
+        """
+      )
+
+      if (xmlStatus == 0) {
+        junit allowEmptyResults: true, testResults: "*.xml"
```
mattt416: From @odyssey4me in PR #744:

```This will unfortunately skip the import for all xml files if only one is invalid. Ideally we should check each individual file and execute the import for each one that is OK - only skipping the file that is broken.

This way we can still ingest the ARA results, even if the tempest results are broken.```

Comment (hughsaunders) [reply pipeline_steps/common.groovy]: I concur.. test each one, and move broken files into a subdir so that they won't be globbed in `*.xml`.

**Comment by mattt416** [comment]
@hughsaunders @odyssey4me this should be ready for a re-review.  Testing these changes [here](https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-master-xenial_no_artifacts-swift-deploy/).

File: pipeline_steps/common.groovy [hunk 2]
```diff
@@ -302,6 +302,30 @@ def archive_artifacts(Map args = [:]){
...
+      // all XML input is valid, and if not we move the affected files aside so
+      // the junit pipeline step won't parse them.
+      if (testXmlLintRc == 0) {
+        List xmlFiles = findFiles(glob: '*.xml')
+
+        for (file in xmlFiles) {
+          Integer xmlStatus = sh(
+            returnStatus: true,
+            script: """#!/bin/bash -xe
+              /usr/bin/xmllint ${file} || mv ${file} ${file}-broken
```
odyssey4me: Given that the file was originally copied from the artefacts directory, I don't think we need to keep it here - we can just delete it. But meh, this is fine too.

mattt416:
Made a small update because the ara XML is huge and this ends up getting echo'd to the consoleLog. I'm re-testing this [here](https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-master-xenial_no_artifacts-swift-deploy/200/).

mattt416:
@odyssey4me tested this against a commit w/ a "good" and "bad" XML file: https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-product-1-master-xenial-functional-test/48/console

odyssey4me:
@mattt416 That's friggin' awesome. Let's put this puppy to bed!

PR #796, opened 2018-05-02
RE-1317 Add support for automated releases

git-harry:
Issue: [RE-1317](https://rpc-openstack.atlassian.net/browse/RE-1317)

### DIFF
--- a/pipeline_steps/github.groovy
+++ b/pipeline_steps/github.groovy
@@ -71,4 +71,31 @@ void add_issue_url_to_pr(){
 }
 
 
+def merge_pr(
+    String org,
+    String repo,
+    String pr_number,
+    String retries=0
+    ){
+  withCredentials([
+    string(
+      credentialsId: 'rpc-jenkins-svc-github-pat',
+      variable: 'pat'
+    )
+  ]){
+    sh """#!/bin/bash -xe
+      cd ${env.WORKSPACE}
+      set +x; . .venv/bin/activate; set -x
+      python rpc-gating/scripts/ghutils.py\
+        --org '$org'\
+        --repo '$repo'\
+        --pat '$pat'\
+        merge_pr\
+        --pull-request-number '$pr_number'\
+        --retries '$retries'
+    """
+  }
+}
+
+
 return this;
--- a/rpc_jobs/merge_pull_request.yml
+++ b/rpc_jobs/merge_pull_request.yml
@@ -1,0 +1,29 @@
+- job:
+    name: 'Merge-Pull-Request'
+    project-type: pipeline
+    concurrent: false
+    properties:
+      - build-discarder:
+          days-to-keep: 30
+    parameters:
+      - rpc_gating_params
+      - validating-string:
+          name: pr_repo
+          regex: "[a-zA-Z0-9-]+/[a-zA-Z0-9._-]+"
+          msg: "The release repository must be supplied."
+      - validating-string:
+          name: pr_number
+          regex: "[0-9]+"
+          msg: "The number of the pull request to merge is required."
+      - validating-string:
+          name: retries
+          regex: "[0-9]+"
+          default: "1"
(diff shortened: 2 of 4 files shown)


rpc_jobs/standard_job_release.yml
```diff
@@ -78,10 +78,197 @@
...
+            )
+            sh """#!/bin/bash -xe
+                virtualenv --python python3 .venv3
+                set +x; . .venv3/bin/activate; set -x
+                pip install -c constraints.txt -r requirements.txt
+                pushd rpc_component
+                    pip install .
+                popd
+            """
+          component = readYaml text: component_text
```
mattt416: Indentation here to line 184 is off.

pipeline_steps/github.groovy
```diff
@@ -71,4 +71,31 @@ void add_issue_url_to_pr(){
 }
 
 
+def merge_pr(
+    String repo,
+    String pr_number,
+    String retries=5,
+    String org="rcbops"
```
mattt416: Could move `org` to first arg to keep associated args together, but complete nit.

rpc_jobs/merge_pull_request.yml
```diff
@@ -0,0 +1,29 @@
...
+          regex: "[a-zA-Z0-9-]+/[a-zA-Z0-9._-]+"
+          msg: "The release repository must be supplied."
+      - validating-string:
+          name: pr_number
+          regex: "[0-9]+"
+          msg: "The number of the pull request to merge is required."
+      - validating-string:
+          name: retries
+          regex: "[0-9]+"
+          default: "1"
```
mattt416: Any reason we default this to 1 here but 0 in the python lib itself?

rpc_jobs/standard_job_release.yml
```diff
@@ -78,10 +78,203 @@
...
+            returnStdout: true
+          )
+
+          println "=== predecessor component CLI standard out ==="
+          println pred_component_text
+
+          pred_component = readYaml text: pred_component_text
+        }
+
+        ORG="rcbops"
```
mattt416: Do we need to temporarily flip this to my username until we start merging all this into rcbops?

rpc_jobs/standard_job_release.yml
```diff
@@ -78,10 +78,203 @@
...
+                    --ref "${MAINLINE}" \
+                publish_tag \
+                    --ref "${MAINLINE}" \
+                    --version "${VERSION}" \
+                generate_release_notes \
+                    --version "${VERSION}" \
+                    --prev-version "${PREVIOUS_VERSION}" \
+                    --script "optional:gating/generate_release_notes/pre" \
+                    --script "optional:gating/generate_release_notes/run" \
+                    --script "optional:gating/generate_release_notes/post" \
```
mattt416: I know this isn't being executed yet, but what happens if a repo doesn't have release notes and we try to run the above?  Does it cause the script to fail?

rpc_jobs/standard_job_release.yml
git-harry: No, it doesn't do anything. If you follow it through it's only used to print a string. I wasn't sure what would happen if RE-Release was used to tag on your repo, given we know it works for rcbops repos I've just got it printing the command it would use when we're ready to release.

rpc_jobs/standard_job_release.yml
git-harry: That's why they're all set to optional, if I read the code correctly it should skip if the script is missing.

rpc_jobs/standard_job_release.yml
git-harry: done

rpc_jobs/merge_pull_request.yml
git-harry: It seemed reasonable for the function to just make the request by default but allow retries if desired.

PR #801, opened 2018-05-03
RE-1408 Implement nodepool metrics stack

odyssey4me:
In order to help understand the health of nodepool,
its nodes and a bit about its usage, we implement
the configuration and new systems to allow nodepool
to send metrics to statsd, then push them to graphite,
then graph them in grafana with dashboards defined in
the git tree using yaml configuration.

This combination mirrors the OpenStack-Infra implementation.

The DNS entries have not yet been created as I need the
IP addresses of the instances before I can do so. I'd
like to get some reviews before going ahead and completing
that. Also, the API key for grafana can only be generated
once it's up - so for now I've not defined it. Once the
servers are built I'll generate it, drop it into password
safe and implement the extra variable.

Finally, several new jobs are added here which duplicate
a lot of things. I've registered [RE-1408](https://rpc-openstack.atlassian.net/browse/RE-1472) to follow up on that
and reduce the duplication.

Issue: [RE-1408](https://rpc-openstack.atlassian.net/browse/RE-1408)


odyssey4me:
recheck_unit_tests

rpc_jobs/setup_statsd.yml
```diff
@@ -0,0 +1,129 @@
...
+    name: Setup-StatsD
+    project-type: pipeline
+    concurrent: false
+    properties:
+      - build-discarder:
+          num-to-keep: 30
+    parameters:
+      - rpc_gating_params
+    triggers:
+      - timed: "@daily"
```
mattt416: Is this necessary if we already have a push trigger for this?

odyssey4me: The daily execution ensures that even if there are no merges going in, the playbook still executes. This ensures that packages are updated and the environment remains configured with the latest keys and such.

rpc_jobs/setup_statsd.yml
```diff
@@ -0,0 +1,129 @@
...
+           - master
+    properties:
+      - build-discarder:
+          num-to-keep: 30
+    triggers:
+        - github
+    builders:
+      - trigger-builds:
+        - project:
+            - "Setup-StatsD"
```
mattt416: Cleaner to have a separate merge trigger job for both statsd and graphite, or better to combine them into one?

odyssey4me: I've registered a follow up card to do the deduplication: https://rpc-openstack.atlassian.net/browse/RE-1472

rpc_jobs/setup_statsd.yml
mattt416: Fair. :)

odyssey4me:
@mattt416 Quite right, I guess I could have pulled the card # off it - I didn't see any harm including it given it's an improvement relating to the epic anyway.

PR #804, opened 2018-05-03
RE-1425 RPC Gating VirtualEnv Container Optimizations

dealako:
- Moved tar and ssh to inside the container
- Added ssh agent support
- Added logic to skip building virtual env if archive exists.
- Changed concurrent to true
- Removed lock

Issue: [RE-1425](https://rpc-openstack.atlassian.net/browse/RE-1425)

### DIFF
--- a/rpc_jobs/build_gating_venv.yml
+++ b/rpc_jobs/build_gating_venv.yml
@@ -40,7 +40,7 @@
 
                     # Determine the SHA and archive name that we'll use
                     pushd rpc-gating; SHA=\$(git rev-parse HEAD); popd
-                    archive="rpcgatingvenv_\${SHA}_test2.tbz"
+                    archive="rpcgatingvenv_\${SHA}.tbz"
 
                     # Set some SSH options
                     SSH_OPTS="-o StrictHostKeyChecking=no"
@@ -124,7 +124,7 @@
 
                       # Determine the SHA and archive name that we'll use
                       pushd rpc-gating; SHA=\$(git rev-parse HEAD); popd
-                      archive="rpcgatingvenv_\${SHA}_test2.tbz"
+                      archive="rpcgatingvenv_\${SHA}.tbz"
 
                       # Should we skip the upload? This file will exist from the previous step if we didn't generate a new archive.
                       if [[ -e \$archive.skip.upload ]]; then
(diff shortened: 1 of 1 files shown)


git-harry:
Moving the code inside the container and using sshagent seem relatively simple changes, I'm not sure if they give functional benefits but they tidy the code a bit.

Skipping the build if it already exists - the job is only ever run if the required venv doesn't exist so this seems to be added to reduce the window.
  - With the current serialisation in place, this would catch the situation where new builds are initiated for a SHA while there is a build in progress to add that venv to the archive. This should prevent the archive from being modified after initial creation.
  - Without the serialisation this reduces the window to the time it takes to complete the job after the test has been made but it doesn't prevent the venv from being rebuilt. Given builds to create the venv are only initiated when it's missing, I suspect this test will likely never catch anything when builds are concurrent.

Enabling concurrent builds / removing the lock. Given they both are attempting similar things, I'm not sure they both need to be there but I'd be inclined to keep the lock and wrap the 4 ssh/scp commands with it. This allows the venvs to be built in parallel, which is what takes the time, whilst serialising the bit that actually modifies the archive to ensure it is only updated when necessary and there is no possibility of conflicting changes.

Most builds never need to run this job. When master is updated on rpc-gating this triggers Jenkins-Job-Builder which creates the new venv and then updates all the jobs. This process isn't perfect, jobs should specify a SHA instead of master to prevent a mismatch between the libraries and the job. Allowing concurrent builds would likely reduce the time for changes in master to be deployed which in turn should reduce the number of jobs that start where master does not match the deployed jobs. That said, I'd say this is the wrong approach to address this issue, I think it would be better to fix the SHA used by the jobs each time they are built by Jenkins-Job-Builder.

rpc-gating development and testing is where the venv build process can be a pain, when several people are testing at the same time the backlog of venv builds can cause a frustrating delay. This will manifest it two different ways, if people are manually testing I would generally expect that to mean they are manually starting a single job. In this situation, these changes will help to reduce the time for the venv to build. The other is the pull request tests that get launched when a pull request is made, in this situation the builds will all start at the same time and so they will all likely be updating the archive around the same time.

odyssey4me:
Given this is one issue being solved, and each commit uses the same title - could we squash the commits?

PR #809, opened 2018-05-09
RE-1408 Fix grafyaml import and dashboard URL

odyssey4me:
This PR fixes the following:

- The URL in the dashboard is incorrect.
- The credentials are not available to the job to use.
- The grafyaml import job fails if the key is not present.

Issue: [RE-1408](https://rpc-openstack.atlassian.net/browse/RE-1408)


odyssey4me:
recheck_unit_tests

PR #814, opened 2018-05-14
RE-1463 Use nodepool nodes for functional tests

odyssey4me:
* Rename g1-8 nodepool nodes and add p2-15 nodes
* Move RPC-O PR jobs to use nodepool
* Move RPC-MAAS PR jobs to use nodepool
* Move RPC-CEPH PR/PM jobs to use nodepool 
* Move RPC-HUMMINGBIRD PR/PM jobs to use nodepool
* Move RPC-ROLE-LOGSTASH PR jobs to use nodepool
* Move RPC-ENG-OPS jobs to use nodepool

### DIFF
--- a/nodepool/files/nodepool.yml
+++ b/nodepool/files/nodepool.yml
@@ -10,29 +10,17 @@ zookeeper-servers:
     port: 2181
 
 labels:
-  - name: ubuntu-xenial
-    # TODO(odyssey4me):
-    # This is a temporarily low setting as the nodes will
-    # not be used until the jenkins integration is complete.
-    # It is here purely to validate that everything is
-    # working as it should be.
+  - name: ubuntu-xenial-g1-8
     min-ready: 3
-    # TODO(odyssey4me):
-    # This is purely here for test purposes to make sure any
-    # existing nodes that haven't been used will get recreated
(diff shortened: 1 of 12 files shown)


nodepool/files/nodepool.yml
```diff
@@ -44,27 +32,39 @@ providers:
...
+      - name: ubuntu-trusty-g1-8
         config-drive: true
-      - name: ubuntu-trusty
+      - name: ubuntu-xenial-p2-15
+        config-drive: true
+      - name: ubuntu-trusty-p2-15
         config-drive: true
     pools: &provider_pools_block
       - name: main
         max-servers: 10
```
hughsaunders: This is two low for 4 labels each with min ready 3.

odyssey4me: It might be, but might not be. This is 10 max-servers per pool, with 3 pools - giving a total of 30 nodepool nodes at the ready at any time. I can push it up if you think it's too low, though. Any suggestions for a number? Perhaps 20 per pool?

hughsaunders: I always forget the pool is spread across regions... so the max available is actually max*regions.

PR #816, opened 2018-05-14
RE-1463 Correct nodepool config

odyssey4me:
1. Switching the ubuntu-xenial and ubuntu-trusty nodes
   to have 0 min-ready means that they can be phased
   out, instead of being yanked out. This is in case
   any existing jobs are still set to use them.

2. The providers/diskimages section refers to the disk
   images used, not the node labels. This patch corrects
   the configuration.

### DIFF
--- a/nodepool/files/nodepool.yml
+++ b/nodepool/files/nodepool.yml
@@ -10,6 +10,12 @@ zookeeper-servers:
     port: 2181
 
 labels:
+  - name: ubuntu-xenial
+    min-ready: 0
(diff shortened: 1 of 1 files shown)


odyssey4me:
recheck_xenial_lint_test

odyssey4me:
recheck_unit_tests

odyssey4me:
recheck_xenial_lint_test

odyssey4me:
recheck_xenial_lint_test

PR #817, opened 2018-05-14
RE-1463 Revert g1-8 to working label and p2-15 to old config

odyssey4me:
Something is not quite right when trying to use the new labels
so we switch to using the old labels for the g1-8 flavor, and
revert the job configuration to use the old mechanism for the
p2-15 jobs until we can work out the issue.

We also adjust the min-ready settings to accommodate this new
strategy.


odyssey4me:
recheck_unit_tests
recheck_lint_test

odyssey4me:
recheck_xenial_lint_test

PR #818, opened 2018-05-14
Categorised UI

hughsaunders:
* Moved statistics generation to javascript/browser.
* Removed the use of pickle
* Extracted data from generated page into separate json file
  which is used as the cache, and the data source for the UI.
* Added failure categorisations.
* Created Failure class and translated previous failure match
  functions into subclasses
* Added categories, and assigned a category to each failure matcher
* Added support for reading Junit files and adding test failure
  information.


odyssey4me:
recheck_xenial_lint_test
recheck_unit_tests

PR #822, opened 2018-05-17
RE-1463 Extend nodepool usage to more PR jobs

odyssey4me:
Now that the issue of nodepool rejecting jobs for labels which
end with a number [a] is resolved, we can revert 935af07.

This extends the use of nodepool to more jobs and a broader set
of flavors, allowing us to monitor these for stability before
moving on to the next stage of job migrations.

[a] rpc-openstack.atlassian.net/browse/RE-1502

In order to cater to nodepool now handling almost all PR jobs,
we adjust the quotas.

1. We set max-servers in the main pool for each region to 20,
   providing a total maximum of 60 servers from the main pool
   which will be handling non-container tests for PR's.

2. We set the min-ready for the ubuntu-xenial and ubuntu-trusty
   nodes to 0 as they're being phased out and replaced with the
   more explicitly named (according to flavor profile)
   ubuntu-xenial-g1-8 and ubuntu-trusty-g1-8.

3. We leave the ubuntu-trusty-g1-8 and ubuntu-trusty-p2-15
   min-servers at 0 (this makes it build nodes on-demand rather
   than keep one ready) as these profiles are very rarely used.

4. We leave the min-ready for ubuntu-xenial-p2-15 at 1 as it is
   relatively rarely used (only by three jobs), and having 3
   available (1 x 3 regions) at all times is more than enough.

5. We up the ubuntu-xenial-g1-8 min-ready to 3 as it is the most
   commonly used. This provides 9 ready nodes at any time.

### DIFF
--- a/nodepool/files/nodepool.yml
+++ b/nodepool/files/nodepool.yml
@@ -11,13 +11,13 @@ zookeeper-servers:
 
 labels:
   - name: ubuntu-xenial
-    min-ready: 3
+    min-ready: 0
     max-ready-age: 3600
   - name: ubuntu-trusty
-    min-ready: 3
+    min-ready: 0
     max-ready-age: 3600
   - name: ubuntu-xenial-g1-8
-    min-ready: 1
+    min-ready: 3
     max-ready-age: 3600
   - name: ubuntu-trusty-g1-8
     min-ready: 0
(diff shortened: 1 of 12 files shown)


odyssey4me:
FYI Tested whether the different profile of node was working in https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-testjp/1 - the job itself failed for other reasons, but the  use of a nodepool node worked just fine.

PR #823, opened 2018-05-17
RE-1514 Allow JJB to execute concurrently

hughsaunders:
With a lock around the actual execution of JJB. This means
JJB will still only be running in serial, but setup stuff such
as building venvs can happen in parallel.

Issue: [RE-1514](https://rpc-openstack.atlassian.net/browse/RE-1514)


hughsaunders:
@mattt416 yeah to reduce the size of the review.

PR #827, opened 2018-05-17
RE-1489 Remove CIT reference from recheck triggers

mattt416:
I suspect we added this when we had multiple jenkins instances running
at the same time, but this is no longer necessary.

We will need to communicate this change to users and merge on a set
date to avoid any surprises.

Issue: [RE-1489](https://rpc-openstack.atlassian.net/browse/RE-1489)

### DIFF
--- a/rpc_jobs/rpc_artifact_build.yml
+++ b/rpc_jobs/rpc_artifact_build.yml
@@ -77,7 +77,7 @@
           org-list:
             - rcbops
           github-hooks: true
-          trigger-phrase: '.*recheck(_cit)?_all.*|.*recheck(_cit)?_artifact.*'
+          trigger-phrase: '.*recheck_all.*|.*recheck_artifact.*'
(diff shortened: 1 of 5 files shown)


mattt416:
Side note -- there is still `CIT` referenced in the `status-context`, but changing this is a much larger piece of work since we'd have to mark the new `status-context` as required in GitHub if it was previously set to required.

git-harry:
Why didn't you change rpc_jobs/unit/re_unit_tests.yml as well?

This change is pretty low impact. I'd say send out an e-mail to give notice, leave it a few days then merge and send out another e-mail or update the notification channel to confirm the change.

mattt416:
@git-harry not sure how I missed `rpc_jobs/unit/re_unit_tests.yml` -- I've updated

mattt416:
Communicated that I'd make this change on June 1, so I'll hold off on merging this until then.

PR #831, opened 2018-05-18
RE-1488 Point to rcbops/releases repo

mattt416:
This commit moves away from the `mattt416/rpc-metadata` repo and moves
towards the `rcbops/releases` repo.

The file `release.yml` is renamed to `re_release_manual.yml` to avoid
any confusion with the newly created `releases.yml` file (which maps
to the `rcbops/releases` repo).

Issue: [RE-1488](https://rpc-openstack.atlassian.net/browse/RE-1488)


scripts/dep_update.sh
```diff
@@ -29,7 +29,7 @@ if [[ "${COMPONENT_DEPENDENCIES_UPDATE}" == 'true' ]]; then
   apt-get update
   apt-get install -y python3-pip
   pip3 install -c "${WORKSPACE}/rpc-gating/constraints_rpc_component.txt" rpc_component
-  component --releases-repo 'https://github.com/mattt416/rpc-metadata' dependency update-requirements
+  component --releases-repo 'https://github.com/rcbops/releases' dependency update-requirements
```
git-harry: That URL is the default, so you can just remove this argument.

rpc_jobs/dummy_pipeline.yml
```diff
@@ -19,10 +19,10 @@
       - '{trigger}-Dep-Update_{repo_name}-{branch}'
 
 - project:
```
git-harry: This should probably be added to a separate file, its purpose is a bit confusing in this dummy pipeline file.

mattt416: Moved into `rpc_releases.yml`. I gave it the `rpc_` prefix because we already have a file called `release.yml`. I figured I'd leave `release.yml` as-s since we'll eventually remove that project once the pipeline is fully operational.

scripts/dep_update.sh
mattt416: Done.

mattt416:
recheck_issue_link

PR #837, opened 2018-05-22
RE-1491 Update use of rpc_component

git-harry:
rpc_component has been moved from mattt416/rpc-metadata to
rcbops/rpc_component. This change updates the use of rpc_component to
pin it to the latest available version.

In addition a few breaking changes have been addressed so that the new
version works where used. These changes centre around more consistent
use of the terms release and version.

Issue: [RE-1491](https://rpc-openstack.atlassian.net/browse/RE-1491)

### DIFF
--- a/constraints_rpc_component.txt
+++ b/constraints_rpc_component.txt
@@ -1,0 +1,3 @@
+GitPython==2.1.9
+git+https://github.com/rcbops/rpc_component@776c7d7ddc96fd62b503c78c980256906db5a6fd#egg=rpc_component
+schema==0.6.7
--- a/rpc_jobs/standard_job_release.yml
+++ b/rpc_jobs/standard_job_release.yml
@@ -28,16 +28,13 @@
           sh """#!/bin/bash -xe
               virtualenv --python python3 .venv3
               set +x; . .venv3/bin/activate; set -x
-              pip install -c constraints.txt -r requirements.txt
-              pushd rpc_component
-                  pip install .
-              popd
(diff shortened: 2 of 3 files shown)


constraints.txt
```diff
@@ -102,3 +102,4 @@ urllib3==1.18.1
 virtualenv==15.1.0
 wheel==0.30.0
 wrapt==1.10.10
+git+https://github.com/rcbops/rpc_component@776c7d7ddc96fd62b503c78c980256906db5a6fd#egg=rpc_component
```
mattt416: Would it be cleaner to have a separate constraints file for rpc_component itself? We never install rpc_component into the standard gating venv so it feels a bit odd having it in here. Also, the constraints we have here don't match what was in https://github.com/mattt416/rpc-metadata/blob/master/constraints.txt -- ie. differing GitPython version and schema is unconstrained.

git-harry: While it's not currently required in the standard gating venv, I wouldn't be surprised if that changes. I thought it would be easier to keep everything in step so there aren't any issues if that happens. I've updated the patch to add a constraint for schema and updated GitPython (it's only used be release related code) but I'm I don't have an issue with creating a separate constraints file if you prefer.

mattt416: As mentioned in chat, you can't install requirements.txt into a python3 venv so it doesn't make much sense bundling these together.

git-harry:
recheck_all cloning library stuck

PR #839, opened 2018-05-23
 RE-1487 Refactor release pull request validation

git-harry:
Issue: [RE-1487](https://rpc-openstack.atlassian.net/browse/RE-1487)


pipeline_steps/common.groovy
```diff
@@ -1438,3 +1438,104 @@ Boolean isSkippable(String skip_pattern, String credentials) {
...
+    throw new Exception("The pull request does not correspond to any of ${types}.")
+  }
+
+  println "Pull request is a ${type}."
+  println "=== component CLI standard out ==="
+  println component_text
+  if (type == "release"){
+    testRelease(component_text)
+  }else{
+    throw new Exception("Unknown pull request type ${type}")
```
dealako: `types` array seems to have two valid values, but we only test for "release".  Perhaps this exception should say "Pull request type ${type} currently not supported" as the previous loop and null check identifies types that are not defined in the array.

git-harry: Reworded.

Good spot, I missed this when I rewrote my first version.

mattt416:
recheck_issue_link
recheck_xenial_lint_test

pipeline_steps/common.groovy
```diff
@@ -1438,3 +1438,104 @@ Boolean isSkippable(String skip_pattern, String credentials) {
...
+            ]
+          ]
+        )
+      }
+    }
+  }
+
+  parallel parallelBuilds
+
+  build(
```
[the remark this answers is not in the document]
git-harry: @mattt416 I've split things up a bit to, hopefully, make it clearer. I'm testing it now but take a look and let me know what you think.

PR #841, opened 2018-05-23
RE-1516 Wait for up to 10 mins for SSH connectivity

odyssey4me:
OnMetal hosts take ~5 mins to be connectible via SSH. We
set the timeout to 10 mins to give it some breathing room.

Issue: [RE-1516](https://rpc-openstack.atlassian.net/browse/RE-1516)


mattt416:
recheck_all

PR #842, opened 2018-05-24
RE-1466 Add nodepool config for RPC-O artifacted base images

odyssey4me:
In order to implement the ability for nodepool to build a base
image which is usable for RPC-O newton strict artifact builds,
we add the appropriate diskimage-builder elements and config
to build an image using the apt artifact repository, and the
appropriate nodepool config to ensure that we are able to
request nodes using this image.

We deliberately do not want to convert any jobs over to using
this image yet as we need to confirm that the nodes provided
with this image are usable and do not break the tests.

Issue: [RE-1466](https://rpc-openstack.atlassian.net/browse/RE-1466)

### DIFF
--- a/nodepool/files/nodepool.yml
+++ b/nodepool/files/nodepool.yml
@@ -25,6 +25,9 @@ labels:
   - name: ubuntu-xenial-om-io2
     min-ready: 0
     max-ready-age: 3600
+  - name: rpco-14.2-xenial-base
+    min-ready: 0
+    max-ready-age: 3600
 
 providers:
   - name: pubcloud-iad
@@ -40,6 +43,8 @@ providers:
       - name: ubuntu-xenial
         config-drive: true
       - name: ubuntu-trusty
+        config-drive: true
+      - name: rpco-14.2-xenial-base
         config-drive: true
     pools:
       - name: main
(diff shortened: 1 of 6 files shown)


nodepool/files/nodepool.yml
```diff
@@ -154,3 +164,26 @@ diskimages:
       DIB_GRUB_TIMEOUT: '0'
       DIB_DISTRIBUTION_MIRROR: 'http://mirror.rackspace.com/ubuntu'
       DIB_DEBIAN_COMPONENTS: 'main,universe'
+  - name: rpco-14.2-xenial-artifacts
```
odyssey4me: This image is intended to replace https://github.com/rcbops/rpc-gating/blob/master/rpc_jobs/rpc_openstack.yml#L343 with something built and managed by nodepool, rather than relying on something that was a once-off and not repeatable.

nodepool/files/nodepool.yml
```diff
@@ -66,6 +71,11 @@ providers:
             flavor-name: performance2-15
             key-name: jenkins
             console-log: true
+          - name: rpco-14.2-xenial-base
+            diskimage: rpco-14.2-xenial-artifacts
+            flavor-name: 7
```
odyssey4me: This flavor comes from https://github.com/rcbops/rpc-gating/blob/master/rpc_jobs/rpc_openstack.yml#L357 which is where the new image would be used.

PR #843, opened 2018-05-24
RE-1463 Adjust rpc-maas periodic jobs

mattt416:
Currently we have pike, newton, and ceph all gating on a trusty image,
but pike only supports xenial and it appears that the ceph job is also
only run on xenial in the pre-merge tests.

This commit splits out the periodic tests into a separate trusty and
xenial project.


mattt416:
@odyssey4me @tonytan4ever We don't have full periodic coverage (we're missing master, ocata) -- are those needed or are we happy w/ how things are right now?

PR #844, opened 2018-05-24
RE-1489 Rename hook dirs

mattt416:
We've opted to use `check` instead of `pre_merge_test`, and `periodic`
instead of `post_merge_test`. This commit renames both accordingly.

Issue: [RE-1489](https://rpc-openstack.atlassian.net/browse/RE-1489)


mattt416:
recheck_re_unit_tests

mattt416:
recheck_unit_tests

mattt416:
@git-harry Yep, thought I herp derped but tried again and it did it again.

PR #845, opened 2018-05-24
RE-1505 Remove large values from summary data

hughsaunders:
Currently build summary is very laggy becuase data.json is very
large. Its large because some jobs have very long lines (an
entire RPCO run in one case). And that entire line is getting
embedded into the data.json file.

This commit puts an 1000 char limit on the detail field.

Issue: [RE-1505](https://rpc-openstack.atlassian.net/browse/RE-1505)


hughsaunders:
I've already run this and it reduced the json blob from ~19MB to ~4MB. The page still takes a few seconds to load data, but at least chrome doesn't get bored and ask you if you want to kill the tab..

PR #853, opened 2018-05-29
RE-1526 Browsable Artifacts Attempt 2

hughsaunders:
The first attempt at this failed, as we were using a cloud files
api call which takes an archive and expands it out into a container.
That operation was too slow and unreliable to be useful.

As a recap, the aim is to have artifacts be browsable (and therefore
linkable) but also downloadable in bulk for local analysis.

This commit achieves both aims, by uploading a single archive but
also uploading each file individually. Both of these operations
have potential to take significant amounts of time. Archive creation
has taken 3 hours in the past due to compression. In order to speed
that up, this commit switches from bzip compression to gzip --fast
which can be an order of magnitude faster. Single file upload can
also be slow due to the number of api operations it requires. The
api rate limit is 100r/s so the upload is done with 100 threads.

In my testing, compression and both upload operations for an RPCO
artifact set took 7 minutes. That included 20k files totalling ~450mb.

Issue: [RE-1526](https://rpc-openstack.atlassian.net/browse/RE-1526)

### DIFF
--- a/playbooks/templates/artifact/index.html
+++ b/playbooks/templates/artifact/index.html
@@ -1,0 +1,393 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>RE Artefacts</title>
+  <link href='https://fonts.googleapis.com/css?family=Roboto:300,400,500,700|Material+Icons' rel="stylesheet" type="text/css">
+  <link href="https://unpkg.com/vuetify/dist/vuetify.min.css" rel="stylesheet" type="text/css"></link>
+  <link href="styles.css" rel="stylesheet" type="text/css">
+</head>
+<body>
+  <div id="app">
+    <v-app>
+      <v-navigation-drawer
+        fixed
+        :mini-variant="miniVariant"
+        :clipped="clipped"
+        v-model="drawer"
+        app
+      >
+        <v-list>
+          <v-list-tile :value="true" v-for="(item, i) in items" :key="item.title" :href="item.url">
+            <v-list-tile-action>
+              <v-icon light v-html="item.icon"></v-icon>
+            </v-list-tile-action>
+            <v-list-tile-content>
+              <v-list-tile-title v-text="item.title"></v-list-tile-title>
+            </v-list-tile-content>
+          </v-list-tile>
+        </v-list>
+      </v-navigation-drawer>
+      <v-toolbar fixed app :clipped-left="clipped">
+        <v-toolbar-side-icon @click.native.stop="drawer = !drawer"></v-toolbar-side-icon>
+        <v-toolbar-title v-text="title"></v-toolbar-title>
+        <v-spacer></v-spacer>
+      </v-toolbar>
+      <main>
+        <v-content>
+          <v-container fluid>
+            <v-layout column>
+              <downloadcard
+                :archives="this.archives"
+              ></downloadcard>
+              <filelist
+                :files="this.files"
+                :container_public_url="this.container_public_url"
+                :dataloaded="this.dataloaded"
+                :archive_base_name="this.archive_base_name"
+              ></filelist>
+            </v-layout>
+          </v-container>
+        </v-content>
+      </main>
+      <v-footer :fixed="fixed" app>
+        <span>&copy; Rackspace 2018</span>
+      </v-footer>
+    </v-app>
+  </div>
+
+  <script src="https://unpkg.com/vue/dist/vue.js"></script>
+  <script src="https://unpkg.com/vuetify/dist/vuetify.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/vue-resource@1.5.0"></script>
+  <script>
+
+  // Add human readable display func to Number objects
+  //https://stackoverflow.com/a/20463021
+    Object.defineProperty(Number.prototype,'fileSize',{value:function(a,b,c,d){
+      return (a=a?[1e3,'k','B']:[1024,'K','iB'],b=Math,c=b.log,
+        d=c(this)/c(a[0])|0,this/b.pow(a[0],d)).toFixed(2)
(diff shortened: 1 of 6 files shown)


hughsaunders:
Test build: https://rpc.jenkins.cit.rackspace.net/job/RE-unit-test-artefact-publication/1091/

Test artifact set: https://c1a782ed17e259edf5ef-2b88460b05191c0636b516667c68e711.ssl.cf1.rackcdn.com/RE-unit-test-artefact-publication_1091/index.html

hughsaunders:
Note that the artifact unit test job has been rebuilt from the master branch to prevent branch incompatibilities causing failures.

hughsaunders:
recheck_re_unit_tests

hughsaunders:
recheck_unit_tests

hughsaunders:
Failed in artifact publication test due to incompatibility with the running job and this branch, however when the unit test is built from this branch it works fine (see job linked above). All the other unit tests passed.

playbooks/upload_to_swift.yml
```diff
@@ -3,10 +3,42 @@
...
+      args:
+        chdir: "{{ adir |dirname }}"
+
+    # This task runs async while the individual files are being uploaded
+    # then when compression is complete, the resultant archive is uploaded.
+    - name: Create archive
+      shell: "tar -c {{ adir | basename }} | gzip --fast > {{archive_base_name}}.tar.gz"
+      args:
+        chdir: "{{ adir | dirname }}"
+      async: 10800
```
odyssey4me: It'd be super-nice to have a comment telling the reader that this is a 3 hour timeout.

hughsaunders:
New test build, now with full length expiry, so the artefacts don't disappear during the review process...

https://rpc.jenkins.cit.rackspace.net/job/RE-unit-test-artefact-publication/1097/

https://9c30252635e5fd19e5cf-bfd1b1e72ab7517418453cfa37d567ac.ssl.cf1.rackcdn.com/RE-unit-test-artefact-publication_1097/index.html

playbooks/upload_to_swift.yml
hughsaunders: Done

playbooks/upload_to_swift.yml
```diff
@@ -62,23 +101,107 @@
...
+    # os_object does not currently support setting object expiration header field
+    # production retention
+    - name: Upload Artifacts to Cloud Files
+      command: "swift upload --object-threads 100 --ignore-checksum --header 'X-Delete-After:{{ retention }}' {{ container }} {{ adir | basename }}"
+      args:
+        chdir: "{{ adir |dirname }}"
+      environment:
+        OS_AUTH_TOKEN: "{{ auth_token }}"
+        OS_STORAGE_URL: "{{ (object_store['endpoints'] | selectattr('region', 'equalto', region) | first)['publicURL'] }}"
+      failed_when: false
```
odyssey4me: Why are we forcing this never to fail? If there's a legit reason, we should probably add a comment as to why.

Does the swift client do retries on its own, or shouldn't this task do retries if there's a failure? If the swift client retries, does it see whether some files are there already and skip them?

playbooks/upload_to_swift.yml
```diff
@@ -62,23 +101,107 @@
...
+    # os_object does not currently support setting object expiration header field
+    - name: Upload archive and index files to Cloud Files
+      command: >-
+        swift upload {{ container }} {{ archive_base_name }}.tar.gz
+        --header 'X-Delete-After: {{ retention }}'
+      args:
+        chdir: "{{ adir | dirname }}"
+      environment:
+        OS_AUTH_TOKEN: "{{ auth_token }}"
+        OS_STORAGE_URL: "{{ (object_store['endpoints'] | selectattr('region', 'equalto', region) | first)['publicURL'] }}"
```
odyssey4me: Should we add retries here?

playbooks/upload_to_swift.yml
```diff
@@ -62,23 +101,107 @@
...
+
+    # used by common.archive_artifacts to put the link into the build description
+    - name: "Write public url file"
+      shell: |
+        echo "{{ container_public_url }}/{{archive_base_name}}/index.html" > ${WORKSPACE}/artifact_public_url
   vars:
     region: "DFW"
     cloud_name: "public_cloud"
+    retention: 2592000
```
odyssey4me: Can we add a comment here to describe how long this default retention period is?

hughsaunders: Done

playbooks/upload_to_swift.yml
hughsaunders: done

playbooks/upload_to_swift.yml
odyssey4me: Ah ok, so sift continues uploading anything else. Sounds good then. Is there a way to capture the failed files (if there are any) so that we can also put that into a file to show?

playbooks/upload_to_swift.yml
hughsaunders: Added a section for upload failures, that only appears if there is at leat one failure: https://901a6f3144ea1698911a-fb37949d4321c68f6e404c5dda72c3d1.ssl.cf5.rackcdn.com/hstest_123/index.html

playbooks/templates/artifact/index.html
```diff
@@ -0,0 +1,454 @@
...
+
+    Vue.component('linkwarning', {
+      props: ["removed_links_count"],
+      template: `
+        <v-card v-if="removed_links_count > 0">
+         <v-card-title>
+           <div>
+             <h3>Symlinks Removed</h3>
+             <div>
+              {{ removed_links_count }} symlinks have been remvoved from this
```
odyssey4me: ``removed``

hughsaunders: Ooops, done.

hughsaunders:
recheck_all

hughsaunders:
Test run of the artefacts unit test (the only test that failed) from the head of this branch: https://rpc.jenkins.cit.rackspace.net/job/RE-unit-test-artefact-publication/1126/

PR #858, opened 2018-05-30
RE-1537 Add test for constraints leaks

hughsaunders:
Also solves the cmd2 >= 0.9 requires python 3.4 problem.

Issue: [RE-1537](https://rpc-openstack.atlassian.net/browse/RE-1537)


rpc_jobs/unit/venvconstraintleak.yml
```diff
@@ -0,0 +1,56 @@
...
+    parameters:
+      - rpc_gating_params
+    dsl: |
+      library "rpc-gating@${RPC_GATING_BRANCH}"
+      common.globalWraps(){
+        stage("Check constraints file matches built venv"){
+          sh """
+            .venv/bin/activate
+            rc=0
+            pip freeze |sort > new-constraints.txt
```
odyssey4me: ``pip freeze --all`` will include pip/setuptools/wheel, which we should constrain

PR #861, opened 2018-05-31
Change Jira key from K8S to RE

xgerman:
With Octavia becoming officially supported in RPC-O we should report
CI failures to the RE queue instead of K8S

### DIFF
--- a/rpc_jobs/rpc_octavia.yml
+++ b/rpc_jobs/rpc_octavia.yml
@@ -27,7 +27,7 @@
     action:
       - "test"
 
-    jira_project_key: "K8S"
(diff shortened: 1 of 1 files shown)


xgerman:
Ah, I copied designate and they send some to RE -- but the rest to RI... will change them all to RI then

mancdaz:
Hi @xgerman - apologies it's taken us a while to get to this PR. If it's still something you'd like changed, please rebase and ping us in #rpc-releng so we can review and merge.

odyssey4me:
It appears that this has been abandoned.

PR #865, opened 2018-06-05
RE-1486 Clean-up pull request branch on merge

git-harry:
This change updates the merge_pr function to attempt to delete the pull
request branch. If rpc-jenkins-svc has the correct permissions and the
branch isn't protected, it will delete the branch. This functionality is
used by the Merge-Pull-Request job.

The repo context object is modified to add the github session object to
allow interaction with the pull request repo because it may not be the
same as the base branch repo.

Issue: [RE-1486](https://rpc-openstack.atlassian.net/browse/RE-1486)

### DIFF
--- a/scripts/ghutils.py
+++ b/scripts/ghutils.py
@@ -45,6 +45,7 @@ def cli(ctxt, org, repo, pat, debug):
     logging.basicConfig(level=level)
     gh = github3.login(token=pat)
     repo_ = gh.repository(org, repo)
+    repo_.github = gh
     repo_.org = gh.organization(org)
     if not repo_:
         raise ValueError("Failed to connect to repo {o}/{r}".format(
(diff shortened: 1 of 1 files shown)


scripts/ghutils.py
```diff
@@ -152,6 +153,20 @@ def merge_pr(repo, pull_request_number, commit, message, retries):
         else:
             raise Exception("There was a failure updating the pull request.")
 
+    pr_repo = repo.github.repository(*pull_request.head.repo)
+    pr_ref = pr_repo.ref("heads/{ref}".format(ref=pull_request.head.ref))
+    try:
+        pr_branch_deleted = pr_ref.delete()
```
dealako: Looks good.  As a side question, is there a need to confirm if the ref exists before we execute the delete?

git-harry: I guess it depends why you want the information. It's pretty unlikely that it would have been deleted after the merge on 146 and before this attempt but if it is gone, the code won't error so it's not going to cause the build to fail.

PR #866, opened 2018-06-05
RE-1297 Remove `re` ref from status context

mattt416:
Currently the status context and trigger phrase doesn't match. Instead
of adding `re` to the trigger phrase, we just remove it from the status
context since we don't use this prefix anywhere else.

Once this merges, we'll need to update the repo settings to make this
test required (since it's going to appear as a new test).

Additionally, we remove an unnecessary `?` in the trigger phrase. This
was left over when we removed the `cit` reference from the trigger
phrase in 36f717f5.

Issue: [RE-1297](https://rpc-openstack.atlassian.net/browse/RE-1297)

### DIFF
--- a/rpc_jobs/unit/re_unit_tests.yml
+++ b/rpc_jobs/unit/re_unit_tests.yml
@@ -7,10 +7,10 @@
           org-list:
             - rcbops
           github-hooks: true
-          trigger-phrase: '.*recheck_all.*|.*recheck_unit_?tests.*'
+          trigger-phrase: '.*recheck_all.*|.*recheck_unit_tests.*'
           only-trigger-phrase: false
           auth-id: "github_account_rpc_jenkins_svc"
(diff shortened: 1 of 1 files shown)


mattt416:
Adding `status-dont-merge` label so we can coordinate the merge w/ making this new job required.

rpc_jobs/unit/re_unit_tests.yml
```diff
@@ -10,7 +10,7 @@
           trigger-phrase: '.*recheck_all.*|.*recheck_unit_?tests.*'
```
git-harry: Why is there a `?` in the trigger phrase?

mattt416: Fixed, this was my doing in 36f717f5 it seems. :(

mattt416:
To document, no jobs are required on rpc-gating, so nothing to change here.

PR #871, opened 2018-06-07
RE-1575 Removes Mitaka Gating Jobs

antonym:
Mitaka deploy test jobs are no longer needed, so removing

Issue: [RE-1575](https://rpc-openstack.atlassian.net/browse/RE-1575)


antonym:
@hughsaunders We aren't doing any greenfield installs of Mitaka anymore so there's not much need for them anymore.

PR #873, opened 2018-06-07
RE-1483 Add release jobs for rpc-maas

mattt416:
Issue: [RE-1483](https://rpc-openstack.atlassian.net/browse/RE-1483)


mattt416:
We probably want to ensure [this](https://github.com/rcbops/rpc-maas/pull/518) has merged before this one merges.

PR #874, opened 2018-06-08
RE-1535 Fix release when no previous version

git-harry:
If the first version of a component is being released the previous
version argument, used for generating release notes, gets set to an
empty string. It is left the component to determine how to use that
information.

Issue: [RE-1535](https://rpc-openstack.atlassian.net/browse/RE-1535)


git-harry:
@hughsaunders we'll I'm using an empty string to show the absence of a previous version, it will be up to the projects to determine what that means in the context of generating the relevant release notes.

PR #878, opened 2018-06-09
RE-1550 Add rpc-octavia pike snapshot

mattt416:
We also flip the flavor size back to 7, since we've since had a
successful run on this flavor [1].

[1] https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-octavia-master-xenial_snapshot-functional-test/16/

Issue: [RE-1550](https://rpc-openstack.atlassian.net/browse/RE-1550)


mattt416:
We should let [this](https://github.com/rcbops/rpc-octavia/pull/52) merge before this one merges.

PR #880, opened 2018-06-11
RE-1513 Update rpc_component SHA

git-harry:
This change updates rpc_component to the latest version to ensure
rpc-gating works with the new component file format, where release order
is reversed.

Issue: [RE-1513](https://rpc-openstack.atlassian.net/browse/RE-1513)


git-harry:
recheck_unit_tests

PR #885, opened 2018-06-12
RE-1484 Add rpc-openstack release jobs

mattt416:
This commit adds a pike and newton MNAIO release job for rpc-openstack.
The MNAIO jobs are often referred to to determine the health of a
branch before release, so it makes most sense to run these prior to
releasing.

Additionally, we update a comment in rpc_maas.yml to better reflect our
intention of splitting out check and gate jobs after rpc-maas has been
successfully added to the pipeline. This comment was inadvertently
copied from rpc-ceph and since rpc-maas already has a lint and
functional jobs this should be a straight-forward exercise for us to
do.

Issue: [RE-1484](https://rpc-openstack.atlassian.net/browse/RE-1484)

### DIFF
--- a/rpc_jobs/rpc_maas.yml
+++ b/rpc_jobs/rpc_maas.yml
@@ -84,9 +84,9 @@
     jobs:
       - 'PR_{repo_name}-{series}-{image}-{scenario}-{action}'
 
-# NOTE(mattt): the rpc-maas team would rather run all functional
-#              tests on PR (`check`), rather than `gate`. Leaving these
-#              all commented for now.
+# TODO(mattt): We need to split out existing pre-merge jobs between check and
+#              gate, but this will be done after rpc-maas has been
+#              successfully added to the release pipeline.
 #- project:
 #    name: "rpc-maas-master-xenial-gate"
 #    repo_name: "rpc-maas"
--- a/rpc_jobs/rpc_openstack.yml
+++ b/rpc_jobs/rpc_openstack.yml
@@ -1,3 +1,7 @@
+# TODO(mattt): We need to split out existing pre-merge jobs between check and
+#              gate, but this will be done after rpc-openstack has been
+#              successfully added to the release pipeline.
+
 - project:
     name: "rpc-openstack-master-premerge"
     repo_name: "rpc-openstack"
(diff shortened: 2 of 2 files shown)


rpc_jobs/rpc_openstack.yml
```diff
@@ -139,6 +139,31 @@
...
+#              gate, but this will be done after rpc-openstack has been
+#              successfully added to the release pipeline.
+
+- project:
+    name: 'rpc-openstack-mnaio-release'
+    repo_name: 'rpc-openstack'
+    repo_url: 'https://github.com/rcbops/rpc-openstack'
+    branch:
+      - pike
+      - newton
```
odyssey4me: The 'newton' branch cannot do a 'no_artifacts' job. It only has the 'loose_artifacts' job option. This will need to be split into two different project definitions, or use exclusions.

rpc_jobs/rpc_openstack.yml
```diff
@@ -437,6 +441,65 @@
     jobs:
       - 'PM_{repo_name}-{branch}-{image}-{scenario}-{action}'
 
+- project:
+    name: 'rpc-openstack-master-mnaio-release'
+    repo_name: 'rpc-openstack'
+    repo_url: 'https://github.com/rcbops/rpc-openstack'
+    branch: "master"
```
odyssey4me: The rpc-openstack jobs to validate a release should actually come from the -rc branches, not the mainline branches. I guess there may actually be a requirement to run both the mainline and -rc branch jobs as the mainline becomes the next rc. Will this automagically figure out which right jobs are, or does this have to be more specific?

mattt416: The release jobs actually build w/ the version being proposed, so the `branch` listed in these jobs is really just to distinguish them from one-another. I can certainly change them to `master-rc` and `newton-rc` if that makes it more clear (and I suppose it should, since that's where the proposed changes are coming from).

mattt416: Changed the branch to `-rc` to better reflect the branch the change we're running with is from.

rpc_jobs/rpc_maas.yml
```diff
@@ -84,9 +84,9 @@
     jobs:
       - 'PR_{repo_name}-{series}-{image}-{scenario}-{action}'
 
-# NOTE(mattt): the rpc-maas team would rather run all functional
-#              tests on PR (`check`), rather than `gate`. Leaving these
-#              all commented for now.
+# TODO(mattt): We need to split out existing pre-merge jobs between check and
+#              gate, but this will be done after rpc-maas has been
+#              successfully added to the release pipeline.
```
hughsaunders: Even if gate is a no-op, we should still be creating a gate job, so people get used to @rpc-jenkins-svc doing the merges.

mattt416: @hughsaunders they can still run `(re)?check_component_gate` without there being any gate jobs.

mattt416: @hughsaunders You don't need any actual gate jobs defined for them to be able to run `(re)?check_component_gate` and have Jenkins handle the merge.

PR #887, opened 2018-06-12
RE-1544 Add rpc-hummingbird release job

mattt416:
Issue: [RE-1544](https://rpc-openstack.atlassian.net/browse/RE-1544)


mattt416:
recheck_all

PR #893, opened 2018-06-14
RE-1513 Improve gate test guarantees

git-harry:
Without this change, any Component-Gate-Trigger_ job tests a pull
request against the target branch using the merge branch supplied by
GitHub but does not account for any changes that merge while those tests
are running.

This change ensures that a pull request is tested with any other pull
requests that are also going through the gate. To achieve this a build
looks for any builds that were started before it and includes their pull
requests in its own tests. If any of those builds fails, the tests on
dependent builds are terminated with a link to the upstream build and
new ones are started.

This change should provide a stronger guarantee as to the state of the
repository however if tests are unstable or pull requests are approved
that are broken, the time to merge a working pull request may increase.

This functionality can be disabled by setting `test_with_antecedents:
false` in a project.

Component-Gate-Trigger_ jobs that use this functionality no longer
depend on Merge-Pull-Request. This was done to simplify the tracking of
pull requests and the success of the tests. To enable merging to still
work, Component-Gate-Trigger_ will now update the status context for
"CIT/gate" so that the GitHub required status does not block the build
from merging the pull request.

REException class added to help reduce the noise in the build console
log. The new exception type is used to allow an exception to be raised
to exit the build without the stack trace appearing in the log.

The following test builds show what happens under various situations:
https://rpc.jenkins.cit.rackspace.net/job/Component-Gate-Trigger_rpc-product-1-gh/273
https://rpc.jenkins.cit.rackspace.net/job/Component-Gate-Trigger_rpc-product-1-gh/274
https://rpc.jenkins.cit.rackspace.net/job/Component-Gate-Trigger_rpc-product-1-gh/275

Issue: [RE-1513](https://rpc-openstack.atlassian.net/browse/RE-1513)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -549,6 +549,30 @@ void clone_external_repo(String directory, String repo, String ref, String refsp
 }
 
 
+/**
+ * Merge list of pull requests.
+ *
+ * Iterate through the list of pull request IDs and merge each change
+ * on top of the current HEAD. If a merge fails an exception is raised.
+ *
+ * @param directory location of the Git repository
+ * @param pullRequestIDs ordered list of pull request numbers to merge
+ * @return null
+ */
+void merge_pr_chain(String directory='./', List pullRequestIDs=null){
+  println("Attempting to merge the following pull requests onto the base branch: ${pullRequestIDs}.")
+  dir(directory){
+    for (pullRequestID in pullRequestIDs) {
+      println("Merging pull request ${pullRequestID}.")
+      sh """#!/bin/bash -xe
+        git merge origin/pr/${pullRequestID}/head
+      """
+    }
+  }
+  println("Finished merging the pull requests onto the base branch.")
+}
+
+
 void configure_git(){
   print "Configuring Git"
   // credentials store created to ensure that non public repos
@@ -1355,6 +1379,10 @@ void stdJob(String hook_dir, String credentials, String jira_project_key, String
                   env.REPO_URL,
                   env.BRANCH,
                 )
+                if (env.pullRequestChain) {
+                  pullRequestIDs = loadCSV(env.pullRequestChain)
+                  merge_pr_chain("${env.WORKSPACE}/${env.RE_JOB_REPO_NAME}", pullRequestIDs)
+                }
               } else {
                 print("Triggered by PR: ${env.ghprbPullLink}")
                 clone_with_pr_refs(
@@ -1655,4 +1683,14 @@ void createRelease(){
       ],
     ]
   )
+}
+
+
+List loadCSV(String str){
+  str.split(",").collect {it.trim()}
+}
+
+
+String dumpCSV(List l){
+  l.join(",")
 }
--- a/pipeline_steps/gate.groovy
+++ b/pipeline_steps/gate.groovy
@@ -1,0 +1,363 @@
+import org.jenkinsci.plugins.workflow.job.WorkflowJob
+import org.jenkinsci.plugins.workflow.job.WorkflowRun
+import groovy.transform.Field
+
+import com.rackspace.exceptions.REException
+
+
+void testPullRequest(String repoName, Boolean testWithAntecedents, String githubStatusContext){
+    try {
+        if (testWithAntecedents) {
+           testPullRequestWithAntecedents(repoName, githubStatusContext)
+        } else {
+           testPullRequestOnly(repoName)
+        }
+    }catch (REException e){
+        currentBuild.result="FAILURE"
+    }
+}
+
+
+void testPullRequestOnly(String repoName){
+    List jobNames = Hudson.instance.getAllItems(org.jenkinsci.plugins.workflow.job.WorkflowJob)*.fullName
+
+    def parallelBuilds = [:]
+
+    // Cannot do for (job in jobNames), see:
+    // https://jenkins.io/doc/pipeline/examples/#parallel-multiple-nodes
+    for (j in jobNames) {
+        def job = j
+        def m = job =~ /GATE_${repoName}/
+        if (m) {
+            parallelBuilds[job] = {
+                build(
+                    job: job,
+                    wait: true,
+                    parameters: [
+                        [
+                            $class: "StringParameterValue",
+                            name: "RPC_GATING_BRANCH",
+                            value: RPC_GATING_BRANCH,
+                        ],
+                        [
+                            $class: "StringParameterValue",
+                            name: "BRANCH",
+                            value: sha1,
+                        ]
+                    ]
+                )
+            }
+        }
+    }
+
+    parallel parallelBuilds
+
+    build(
+      job: "Merge-Pull-Request",
+      wait: false,
+      parameters: [
+        [
+          $class: "StringParameterValue",
+          name: "RPC_GATING_BRANCH",
+          value: RPC_GATING_BRANCH,
+        ],
+        [
+          $class: "StringParameterValue",
+          name: "pr_repo",
+          value: ghprbGhRepository,
+        ],
+        [
+          $class: "StringParameterValue",
+          name: "pr_number",
+          value: ghprbPullId,
+        ],
+        [
+          $class: "StringParameterValue",
+          name: "commit",
+          value: ghprbActualCommit,
+        ],
+      ]
+    )
+}
+
+
+@Field WorkflowRun triggerBuild = currentBuild.rawBuild
+@Field Result SUCCESS = Result.fromString("SUCCESS")
+/**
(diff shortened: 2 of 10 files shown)


pipeline_steps/common.groovy
```diff
@@ -537,6 +537,18 @@ void clone_external_repo(String directory, String repo, String ref, String refsp
 }
 
 
+void merge_pr_chain(String directory='./', List pullRequestIds=null){
+  println("Attempting to merge the following pull requests onto the base branch: ${pullRequestIds}.")
+  for (pullRequestId in pullRequestIds) {
+    println("Merging pull request ${pullRequestId}.")
+    sh """#!/bin/bash -xe
+      cd ${directory}
```
mattt416: Nit, but can you move the `cd` outside of the loop so it doesn't keep `cd`ing back to the same place?

pipeline_steps/gate.groovy
```diff
@@ -0,0 +1,194 @@
...
+                            value: ghprbTargetBranch,
+                        ],
+                        [
+                            $class: "StringParameterValue",
+                            name: "pullRequestChain",
+                            value: common.dumpCSV(pullRequestIDs),
+                        ],
+                    ]
+                )
+                while(true){
```
hughsaunders: Could you kick off all the builds in a loop, then in a second loop, wait for queuing and find each of the new builds? Then the queue processing (which may involve nodepool requests) can happen in parallel.

pipeline_steps/gate.groovy
```diff
@@ -0,0 +1,194 @@
...
+        }else {
+            sleep(time: 120, unit: "SECONDS")
+        }
+    }
+
+    // Not sure where to get the context from so definition duplicated from job definition
+    statusContext = "CIT/gate"
+    (pr_repo_org, pr_repo_name) = ghprbGhRepository.split("/")
+    description = "Gate tests passed, merging..."
+    github.create_status(pr_repo_org, pr_repo_name, ghprbActualCommit, "success", triggerBuild.getAbsoluteUrl(), description, statusContext)
```
hughsaunders: A comment in here to say that this allows the PR to merge when `CIT/gate` is a required context in github configuration. Also that this does duplicate the status setting that GHPRB will do at the end of the job, but that github doesn't mind the same status/context getting set multiple times.

pipeline_steps/gate.groovy
```diff
@@ -0,0 +1,194 @@
...
+          $class: "StringParameterValue",
+          name: "commit",
+          value: ghprbActualCommit,
+        ],
+      ]
+    )
+}
+
+
+void testPullRequestWithAntecedents(String repo_name){
```
hughsaunders: Comment with a high level overview of the logic of this function would be useful here.

pipeline_steps/gate.groovy
```diff
@@ -0,0 +1,194 @@
...
+            sleep(time: 120, unit: "SECONDS")
+        }
+    }
+
+    // Not sure where to get the context from so definition duplicated from job definition
+    statusContext = "CIT/gate"
+    (pr_repo_org, pr_repo_name) = ghprbGhRepository.split("/")
+    description = "Gate tests passed, merging..."
+    github.create_status(pr_repo_org, pr_repo_name, ghprbActualCommit, "success", triggerBuild.getAbsoluteUrl(), description, statusContext)
+    github.merge_pr(pr_repo_org, pr_repo_name, ghprbPullId, ghprbActualCommit)
```
hughsaunders: I like that this is not a separate job.

pipeline_steps/common.groovy
```diff
@@ -537,6 +537,18 @@ void clone_external_repo(String directory, String repo, String ref, String refsp
 }
 
 
+void merge_pr_chain(String directory='./', List pullRequestIds=null){
```
hughsaunders: My first thought reading this was that it was going to merge a series of PRs publicly in github. It may be useful to add a comment that this is for merging a series of PRs into a local repo for testing and doesn't affect the public PRs.

pipeline_steps/gate.groovy
git-harry: As we discussed, I've been working on this already. It doesn't make a great deal of difference but for projects with a lot of tests there will be a small improvement. I'm looking to remove the loop completely and have the checking done as part of the outer loop's task.

pipeline_steps/gate.groovy
```diff
@@ -0,0 +1,194 @@
...
+        if (currentAntecedentTriggerBuilds){
+            println("Active antecedent builds:\n${currentAntecedentTriggerBuilds.join("\n")}")
+        }else {
+            println("No active antecedent builds.")
+        }
+        if (antecedentTriggerBuilds.collect(){it.getNumber()} != currentAntecedentTriggerBuilds.collect(){it.getNumber()}) {
+            failed = antecedentTriggerBuilds.findAll() {it.getResult() && it.getResult().isWorseThan(success)}
+            if (failed) {
+                println("Antecedent builds that have completed with failure:\n${failed.join("\n")}")
+                println("Killing existing builds that depend on failed antecedents.")
```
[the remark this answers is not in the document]
git-harry: I'll take a look at this. I'm tempted to say it should be a follow up task because end users shouldn't really need to look at these builds. This code lists the cause of the failure, i.e. the downstream builds that failed, at the end of the log file so I think it's clear what's happening.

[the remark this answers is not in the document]
hughsaunders: Its clear in the trigger job, but not in the job that gets killed, and the trigger is not necessarily an obvious place to look for the cause if you're not familiar with the structure of related jobs.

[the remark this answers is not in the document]
git-harry: The pull request only links to the trigger build so the only way the user finds the downstream build is via the trigger build log, well unless they are searching through all the builds on particular jobs but that would be pretty silly. The trigger build prints failures at the end of the log so there is a logical sequence from pull request to any failures. Again I'm not dismissing the change but a standard user workflow in terms of reviewing any failure should not require them to view builds killed by the trigger build.

git-harry:
@mattt416 @hughsaunders updated to address feedback.

hughsaunders:
Re: my review comment, we already have some dummy classes to get round this problem in https://github.com/rcbops/rpc-gating/tree/master/scripts/lint_support_groovy

pipeline_steps/common.groovy
```diff
@@ -549,6 +549,30 @@ void clone_external_repo(String directory, String repo, String ref, String refsp
...
+ *
+ * Iterate through the list of pull request IDs and merge each change
+ * on top of the current HEAD. If a merge fails an exception is raised.
+ *
+ * @param directory location of the Git repository
+ * @param pullRequestIDs ordered list of pull request numbers to merge
+ * @return null
+ */
+void merge_pr_chain(String directory='./', List pullRequestIDs=null){
+    println("Attempting to merge the following pull requests onto the base branch: ${pullRequestIDs}.")
```
mattt416: Nit: inconsistent indentation. Also, we're adding a mix of CamelCase and snake_case names here, any reason for that? I'm not suggesting to change, perhaps we can standardise on this later and update once we've got agreement.

git-harry: The repo is not consistent on any of this stuff. I used snake case for the function name to be consistent with the other functions, `pullRequestIDs` is camel case because there is no consistency in the project and that was what I adopted in gate.groovy. I meant to keep the indentation the same but forgot when I added the println/dir bits. If I update the indentation does that work for you?

pipeline_steps/gate.groovy
```diff
@@ -0,0 +1,363 @@
...
+            sleep(time: 120, unit: "SECONDS")
+        }
+    }
+
+    (pr_repo_org, pr_repo_name) = ghprbGhRepository.split("/")
+    /* When a check is required by GitHub, it prevents the pull request being merged if
+       it is not marked as `"success"`. Normally GHPRB would be soley responsible for
+       updating the status context however this would necessitate a separate job to
+       perform the merge. The following section updates the pull request's status
+       context on GitHub if all tests were successful to enable the same bulid to merge
```
mattt416: Nit: build

pipeline_steps/common.groovy
mattt416: Totally

pipeline_steps/gate.groovy
git-harry: Don't tell me you don't know what a bulid is...

git-harry: done

pipeline_steps/common.groovy
git-harry: done

PR #894, opened 2018-06-14
ASC-655 Add sdqc action

johnduarte:
This commit adds an 'sdqc' action to the
`rpc-openstack-master-mnaio-postmerge` project. The `ironic` scenario is
excluded when this action is used. So, the only job that should be
created by this change is:

    PM_rpc-openstack-master-xenial_mnaio_no_artifacts-swift-sdqc

The purpose of this is to run the deployment and validation of
[openstack-ops](https://github.com/rsoprivatecloud/openstack-ops/) in
isolation from other system test suites. Currently, this is necessary
due to a conflict in the deployment ansible code.

Issue: [ASC-655](https://rpc-openstack.atlassian.net/browse/ASC-655)

### DIFF
--- a/rpc_jobs/rpc_openstack.yml
+++ b/rpc_jobs/rpc_openstack.yml
@@ -201,9 +201,12 @@
       - swift
     action:
       - system
+      - sdqc
       - deploy
     exclude:
       - action: system
+        scenario: ironic
+      - action: sdqc
         scenario: ironic
     # Required by RPC-ASC team to upload test results qTest
     credentials: "rpc_asc_creds"


johnduarte:
@mattt416 we are trying to automate the sign-off process for SD. We are trying to use a deployment that is as representative of a customer/lab deploy that we can. That is the reason for the MNAIO deploy.
Ideally, this would occur as part of the system test job, but we are experiencing conflicts between the ansible system preparation done by `openstack-ops` and `openstack-ansible-ops`.
We expect this job to long be long lived as we migrate the validations performed here into the system testing suite.

PR #900, opened 2018-06-19
RE-1619 Catch Artifact Archival Failures

hughsaunders:
Artifact archival is not part of the test being executed, its a
convenience provided by RE. As such failures in that process
should not cause a build to fail.

This commit adds try/catch around artifact archival, and creates
a Jira issue on failure, instead of letting the exception propagate
up and fail the build.

Issue: [RE-1619](https://rpc-openstack.atlassian.net/browse/RE-1619)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -299,46 +299,58 @@ String container_name(){
 
 def archive_artifacts(Map args = [:]){
   stage('Compress and Publish Artifacts'){
+    try{
+      artifacts_dir = args.get("artifacts_dir", "${env.WORKSPACE}/artifacts")
+      results_dir = args.get("results_dir", "${env.WORKSPACE}/results")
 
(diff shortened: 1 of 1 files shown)


mancdaz:
recheck_unit_tests

pipeline_steps/common.groovy
```diff
@@ -299,46 +299,58 @@ String container_name(){
...
+      if(fileExists(file: "artifact_public_url")){
+        artifact_public_url = readFile(file: "artifact_public_url")
+        currentBuild.description = "<h2><a href='"+artifact_public_url+"'>Build Artifacts</a></h2>"
+      }
+    } catch (e){
+      // This function is called from stdJob, so we must catch exceptions to
+      // prevent failures in RE code eg (artifact archival) from causing
+      // the build result to be set to failure.
+      // try-catch placed here because this function is used in multiple places.
+      print "Error while arching artifacts, swallowing this exception to prevent "\
```
git-harry: typo: arching

PR #903, opened 2018-06-20
ASC-665 Externalize RE_JOB_BRANCH for asc staging

johnduarte:
This commit converts the asc staging job from a jjb project into a jjb
job in order to create the RE_JOB_BRANCH as an available parameter to be
set when a build is started. The RE_JOB_BRANCH is used internally by
the gating scripts to determine the branch of
`rpc-openstack-system-tests` to be checked out. Making this available as
a settable variable will allow topic branches of the testing repo to be
specified when triggering the job.

Issue: [ASC-665](https://rpc-openstack.atlassian.net/browse/ASC-665)

### DIFF
--- a/rpc_jobs/rpc_openstack.yml
+++ b/rpc_jobs/rpc_openstack.yml
@@ -533,51 +533,23 @@
     jobs:
       - 'Component-Gate-Trigger_{repo_name}'
 
-- job:
-    name: "ASC-staging-rpc-openstack-master-xenial-mnaio-no-artifacts-swift-system"
+- project:
+    name: "experimental-asc-rpc-openstack-master-mnaio-postmerge"
+    repo_name: "rpc-openstack"
+    repo_url: "https://github.com/rcbops/rpc-openstack"
+    branch:
+      - "master"
+    CRON: "@monthly"
     jira_project_key: ""
-    project-type: pipeline
-    concurrent: true
-    properties:
-      - build-discarder:
-          num-to-keep: 14
-      - github:
-          url: "https://github.com/rcbops/rpc-openstack"
-      - inject:
-          properties-content: |
-            STAGES="Allocate Resources, Connect Slave, Cleanup, Destroy Slave"
-            BOOT_TIMEOUT=900
-            RE_JOB_NAME=ASC-Staging
-            RE_JOB_IMAGE=staging_asc_xenial_mnaio_no_artifacts
-            RE_JOB_SCENARIO=swift
-            RE_JOB_ACTION=system
-            RE_JOB_FLAVOR=onmetal-io2
-            RE_JOB_REPO_NAME=rpc-openstack
-    parameters:
-      - string:
-          name: RE_JOB_BRANCH
-          default: dev
-          description: Branch of rpc-openstack-system-tests to use
-      - rpc_gating_params
-      - instance_params:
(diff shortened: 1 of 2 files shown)


johnduarte:
@mattt416 there are two problems with kicking off an existing production job and redefining the `BRANCH` and `REPO_URL` variables.

1. Procedural challenge: The `master` job is being watched for the health of the Queens version of `rpc-openstack` and causing it to go red by using to run experimental system jobs is undesirable. This is the core reason for wanting to setup the staging job.
2. Technical challenge: The version of `rpc-openstack-system-tests` deployed during a test run is defined by the `RE_JOB_BRANCH` variable. This variable is set as a property on the Jenkins job at the time the JJB code is interpolated by the value of `branch`. For example on the `PM_rpc-openstack-master-xenial_mnaio_no_artifacts-swift-system` job the JJB value for `branch` is set  to `master` and therefore the value of `RE_JOB_BRANCH` is set to master. If, during a test run, the Jenkins`BRANCH` value is set to a custom value for the run, the value of `RE_JOB_BRANCH` is not re-evaluated and remains set to `master`. So, if we were to execute a test run with the master job using a temporary fork of `rpc-openstack` and defining values for `BRANCH` and `REPO_URL`, the custom fork and branch for `rpc-openstack` would be used but the `master` branch of `rpc-openstack-system-tests` would be used to to the value of `RE_JOB_BRANCH`.

mattt416:
@johnduarte Understand point 1, and no objection to having a staging job for experimental runs, etc. I'd just rather see you use the standard templates if at all possible.

With point 2, I think you've just uncovered a bug here. I don't believe the intention is for RE_JOB_BRANCH, etc. to ignore the parameters passed into the job (which appears to be happening as you point out). I believe this changed [here](https://github.com/rcbops/rpc-gating/commit/e81f490dfd22d41838289527746e11f08fd86cc9#diff-bb9b2383291a458116c9f93cb90c08a5R56), so I'll create a separate bug for this so we can revert how these env vars are injected.

On a side note, could you not modify rpc-openstack's gating/check/run to do something like this?

```
if [[ $RE_JOB_ACTION == "system" ]]; then
  if [[ $RE_JOB_NAME =~ .*ASC-staging.* ]]; then
    export RE_JOB_BRANCH="dev"
  fi
  bash -c "$(readlink -f $(dirname ${0})/run_system_tests.sh)"
fi
```

It may be better to update gating/check/run and gating/check/run_system_tests.sh so that you're overriding SYS_TEST_BRANCH rather than RE_JOB_BRANCH, but I see you're using this pattern already for `sdqc` so should doing something similar should be OK for now.

mattt416:
Created [RE-1620](https://rpc-openstack.atlassian.net/browse/RE-1620) for this bug, @johnduarte. Also accidentally closed this PR, sorry about that! :)

johnduarte:
@mattt416 as an alternative, I have pushed another commit that creates a new job-template that uses the existing standard post-merge template and overrides the parameters in order to add SYS_TEST_BRANCH as a Jenkins parameter. Let me know if this is equally crazy. Thanks.

johnduarte:
The pattern of introducing additional parameters is in conflict with RE team's objective to standardize gating jobs. I am closing this PR and will address the issue in the https://github.com/rcbops/rpc-openstack repo.

PR #904, opened 2018-06-21
RE-1606 Check skip_pattern in gate jobs

mattt416:
Currently, the gate jobs do not check if the commit is only changing
files in the skip_list. This means we may end up running lengthy gate
jobs which don't need to be run. This commit updates
pipeline/gate.groovy to only run gate jobs if isSkippable() is false.

Additionally, we update rpc-maas' skip_list replacing
gating/post_merge_test with gating/periodic, and add gating/release to
the existing skip_pattern. We also duplicate this skip_pattern to the
commented gate jobs in rpc_maas.yml.

Lastly, for testing, we update dummy_pipeline.yml to add a skip_pattern
to rpc-product-1's check and gate jobs.

Issue: [RE-1606](https://rpc-openstack.atlassian.net/browse/RE-1606)

### DIFF
--- a/pipeline_steps/gate.groovy
+++ b/pipeline_steps/gate.groovy
@@ -19,36 +19,35 @@ void testPullRequest(String repoName, Boolean testWithAntecedents, String github
 
 
 void testPullRequestOnly(String repoName){
-    List jobNames = Hudson.instance.getAllItems(org.jenkinsci.plugins.workflow.job.WorkflowJob)*.fullName
+    List allWorkflowJobs = Hudson.instance.getAllItems(WorkflowJob)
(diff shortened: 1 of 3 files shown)


mattt416:
Here's an example of a Component-Gate-Trigger skipping builds because of the specified skip_pattern: https://rpc.jenkins.cit.rackspace.net/job/Component-Gate-Trigger_rpc-product-1/19/

mattt416:
Here's an example of a Component-Gate-Trigger which filters out jobs based on skip_pattern:

https://rpc.jenkins.cit.rackspace.net/job/Component-Gate-Trigger_rpc-product-1/31/

(the failure is expected as it's trying to merge a PR that hasn't been reviewed)

PR #906, opened 2018-06-21
RE-1606 Add a skip_pattern to rpc-hummingbird

mattt416:
Issue: [RE-1606](https://rpc-openstack.atlassian.net/browse/RE-1606)


corystone:
LGTM!

PR #908, opened 2018-06-27
RE-1503 Add rpc-gating merge trigger job

odyssey4me:
In order to properly co-ordinate the execution of merge
trigger jobs when commits are merged into rpc-gating,
we create a single merge trigger job which co-ordinates
the others appropriately.

Please see the individual commits for details of each change.

Tested successfully here:
https://rpc.jenkins.cit.rackspace.net/job/Merge-Trigger-rpc-gating/19

Issue: [RE-1503](https://rpc-openstack.atlassian.net/browse/RE-1503)

### DIFF
--- a/job_dsl/rpc_gating_merge_trigger.groovy
+++ b/job_dsl/rpc_gating_merge_trigger.groovy
@@ -1,0 +1,71 @@
+// Can't use env.FOO = {FOO} to transfer JJB vars to groovy
+// as this file won't be templated by JJB.
+// Alternative is to use parameters with JJB vars as the defaults.
+library "rpc-gating@${RPC_GATING_BRANCH}"
+common.globalWraps(){
+
+  // We do not want to trigger any of these
+  // for rpc-gating forks or branches other
+  // than 'master'. So if this is triggered
+  // by anything else, end now.
+  // TODO(odyssey4me):
+  // Remove once RE-1645 is resolved.
+  if (env.RPC_GATING_BRANCH != "master") {
+    print "Triggered by branch ${RPC_GATING_BRANCH} instead of master. Skipping triggers."
+    currentBuild.result = 'SUCCESS'
+    return
+  }
+
+  // Here we define the jobs to execute when a merge happens.
+  // Some are executed in series, and some parallel. We wait
+  // until they are all complete before completing the pipeline.
+
+  List serialJobs = [
+    'Build-Gating-Venv',
+    'Jenkins-Job-Builder'
+  ]
+
+  List parallelJobs = [
+    'Build-Summary-Docker-Build',
+    'Setup-Grafana',
+    'Setup-Graphite',
+    'Setup-Nodepool',
+    'Setup-StatsD'
+  ]
+
+  List jobParameters = [
+    [
+      $class: 'StringParameterValue',
+      name: 'RPC_GATING_BRANCH',
+      value: env.RPC_GATING_BRANCH
+    ]
+  ]
+
+  for (job in serialJobs) {
+    stage(job) {
+      build(
+        job: job,
+        wait: true,
+        parameters: jobParameters
+      ) // build
+    } // stage
+  } // for
+
+  def parallelBuilds = [:]
+
+  // Cannot do for (job in jobNames), see:
+  // https://jenkins.io/doc/pipeline/examples/#parallel-multiple-nodes
+  for (j in parallelJobs) {
+    def job = j
+    parallelBuilds[job] = {
+      build(
+        job: job,
+        wait: true,
+        parameters: jobParameters
+      ) // build
+    } // parallelBuilds
+  } // for
+
+  parallel parallelBuilds
+
+} // globalWraps
--- a/rpc_jobs/rpc_gating.yml
+++ b/rpc_jobs/rpc_gating.yml
@@ -1,0 +1,20 @@
+- job:
+    name: "Merge-Trigger-rpc-gating"
+    project-type: pipeline
+    concurrent: false
+    properties:
+      - rpc-gating-github
+      - build-discarder:
+          num-to-keep: 30
+    triggers:
+        - github
+    parameters:
+      - rpc_gating_params
+    pipeline-scm:
+      scm:
+        - git:
+            url: https://github.com/rcbops/rpc-gating
+            branches:
+              - "${RPC_GATING_BRANCH}"
+            credentials-id: "github_account_rpc_jenkins_svc"
+      script-path: job_dsl/rpc_gating_merge_trigger.groovy
--- a/rpc_jobs/unit/re_unit_tests.yml
+++ b/rpc_jobs/unit/re_unit_tests.yml
@@ -22,6 +22,25 @@
       if(env.ghprbPullId != null){
         env.RPC_GATING_BRANCH="origin/pr/${env.ghprbPullId}/merge"
       }
+
+      // We execute this here so that when a PR is submitted
+      // all the parallel builds below do not have to do their
+      // own request for the venv to be built. This helps reduce
+      // the load on jenkins and gets the job done faster.
+      stage("Build-Gating-Venv") {
+        build(
+          job: "Build-Gating-Venv",
+          wait: true,
+          parameters: [
+            [
+              $class: 'StringParameterValue',
+              name: 'RPC_GATING_BRANCH',
+              value: env.RPC_GATING_BRANCH
+            ]
+          ]
+        ) // build
+      } // stage
+
       parallel ([
         "With Requested Credentials": {
           build(
--- a/rpc_jobs/build_summary.yml
+++ b/rpc_jobs/build_summary.yml
@@ -1,23 +1,3 @@
-# Trigger a build summary docker image build when a change
-# is merged to rpc-gating master
-- job:
-    name: "Merge-Trigger-Docker-Build"
-    project-type: freestyle
-    scm:
-      - git:
-          url: https://github.com/rcbops/rpc-gating
-          branches:
-           - master
-    properties:
-      - build-discarder:
-          num-to-keep: 30
-    triggers:
-        - github
-    builders:
-      - trigger-builds:
-        - project:
-            - "Build-Summary-Docker-Build"
-
 # Build and publish the build summary docker image
 - job:
     name: 'Build-Summary-Docker-Build'
--- a/rpc_jobs/jjb_setup.yml
+++ b/rpc_jobs/jjb_setup.yml
@@ -1,28 +1,3 @@
-# This job is specified as `freestyle` as using a pipeline job with github
-# trigger causes the job to be triggered irrespective of which branch was
-# pushed to.  Additionally, the environment variables we were trying to use to
-# determine the branch to pushed to were not being passed into the job by the
-# plugins in question.
-# Also, having a separate merge trigger job ensures that JJB still runs on
-# pushes to master, even if the JJB job is used to test an alternative branch.
-- job:
-    name: "Merge-Trigger-JJB"
-    project-type: freestyle
-    scm:
-      - git:
-          url: https://github.com/rcbops/rpc-gating
-          branches:
-           - master
-    properties:
-      - build-discarder:
-          num-to-keep: 30
-    triggers:
-        - github
-    builders:
-      - trigger-builds:
-        - project:
-            - "Jenkins-Job-Builder"
-
 # Node CentOS to ensure the CIT slave is used. If a pub cloud ubuntu slave is used,
 # the Jenkins API won't be reachable.
 - job:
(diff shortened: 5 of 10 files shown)


rpc_jobs/rpc_gating.yml
```diff
@@ -0,0 +1,20 @@
...
+    parameters:
+      - rpc_gating_params
+    pipeline-scm:
+      scm:
+        - git:
+            url: https://github.com/rcbops/rpc-gating
+            branches:
+              - "${RPC_GATING_BRANCH}"
+            credentials-id: "github_account_rpc_jenkins_svc"
+      script-path: job_dsl/rpc_gating_merge_trigger.groovy
```
git-harry: @odyssey4me slack is still broken for me so I've only just seen the e-mail notification. Not had this before but based on the error, try using a dsl block and put the contents of rpc_gating_merge_trigger.groovy in it.

odyssey4me:
recheck_issue_link
recheck_xenial_lint_test

odyssey4me:
recheck_unit_tests

rpc_jobs/rpc_gating.yml
odyssey4me: Thanks, I worked it out. It turns out it comes down to where the script was running from and whether it had been imported correctly. I did some rework and it's working like a charm now.

rpc_jobs/standard_job_premerge.yml
```diff
@@ -1,23 +1,3 @@
...
-    image:
-      - "xenial":
-          SLAVE_TYPE: "nodepool-ubuntu-xenial-g1-8"
-    scenario:
-      - "lint"
-    action:
-      - "test"
-    jobs:
-      - 'PR_{repo_name}-{series}-{image}-{scenario}-{action}'
-
```
mattt416: Is this definitely no longer needed? I'm not seeing a lint job under the unit tests.

rpc_jobs/rpc_gating.yml
```diff
@@ -0,0 +1,20 @@
...
+    triggers:
+        - github
+    parameters:
+      - rpc_gating_params
+    pipeline-scm:
+      scm:
+        - git:
+            url: https://github.com/rcbops/rpc-gating
+            branches:
+              - "${RPC_GATING_BRANCH}"
```
mattt416: This `$` is weirding me out here ... is that correct? I see this snippet of code used elsewhere so it shouldn't break anything, but I do wonder if it's right. Also, presumably we can put anything here and it doesn't matter due to RE-1645?

pipeline_steps/common.groovy
```diff
@@ -1625,35 +1625,44 @@ void createComponentGateTrigger(String name, String repoUrl, String jiraProjectK
   }
 }
 
+List getMatchingJobs(regexMatch){
```
mattt416: We pass in a string here, can set the type to `String`.

job_dsl/rpc_gating_merge_trigger.groovy
```diff
@@ -0,0 +1,78 @@
...
+        ]
+      ]
+    ) // build
+  } // stage
+
+  // To prevent having to continually update a list of jobs to run
+  // in parallel once the two blocking jobs are complete, we match
+  // the jobs against a regex and execute everything that matches
+  // it in parallel.
+  jobNames = common.getMatchingJobs("(?:Build-Summary-Docker-Build|Setup-.*)")
```
mattt416: What's the meaning of `?:`? Also, this will match `Setup-`. Should we lock that down a bit tighter?

odyssey4me: ``?:`` makes the () a non-capturing group, making just a small optimisation as we don't need to use a capture group anywhere. https://www.regular-expressions.info/brackets.html

odyssey4me: By locking it down tighter, what exactly do you mean? The only jobs we have that I can see prefixed by 'Setup-' are the jobs which setup infrastructure for rpc-gating. We should probably making our naming conventions better, but this handles what we have.

pipeline_steps/common.groovy
odyssey4me: Good call, will do that.

rpc_jobs/rpc_gating.yml
odyssey4me: Both good questions. I'm reusing the existing pattern. RE-1645 is not confirmed today and needs to be verified, so I'd rather leave this here in case it actually works. It is a JJB variable so "${var}" is right to me?

rpc_jobs/standard_job_premerge.yml
odyssey4me: Yeah, OK - I've restored this job. Good catch!

pipeline_steps/common.groovy
```diff
@@ -1625,35 +1625,44 @@ void createComponentGateTrigger(String name, String repoUrl, String jiraProjectK
   }
 }
 
+List getMatchingJobs(String regexMatch){
```
odyssey4me: Note that I've not implemented the use of this in ``pipeline_steps/gate.groovy`` on purpose. The implementation there is more complex and using this function is not really the focus of this card's work. I've registered https://rpc-openstack.atlassian.net/browse/RE-1647 to look at how we can consolidate the approaches if it makes sense to do so.

job_dsl/rpc_gating_merge_trigger.groovy
```diff
@@ -0,0 +1,78 @@
...
+library "rpc-gating@${RPC_GATING_BRANCH}"
+common.globalWraps(){
+
+  // We do not want to trigger any of these
+  // for rpc-gating forks or branches other
+  // than 'master'. So if this is triggered
+  // by anything else, end now.
+  // TODO(odyssey4me):
+  // Remove once RE-1645 is resolved.
+  if (env.RPC_GATING_BRANCH != "master") {
```
git-harry: This is a funny one. Given that this is terminating the build right at the start, it essentially negates the ability to override the branch. The only reason the branch wouldn't be master is if someone is doing testing, in most situations that would automatically require this to be commented out. I guess it acts as extra protection against someone doing something stupid.

job_dsl/rpc_gating_merge_trigger.groovy
git-harry: I wonder if a list of jobs would be easier. It would only need modifying when a job is added or removed. There doesn't appear to be a clear discriminator with the names so I wouldn't be surprised if merge jobs end up getting added that don't use this.

job_dsl/rpc_gating_merge_trigger.groovy
```diff
@@ -0,0 +1,78 @@
...
+
+  // Cannot do for (job in jobNames), see:
+  // https://jenkins.io/doc/pipeline/examples/#parallel-multiple-nodes
+  for (j in jobNames) {
+    def job = j
+    parallelBuilds[job] = {
+      stage(job) {
+        build(
+          job: job,
+          wait: false, // no need to wait for these, they're non-blocking
```
git-harry: Maybe it would be better to have concurrent set to false in the merge job and wait here. In theory, it seems like two merges in quick succession could result in a situation where an older build finish second resulting in the job regressing.

pipeline_steps/common.groovy
```diff
@@ -1625,35 +1625,44 @@ void createComponentGateTrigger(String name, String repoUrl, String jiraProjectK
...
+  List jobNames = Jenkins.getInstance().getAllItems()*.fullName
+  List matchingJobs = []
+  // Cannot do for (job in jobNames), see:
```
git-harry: I'm pretty sure that's not the case here because there is no closure. With the for-loop for the builds `parallelBuilds` is a map of `String` to `Closure` whereas here is simply a `List` of `String`.

rpc_jobs/unit/re_unit_tests.yml
```diff
@@ -22,8 +22,27 @@
...
+              name: 'RPC_GATING_BRANCH',
+              value: env.RPC_GATING_BRANCH
+            ]
+          ]
+        ) // build
+      } // stage
+
       parallel ([
-        "With Requested Credentials": {
+        stage("With Requested Credentials") {
```
git-harry: Have you rebuilt the job and tested these changes? The examples I've seen have parallel take a map which would no longer be the case if this is done.

pipeline_steps/common.groovy
hughsaunders: The script console is a good place for trying stuff like this out. Also @NonCPS methods can be used for accessing more interesting groovy language functions.

rpc_jobs/rpc_gating.yml
hughsaunders: This isn't a JJB template, so `${}` will be interpreted by Jenkins, as an environment variable lookup. Parameters are set as env vars before the lookup, so it should work fine.

rpc_jobs/unit/re_unit_tests.yml
hughsaunders: @git-harry makes a good point; parallel takes a map, of string -> closure. Each parallel branch is effectively a stage so theres little value in adding the stage call. Also `stage()` doesn't return a closure so this wouldn't work even if it was a map. I recommend leaving it how it was. 

The docs are confusing on this, as [declarative pipeline](https://jenkins.io/doc/book/pipeline/syntax/#parallel-stages-example) uses 
```
pipeline {
  stages {
    stage('a'){
      ...
    }
    stage('b'){
      ...
    }
  }
}
```

We currently use scripted pipeline which uses the map syntax - Its easy to get the two syntaxes confused.

job_dsl/rpc_gating_merge_trigger.groovy
hughsaunders: I agree, at least its parallel waiting.

odyssey4me: Good point. OK, given this and the suggestion about not using regex, I think I'll convert this into a list and execute with a wait.

rpc_jobs/unit/re_unit_tests.yml
odyssey4me: Alright, will revert it. I primarily did this because it looks prettier in the UI. :p

pipeline_steps/common.groovy
odyssey4me: Yep, it seems to work:

```
List getMatchingJobs(String regexMatch){
  List jobNames = Jenkins.getInstance().getAllItems()*.fullName
  List matchingJobs = []
  for (job in jobNames) {
    if (job =~ /${regexMatch}/) {
      matchingJobs+=job
    }
  }
  return matchingJobs
}

println getMatchingJobs("^Setup-\\w+\$")
```

job_dsl/rpc_gating_merge_trigger.groovy
```diff
@@ -16,58 +16,53 @@ common.globalWraps(){
...
-        [
-          $class: 'StringParameterValue',
-          name: 'RPC_GATING_BRANCH',
-          value: env.RPC_GATING_BRANCH
-        ]
-      ]
-    ) // build
-  } // stage
+  List serialJobs = [
+    'Build-Gating-Venv',
```
git-harry: Jenkins-Job-Builder will start Build-Gating-Venv if the venv doesn't exist so this could be left out.

job_dsl/rpc_gating_merge_trigger.groovy
```diff
@@ -16,58 +16,53 @@ common.globalWraps(){
...
   def parallelBuilds = [:]
 
   // Cannot do for (job in jobNames), see:
   // https://jenkins.io/doc/pipeline/examples/#parallel-multiple-nodes
-  for (j in jobNames) {
+  for (j in parallelJobs) {
     def job = j
     parallelBuilds[job] = {
-      stage(job) {
+      job: {
```
git-harry: This syntax doesn't look correct. I think you want to remove this line. It looks like you're trying to define a `Map` with this line but it's happening inside a `Closure`. I'm pretty sure `parallelBuilds[job]` is already doing what this line wants.

job_dsl/rpc_gating_merge_trigger.groovy
odyssey4me: Yes, that's true, but I'd prefer this to be explicit personally. Also, relying on the one job to kick off the other means further delays (wait for build, then wait a minute, then carry on), which isn't necessary.

job_dsl/rpc_gating_merge_trigger.groovy
odyssey4me: Ah, silly me - I think you're right. Good catch!

git-harry:
@odyssey4me no, that's not what I meant. If you take the commit 'Redefine merge-trigger job to use static lists', it seems to be reworking previous commits in the pull request. I'm suggesting that should be squashed down so that we're not merging commits that change the implementation of other commits in the same pull request.

odyssey4me:
@git-harry Thanks, the more specific answer is more useful. 👍 I'll sort that out tomorrow.

PR #910, opened 2018-06-28
Add CONTRIBUTING.md

mancdaz:
Add some contributing guidelines after a recent discussion around
whether to use branches in-repo or personal forks for development.

Issue: [RE-154](https://rpc-openstack.atlassian.net/browse/RE-154)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,73 @@
+# Contributing to this repository
+
+# Table of Contents
+1. [Commits and pull requests](#commits-release-notes-pull-requests)
+    1. [Fixing a bug](#fixing-a-bug)
+    2. [Commits](#commits)
+    3. [Pull requests](#pull-requests)
+        1. [Reviewing pull requests](#reviewing-pull-requests)
+        2. [Merging pull requests](#merging-pull-requests)
+
+## Commits and pull requests
+### Fixing a bug
+
+1. Fork the repo and commit changes to a new branch in your fork.
+2. Make pull requests against the master branch in the rpc-gating repo.
+3. Reference the JIRA issue in both the commit message and the pull request message.
+
+### Commits
+
+Please avoid the following in a commit:
+
+* Mixing whitespace changes with functional code changes.
+* Mixing two unrelated functional changes in the same commit.
+* Sending large new features in a single commit.
+
+Expected git commit message structure (Based on [this article](https://chris.beams.io/posts/git-commit/)):
+
+* The first line should ideally be limited to 50 characters, but must be no more than 70 characters.
+* The first line should not end with a full stop and the first word should be capitalised.
+* Use the imperative mood in the subject line - e.g. ```Fix a typo in CONTRIBUTING.md```, ```Remove if/else block in myfile.sh```, ```Eat your dinner```.
+* Insert a single blank line after the first line.
+* Subsequent lines should be wrapped at 72 characters.
+* Provide a detailed description of the change in the following lines, using the guidelines in the section below.
+* The last line of the commit should be a reference to the issue being fixed using the word JIRA, followed by the JIRA issue Key e.g. ```JIRA: RE-1500```
+
+In your commit message please consider the following points:
+
+* In the commit subject line, provide a brief description of the change.
+* The commit message must be descriptive and contain all the information required to fully understand and review the patch.
+* Do not assume the reviewer understands what the original problem was.
+* Do not assume the reviewer has access to external web services or site.
+* Do not assume the code is self-evident or self-documenting.
+* Describe why the change is being made.
+* Ensure sufficient information is provided to review your commit.
+* Each commit should ideally solve a single problem. Where there are several things to fix to reach the goal identified by the card, multiple commits should be used. Each commit should be as easy as possible to consume as a single, complete change.
+
+
+
+
+### Pull requests
+
+* A pull request (PR) should contain a single commit, or multiple commits all related to the problem being solved by the PR.
+* The PR title and message should contain information relevant to why the commit(s) are happening. This can be based off the commit message.
+* The PR description should reference the original issue.
+
+#### Reviewing pull requests
+
+When reviewing a PR, please ensure the included commit(s):
+
+* Actually works to fix the issue.
+* Passes the gate checks that are configured to run.
+* Contains a single commit, or multiple commits all related to the problem being solved by the PR.
+* Does not overreach. Each PR should be self contained to only address the issue at hand (see above).
+
+#### Merging pull requests
+
+In order for a PR to be merged, the following criteria should be met:
+
+* All gate tests must have passed.
+* There must be at least 2 members of the release engineering team review and give the PR a +1 approval using the review guidelines above.
+ * Use GitHub's [pull request reviews](https://help.github.com/articles/about-pull-request-reviews/) function so that your comments are collated into a single review.
+* The second +1 reviewer should merge the patch.
+


CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
...
+* Sending large new features in a single commit.
+
+Expected git commit message structure:
+
+* The first line should be limited to 50 characters, should not end with a full stop and the first word should be capitalised.
+* Use the imperative mood in the subject line - e.g. ```Fix a typo in CONTRIBUTING.md```, ```Remove if/else block in myfile.sh```, ```Eat your dinner```.
+* Insert a single blank line after the first line.
+* Subsequent lines should be wrapped at 72 characters.
+* Provide a detailed description of the change in the following lines, using the guidelines in the section below.
+* The last line of the commit should be a reference to the issue being fixed using the JIRA issue Key e.g. ```RE-1500```
```
git-harry: Highlighting this because it's different from what we've been doing, generally we've been adding this as the first word of the commit message title. I'd prefer us to start adding it at the end of the message because having it in the title takes up precious characters.

CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
...
+
+# Table of Contents
+1. [Commits and pull requests](#commits-release-notes-pull-requests)
+    1. [Fixing a bug](#fixing-a-bug)
+    2. [Commits](#commits)
+    3. [Pull requests](#pull-requests)
+        1. [Reviewing pull requests](#reviewing-pull-requests)
+        2. [Merging pull requests](#merging-pull-requests)
+
+## Commits and pull requests<a name="commits-release-notes-pull-requests"></a>
```
hughsaunders: I don't think these anchors are needed, GFM anchors the headings by default, so you should be able to link to another section using `#` followed by the heading lowercased with spaces replaced by dashes eg `[link text](#commits-and-pull-requests)`. 

Also html5 depreciates the name attribute in favour of id.

CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
...
+
+In your commit message please consider the following points:
+
+* In the commit subject line, provide a brief description of the change.
+* The commit message must be descriptive and contain all the information required to fully understand and review the patch.
+* Do not assume the reviewer understands what the original problem was.
+* Do not assume the reviewer has access to external web services or site.
+* Do not assume the code is self-evident or self-documenting.
+* Describe why the change is being made.
+* If the commit message suggests an improved code structure, you may be able to split this into two or more commits.
```
hughsaunders: I don't quite understand this point.

CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
...
+* Ensure sufficient information is provided to review your commit.
+
+
+### Pull requests<a name="pull-requests"></a>
+
+* A pull request (PR) should ideally contain a single commit
```
hughsaunders: Therefore "large new features" should come in multiple PRs?

CONTRIBUTING.md
mancdaz: if you're both restructuring the code, and also fixing a bug, it's possible that the restructure could come in a separate commit depending on how much restructuring you're doing

CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
...
+
+Please avoid the following in a commit:
+
+* Mixing whitespace changes with functional code changes.
+* Mixing two unrelated functional changes in the same commit.
+* Sending large new features in a single commit.
+
+Expected git commit message structure:
+
+* The first line should be limited to 50 characters, should not end with a full stop and the first word should be capitalised.
```
odyssey4me: This is an arbitrary limit which seems to be derived from https://chris.beams.io/posts/git-commit/ but applied in a draconian way. 50 chars is sometimes not enough. I'd say ideally no more than 50, but definitely no more than 72 (git will wrap it anyway).

CONTRIBUTING.md
odyssey4me: Should we prefix the card number with something? Otherwise it's kinda weird having this weird thing hanging about which will mean nothing if we move to another system. eg: ``JIRA: RE-1500``

CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
...
+* Do not assume the reviewer has access to external web services or site.
+* Do not assume the code is self-evident or self-documenting.
+* Describe why the change is being made.
+* If the commit message suggests an improved code structure, you may be able to split this into two or more commits.
+* Ensure sufficient information is provided to review your commit.
+
+
+### Pull requests
+
+* A pull request (PR) should ideally contain a single commit
```
odyssey4me: Why? A PR can container several commits, each targeting a specific problem to solve - whereas the PR pulls them all together into the final solution. Trying to force one commit per PR puts us in a situation where if one thing is based on another we can't continue work until the previous commit is merged. It also means it's harder to test a group of related fixes together.

CONTRIBUTING.md
mancdaz: @odyssey4me yeah the original rpco commit message guidelines were _inspired_ by that post. I'm not sure about the Draconian nature of wanting good grammar in the subject line, but what do you want the text to say regarding the character limit?

CONTRIBUTING.md
mancdaz: The original intent was to prevent huge scope creep in terms of commits in a PR. I can reword this to be more about making commits related to the problem being solved in the PR

CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
...
+* Where absolutely necessary, related commits can be grouped into a single PR, but this should be the exception, not the rule.
+
+#### Reviewing pull requests
+
+When reviewing a PR, please ensure the included commit(s):
+
+* Actually works to fix the issue.
+* Passes the gate checks that are configured to run.
+* Contains a single commit, which has a commit message describing the changes involved in the patch. On a rare occasion, multiple related commits in the same PR could be considered.
+* Does not overreach. Each commit should be self contained to only address the issue at hand (see above).
```
odyssey4me: Are we really wanting to kill any technical debt management which often comes as a side-effect to working on specific problems? For example, sometimes while working on a problem you notice that the problem is easier to solve by doing some simplification of functions which touch other things that don't directly relate to the problem. As such, technical debt is managed as part of the issue at hand, even though some of that is unrelated. Applying this rule will reduce that effect, making it more likely that we grow tech debt instead of continually reduce it where possible.

CONTRIBUTING.md
odyssey4me: I suggest referencing that post, as it's a good read for background. But then we should rather say something like ``The first line should ideally be limited to 50 characters, but must be no more than 70 characters. It should not end with a full stop and the first word should be capitalised.``

CONTRIBUTING.md
mancdaz: I think we can all use our heads when reviewing and making sure that any refactoring is not excessive and unrelated.

CONTRIBUTING.md
odyssey4me: It is a slightly strange thing to add as a guideline, and is a bit confusing. I'm not sure it's actually adding much value here.

odyssey4me: Perhaps it'd be better to say this:

``Each commit should ideally solve a single problem. Where there are several things to fix to reach the goal identified by the card, multiple commits should be used. Each commit should be as easy as possible to consume as a single, complete change.``

PR #911, opened 2018-06-29
RE-1529 Define RE template for new components

git-harry:
This change adds the structure required by new components, this should
simplify the process of integrating into the RE process. When a new
component is added, a pull request is automatically proposed on the
component master branch which adds the expected files for integrating
with the RE framework. These can then be modified or deleted depending
on the project's specific requirements.

Example pull request: https://github.com/mattt416/rpc-product-1/pull/27

Issue: [RE-1529](https://rpc-openstack.atlassian.net/browse/RE-1529)


generate_component_skeleton.py
```diff
@@ -0,0 +1,194 @@
+#!/usr/bin/env python
+
+import argparse
+from os import chmod, makedirs
+import os.path
+
+check = (
+    "Check jobs are triggered when a pull request is created. Any job can be "
+    "added as a check, the most common use is for linting and basic "
+    "integration tests."
```
hughsaunders: unit?

generate_component_skeleton.py
```diff
@@ -0,0 +1,194 @@
...
+)
+
+gate = (
+    "Gate jobs are triggered by a comment on a pull request. The initiate any "
```
hughsaunders: This initiates?

generate_component_skeleton.py
```diff
@@ -0,0 +1,194 @@
...
+    "gate jobs associated with a repository branch and merge the change if "
+    "all the tests pass. The gate job ensures that changes are tested against "
+    "the state of the branch when the code is merged, in comparison check "
```
hughsaunders: target branch

generate_component_skeleton.py
```diff
@@ -0,0 +1,194 @@
...
+)
+
+release = (
+    "Release jobs are used to validate that a particular commit can be used "
+    "for a new release of the project. They will tend to be the most "
+    "expensive tests in terms of time and resources."
+)
+
+periodic = (
+    "Periodic jobs are trigger by a configurable timer, by default they run "
```
hughsaunders: triggered

generate_component_skeleton.py
```diff
@@ -0,0 +1,194 @@
...
+    "Gate jobs are triggered by a comment on a pull request. The initiate any "
+    "gate jobs associated with a repository branch and merge the change if "
+    "all the tests pass. The gate job ensures that changes are tested against "
+    "the state of the branch when the code is merged, in comparison check "
+    "jobs test against the state of the branch when the test is started."
+)
+
+release = (
+    "Release jobs are used to validate that a particular commit can be used "
+    "for a new release of the project. They will tend to be the most "
```
hughsaunders: ~will~

generate_component_skeleton.py
```diff
@@ -0,0 +1,194 @@
...
+    "Dependency updates are checked periodically. A project should codify the "
+    "the process for updating the working directory of the repository with "
+    "any changes to upstream dependencies. If changes are made, the job will "
+    "automatically create a new pull request."
+)
+
+std_pre = (
+    "pre is an optional executable for preparing the node prior to running "
+    "the build. Examples of actions that should be taken here include things "
+    "that would be performed manually such as setting configuration options."
```
hughsaunders: A note saying this stage is retried on failure would be useful here.

scripts/add_component_skeleton.py
```diff
@@ -0,0 +1,196 @@
...
+)
+
+rel_notes = (
+    "The release notes generated by this step are used when releasing a new "
+    "version of the project. The releases should document the changes between "
+    "the next version and the latest released."
+)
+
+dep_update = (
+    "Dependency updates are checked periodically. A project should codify the "
```
mattt416: `the the`

scripts/add_component_skeleton.py
```diff
@@ -0,0 +1,196 @@
...
+    "Check jobs are triggered when a pull request is created. Any job can be "
+    "added as a check, the most common use is for lint, unit and basic "
+    "functional tests."
+)
+
+gate = (
+    "Gate jobs are triggered by a comment on a pull request. The change is "
+    "merged automatically if all the tests pass. The gate job ensures that "
+    "changes are tested against the state of the target branch when the code "
+    "is merged, in comparison check jobs test against the state of the branch "
```
mattt416: `The gate job ensures that changes are tested against the state of the target branch when the code is merged. In comparison, check jobs test against the state of the branch when the test is started.`

Does that read slightly better?

scripts/add_component_skeleton.py
```diff
@@ -0,0 +1,196 @@
...
+    "RE_HOOK_RESULT_DIR: The directory where test results should be written.\n"
+    "RE_JOB_NAME: The name of the job itself. For example "
+    "PM_repo-name-series-image-scenario-action\n"
+    "RE_JOB_IMAGE: The name of the 'image' as defined in the job.\n"
+    "RE_JOB_SCENARIO: The name of the 'scenario' as defined in the job.\n"
+    "RE_JOB_ACTION: The name of the 'action' as defined in the job.\n"
+    "RE_JOB_FLAVOR: The name of the 'flavor' as defined in the job.\n"
+    "RE_JOB_TRIGGER: The name of the 'trigger' that initiated the build, "
+    "examples include 'PULL', 'PUSH' and 'TIMER'.\n"
+    "RE_JOB_REPO_NAME: The project's GitHub repository which gets cloned in "
```
mattt416: May want to clarify that this is just the repo name, not repo URL.

scripts/add_component_skeleton.py
```diff
@@ -0,0 +1,196 @@
...
+    "RE_HOOK_RELEASE_NOTES: The file where release notes should be written. "
+    "The release notes should be formatted with GitHub-flavoured Markdown.\n"
+    "RE_HOOK_REPO_HTTP_URL: The URL for the repository being released e.g. "
+    "https://github.com/rcbops/rpc-openstack.git."
+)
+
+dep_update_env_vars = (
+    "Environment variables\n"
+    "READ_ONLY_TEST: When set to false a pull request is created to update "
+    "any dependencies that have changed."
```
mattt416: Do we need to expose this? Isn't this specified in the component's dep-update job itself and is used internally by dep_update.sh to determine if PRs, etc. should be created?

pipeline_steps/common.groovy
```diff
@@ -1587,6 +1587,46 @@ List getComponentChange(String baseBranch, String prBranch){
...
+void createComponentSkeleton(String name, String repoUrl, String jiraProjectKey){
+  dir("${WORKSPACE}/${name}"){
+    clone_with_pr_refs(".", repoUrl.replace("https://github.com/", "git@github.com:"), "origin/master")
+    withEnv(
+      [
+        "ISSUE_SUMMARY=Add component skeleton to ${name}",
+        "ISSUE_DESCRIPTION=This issue was generated automatically as part of registering a new component.",
+        "LABELS=component-skeleton",
+        "JIRA_PROJECT_KEY=${jiraProjectKey}",
+        "TARGET_BRANCH=master",
```
mattt416: A component could very well not use master as a branch name. Is this a platform requirement we need to impose, something we need to adjust, or something we deal w/ later when we hit a component that isn't using master?

git-harry: I was going with assume it is and deal with problems later. I think it's a pretty strong convention, maybe it would make sense to be explicit about that.

scripts/add_component_skeleton.py
git-harry: I hadn't appreciated it was set in the JJB project definition. I'll remove it because it doesn't make sense here.

scripts/add_component_skeleton.py
git-harry: done

scripts/add_component_skeleton.py
git-harry: Maybe. I'm happy to change it.

scripts/add_component_skeleton.py
git-harry: done

scripts/add_component_skeleton.py
```diff
@@ -0,0 +1,191 @@
...
+    "configuration used, etc) should be written.\n"
+    "RE_HOOK_RESULT_DIR: The directory where test results should be written.\n"
+    "RE_JOB_NAME: The name of the job itself. For example "
+    "PM_repo-name-series-image-scenario-action\n"
+    "RE_JOB_IMAGE: The name of the 'image' as defined in the job.\n"
+    "RE_JOB_SCENARIO: The name of the 'scenario' as defined in the job.\n"
+    "RE_JOB_ACTION: The name of the 'action' as defined in the job.\n"
+    "RE_JOB_FLAVOR: The name of the 'flavor' as defined in the job.\n"
+    "RE_JOB_TRIGGER: The name of the 'trigger' that initiated the build, "
+    "examples include 'PULL', 'PUSH' and 'TIMER'.\n"
```
mattt416: This could also be `OTHER`, though I'm not entirely sure under which conditions. Also, we have `RE_JOB_TRIGGER_DETAIL` as we discovered on Monday -- but again I'm not even sure anyone is using this or if  there's even value in setting it.

odyssey4me:
recheck_issue_link

scripts/add_component_skeleton.py
```diff
@@ -0,0 +1,193 @@
...
+                "dependencies."
+            ],
+        },
+        {
+            "name": "component_requirements.yml",
+            "docs": [
+                "This file records the specific versions of dependencies that "
+                "a project should use. Put another way it is used to pin "
+                "component dependencies. This file should be updated "
+                "automatically be the dependency update job."
```
odyssey4me: should this be 'by' instead of 'be' ?

git-harry: fixed

PR #916, opened 2018-07-03
RE-1683 Correctly reference remote rc branch

mattt416:
Where we run the `git rev-parse --verify` in
`rpc_jobs/standard_job_release.yml`, no local branch exists tracking
the remote rc branch and therefore the verify fails with
`fatal: Needed a single revision`. This commit prefixes `${RC_BRANCH}`
with `remotes/origin` to correctly validate if the branch exists.

Additionally, we add some log statements to easily identify if we're
releasing from `${MAINLINE}` or `${RC_BRANCH}`.

Issue: [RE-1683](https://rpc-openstack.atlassian.net/browse/RE-1683)

### DIFF
--- a/rpc_jobs/standard_job_release.yml
+++ b/rpc_jobs/standard_job_release.yml
@@ -94,6 +94,7 @@
         MAINLINE=component["release"]["series"]
         RC_BRANCH="${MAINLINE}-rc"
         dir(REPO){
+          println "=== Checking for the existence of an RC branch ==="
           common.clone_with_pr_refs(
             "./",
             component["repo_url"],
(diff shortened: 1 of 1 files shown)


mattt416:
recheck_all

mattt416:
Tried testing this change w/ rpc-component-2 by creating `master-rc`, but I don't think the github user has access to alter branch protection on a personal repo so the release job fails. See https://rpc.jenkins.cit.rackspace.net/job/RE-Release/237/console for more info.

Perhaps we have to move our dummy repos over to rcbops?

PR #918, opened 2018-07-04
Add retries and conditionals to cloudfiles upload

odyssey4me:
In order to improve the chances of the job artifact uploads
succeeding, all operations now have retries added, and more
specific conditions for success.

The rax_pub_cloud fact setting, related conditionals and tasks
for the 'not rax_pub_cloud' case are removed as that is an
unused and untested code path.

The 'Get Rackspace CloudFiles CDN URL' task is removed. The
previous task returns the data required, so this removes an
extra API call and point of failure.

The URI tasks have tightened response codes, with comments
for what each code means. This prevents failures leaking
through and makes it easier to understand what they are.

The URI tasks have 'no_log: True' removed, as they do not
expose any sensitive information unless the playbook is
run with '-vvv' which is not the case here.

As a matter of style and easier debugging, the following is
also done:

1. The vars are moved to the top of the playbook to make them
   more obvious when looking at the playbook.
2. All tasks are named according to what they do.
3. Comments are added/improved to make it clearer what is going
   on.
4. The service catalog data extraction is re-ordered to group
   the two object-store tasks together, and the two object-cdn
   tasks together.
5. All jinja vars are spaced. eg: "{{ object_cdn_url }}" instead
   of "{{object_cdn_url}}".
6. The 'Upload Artifacts to Cloud Files' command is split into
   multiple lines to make it easier to read, and is re-arranged
   to be in the same sequence as the other swift commands later
   in the play.

JIRA: RE-1618

Issue: [RE-1618](https://rpc-openstack.atlassian.net/browse/RE-1618)

### DIFF
--- a/playbooks/upload_to_swift.yml
+++ b/playbooks/upload_to_swift.yml
@@ -2,21 +2,26 @@
 - hosts: localhost
   connection: local
   gather_facts: False
+  vars:
+    region: "DFW"
+    cloud_name: "public_cloud"
+    # 30 days
+    retention: 2592000
   tasks:
+    - name: Set the archive base name
+      set_fact:
+        archive_base_name: "{{ job_name }}_{{ build_number }}"
 
-    - set_fact:
-        archive_base_name: "{{job_name}}_{{build_number}}"
-
-    # This is done so that downloaded archives have a useful name
-    - name: Move artifacts dir
+    - name: Move artifacts dir so that downloaded archives have a useful name
       shell: mv "{{ artifacts_dir | basename }}" "{{ archive_base_name }}"
       args:
         chdir: "{{ artifacts_dir | dirname }}"
 
-    - set_fact:
+    - name: Set the archive directory
+      set_fact:
         adir: "{{ artifacts_dir | dirname }}/{{archive_base_name}}"
 
-    # links are unlikely to work when expanded on a different
+    # Links are unlikely to work when expanded on a different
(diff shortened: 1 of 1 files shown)


playbooks/upload_to_swift.yml
```diff
@@ -44,79 +49,81 @@
       tags:
         - skip_ansible_lint
 
-    - name: Create a public Cloud Files container
+    - name: Create a cloudfiles container
```
mattt416: Total nit, but we still have references to `Cloud Files` and `CloudFiles` below. I'd probably just standardise on `Cloud Files` since I think that's what the product is called.

odyssey4me: Done. That was kinda bothering my OCD too.

playbooks/upload_to_swift.yml
```diff
@@ -44,79 +49,81 @@
...
+      register: _create_container
+      until: _create_container | success
+      retries: 5
+      delay: 10
 
     - name: Authenticate to the cloud and retrieve the service catalog
       os_auth:
         cloud: "{{ cloud_name }}"
         region_name: "{{ region }}"
       no_log: true
```
git-harry: Is this one left because otherwise it will still log sensitive stuff?

odyssey4me: Yes, if it fails it spits out a bunch of things which are sensitive, so I thought it better to leave this as no_log.

PR #919, opened 2018-07-05
RE-1682 Tag SHA not branch

mattt416:
Currently, when we release through the pipeline we're tagging with the
branch (mainline or mainline-rc) instead of the specified SHA. This
commit updates rpc_jobs/standard_job_release.yml to tag the SHA incase
the specified SHA is not the latest commit in the branch.

Issue: [RE-1682](https://rpc-openstack.atlassian.net/browse/RE-1682)


mattt416:
Merging now incase a release is done soon.

PR #921, opened 2018-07-06
Add extra validation conditionals for auth_token

odyssey4me:
When doing authentication, just validating that auth_token
is defined may not be enough. In this patch we make sure
that it is defined and that it is not empty.

JIRA: RE-1618

Issue: [RE-1618](https://rpc-openstack.atlassian.net/browse/RE-1618)

### DIFF
--- a/playbooks/upload_to_swift.yml
+++ b/playbooks/upload_to_swift.yml
@@ -66,7 +66,12 @@
         region_name: "{{ region }}"
       no_log: true
       register: _auth
-      until: (_auth | success) and (auth_token is defined) and (service_catalog is defined)
+      until: (_auth | success) and
+             (auth_token is defined) and
+             (auth_token is not none) and
+             (auth_token | trim != '') and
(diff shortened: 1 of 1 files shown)


playbooks/upload_to_swift.yml
```diff
@@ -66,7 +66,12 @@
         region_name: "{{ region }}"
       no_log: true
       register: _auth
-      until: (_auth | success) and (auth_token is defined) and (service_catalog is defined)
+      until: (_auth | success) and
```
git-harry: I think you could get rid of all the `and` and brackets and use a list which may be a little easier on the eye.
```yaml
until:
  - _auth | success
  - auth_token | default(none) is not none
  - auth_token | trim != ''
  - service_catalog | default(none) is not none
```

odyssey4me: I thought so too, but when I tried that yesterday it failed.

PR #922, opened 2018-07-06
RE-1682 Check SHA before running RELEASE_*

mattt416:
This commit updates the release process to check for the following:

- if the specified SHA is valid
- if the component has an RC branch
- if the tip of the RC branch matches the supplied SHA

Rather than checking this after the RELEASE_* jobs have run, we move
these checks into runReleasesPullRequestWorkflow(). This means an
issue with a SHA will surface before we've wasted time running
RELEASE_* jobs.

Additionally, we correctly reference `componentYaml` in
runReleasesPullRequestWorkflow() rather than `component_text` since
`component_text` is being made available from some outer scope.

Lastly, since we're now checking for the RC branch in
runReleasesPullRequestWorkflow(), we add a bool parameter called
`from_rc_branch` to the `Component-Release` job and correctly pass this
in in `createRelease()`.

Issue: [RE-1682](https://rpc-openstack.atlassian.net/browse/RE-1682)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -1530,23 +1530,23 @@ void runReleasesPullRequestWorkflow(String baseBranch, String prBranch, String j
         "master",
       )
       println "=== Checking that the specified SHA exists ==="
-      Integer no_sha = sh(
+      Boolean sha_exists = ! sh(
         script: """#!/bin/bash -xe
           git rev-parse --verify ${SHA}^{commit}
         """,
         returnStatus: true,
       )
-      if (no_sha) {
+      if (! sha_exists) {
         throw new Exception("The supplied SHA ${SHA} was not found.")
       }
       println "=== Checking for the existence of an RC branch ==="
-      Integer no_rc = sh(
+      Boolean has_rc_branch = ! sh(
         script: """#!/bin/bash -xe
           git rev-parse --verify remotes/origin/${RC_BRANCH}
         """,
         returnStatus: true,
       )
-      if (!no_rc) {
+      if (has_rc_branch) {
         println "=== Checking that the specified SHA is the tip of RC branch ${RC_BRANCH} ==="
         String latest_sha = sh(
           script: """#!/bin/bash -xe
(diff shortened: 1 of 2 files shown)


pipeline_steps/common.groovy
```diff
@@ -1511,13 +1511,61 @@ def findHookDir(String hook_dirs) {
...
+        script: """#!/bin/bash -xe
+          git rev-parse --verify ${SHA}^{commit}
+        """,
+        returnStatus: true,
+      )
+      if (no_sha) {
+        throw new Exception("The supplied SHA ${SHA} was not found.")
+      }
+      println "=== Checking for the existence of an RC branch ==="
+      Integer no_rc = sh(
```
hughsaunders: This is only used negatively eg `!no_rc` which is effectively `not no rc`, so for simplicity maybe it should just be `rc`?

pipeline_steps/common.groovy
```diff
@@ -1513,13 +1513,61 @@ def findHookDir(String hook_dirs) {
...
+        script: """#!/bin/bash -xe
+          git rev-parse --verify ${SHA}^{commit}
+        """,
+        returnStatus: true,
+      )
+      if (! sha_exists) {
+        throw new Exception("The supplied SHA ${SHA} was not found.")
+      }
+      println "=== Checking for the existence of an RC branch ==="
+      Integer has_rc_branch = ! sh(
```
hughsaunders: This needs to be Boolean

https://groovyconsole.appspot.com/script/5089364245741568

pipeline_steps/common.groovy
```diff
@@ -1513,13 +1513,61 @@ def findHookDir(String hook_dirs) {
...
+    String RC_BRANCH="${MAINLINE}-rc"
+
+    dir(REPO_NAME){
+      clone_with_pr_refs(
+        "./",
+        REPO_URL,
+        "master",
+      )
+      println "=== Checking that the specified SHA exists ==="
+      Integer sha_exists = ! sh(
```
hughsaunders: Also needs to be Boolean

pipeline_steps/common.groovy
mattt416: Fixed!

pipeline_steps/common.groovy
mattt416: Fixed!

PR #924, opened 2018-07-06
RE-1559 Nodepool Configuration for Phobos

dealako:
- Added phobos-cloud profile/config entry to nodepool.yaml

JIRA: RE-1559

Issue: [RE-1559](https://rpc-openstack.atlassian.net/browse/RE-1559)

### DIFF
--- a/nodepool/files/nodepool.yml
+++ b/nodepool/files/nodepool.yml
@@ -127,6 +127,49 @@ providers:
         availability-zones: []
         labels: *provider_pools_main_labels_block
 
+  - name: phobos-nodepool
+    region-name: "RegionOne"
+    cloud: phobos_nodepool
+    boot-timeout: 120
+    hostname-format: nodepool-{label.name}-{provider.name}-{node.id}
+    image-name-format: nodepool-phobos-{image_name}-{timestamp}
+    diskimages: *provider_diskimage_block
+    pools:
+      - name: main
+        max-servers: 20
+        availability-zones: []
+        labels:
+          - name: ubuntu-xenial-g1-8
+            diskimage: ubuntu-xenial
+            flavor-name: releng-general1-8
+            key-name: openstack_phobos_nodepool
+            console-log: true
+          - name: ubuntu-trusty-g1-8
+            diskimage: ubuntu-trusty
+            flavor-name: releng-general1-8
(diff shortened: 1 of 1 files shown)


nodepool/files/nodepool.yml
```diff
@@ -127,6 +127,19 @@ providers:
         availability-zones: []
         labels: *provider_pools_main_labels_block
 
+  - name: phobos-cloud
+    region-name: "RegionOne"
+    cloud: phobos-cloud
```
odyssey4me: Given the recent merge, this now needs to be ``phobos_nodepool`` to match up to the entry in clouds.yaml.

nodepool/files/nodepool.yml
```diff
@@ -127,6 +127,19 @@ providers:
...
+    boot-timeout: 120
+    hostname-format: nodepool-{label.name}-{provider.name}-{node.id}
+    image-name-format: nodepool-phobos-{image_name}-{timestamp}
+    diskimages: *provider_diskimage_block
+    pools:
+      - name: main
+        max-servers: 20
+        availability-zones: []
+        labels: *provider_pools_main_labels_block
```
odyssey4me: Is the ``key-name`` for phobos also called ``jenkins`` or something else? If something else, then the ``labels`` block will have to be adjusted to use the right name for phobos.

mancdaz:
recheck_issue_link

nodepool/files/nodepool.yml
odyssey4me: According to https://rpc-openstack.atlassian.net/wiki/spaces/RE/pages/449314836/Phobos+OpenStack the key-name should be ``openstack_phobos_nodepool`` - so either we rename the key, or we have to duplicate a ton of content here. I think it may be best for us to just rename the key to keep the configuration simpler.

Also, I don't see a way for nodepool to use a different private key to do the authentication - so the key we use should be the same across Phobos and Public Cloud.

hughsaunders:
recheck_unit_tests - failed on build gating venv library clone

hughsaunders:
recheck_unit_tests - more clone issues, this time in skip_build_check. Reference repo has been added and clones are going better now, lets try again.

PR #928, opened 2018-07-10
RE-1582 Reduce library clone size

hughsaunders:
There are now two global libraries defined in jenkins, rpc-gating
and rpc-gating-master. rpc-gating does a full clone and fetches
additional refs such as PRs. rpc-gating master does a shallow clone
and does not fetch additional refs. Consequently rpc-gating-master
is faster and uses fewer resources on the master. This commit moves
production standard jobs to use rpc-gating-master. The pre merge
standard job will use rpc-gating when running against the rpc-gating
repo.

Issue: [RE-1582](https://rpc-openstack.atlassian.net/browse/RE-1582)

### DIFF
--- a/job_dsl/rpc_gating_merge_trigger.groovy
+++ b/job_dsl/rpc_gating_merge_trigger.groovy
@@ -1,7 +1,7 @@
 // Can't use env.FOO = {FOO} to transfer JJB vars to groovy
 // as this file won't be templated by JJB.
 // Alternative is to use parameters with JJB vars as the defaults.
-library "rpc-gating@${RPC_GATING_BRANCH}"
+library "rpc-gating-master"
 common.globalWraps(){
 
   // We do not want to trigger any of these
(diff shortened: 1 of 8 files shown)


job_dsl/rpc_gating_merge_trigger.groovy
```diff
@@ -1,7 +1,7 @@
 // Can't use env.FOO = {FOO} to transfer JJB vars to groovy
 // as this file won't be templated by JJB.
 // Alternative is to use parameters with JJB vars as the defaults.
-library "rpc-gating@${RPC_GATING_BRANCH}"
+library "rpc-gating-master"
 common.globalWraps(){
 
   // We do not want to trigger any of these
```
odyssey4me: nit: I guess this check is no longer required and could be removed.

hughsaunders: I think this check can stay, as we still want to avoid triggering from pushes to non-master branches, even if the groovy lib is always loaded from master.

PR #929, opened 2018-07-10
RE-1555 Implement VPN Connectivity

dealako:
- Updated host value to use target_hosts when available instead of localhost
- Added phobos_vpn to Prepare Instances Stage

JIRA: RE-1555

Issue: [RE-1555](https://rpc-openstack.atlassian.net/browse/RE-1555)

### DIFF
--- a/playbooks/phobos_vpn.yml
+++ b/playbooks/phobos_vpn.yml
@@ -1,5 +1,5 @@
-- name: Configure Phobos VPN Tunnel (should only be run on single use slaves)
-  hosts: localhost
+- name: Configure Phobos VPN Tunnel
+  hosts: "{{ target_hosts | default('localhost') }}"
   tasks:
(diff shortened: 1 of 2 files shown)


playbooks/phobos_vpn.yml
```diff
@@ -1,5 +1,5 @@
-- name: Configure Phobos VPN Tunnel (should only be run on single use slaves)
-  hosts: localhost
+- name: Configure Phobos VPN Tunnel
+  hosts: {{ target_hosts | default(‘localhost’) }}
```
odyssey4me: I think this needs to be quoted, ie:

hosts: "{{ target_hosts | default(‘localhost’) }}"

dealako: LOL, will fix. I believed I copied this handy fragment from a Zoom chat session with you.

PR #935, opened 2018-07-12
Verify ssh connectivity for jenkins slaves

odyssey4me:
Sometimes when an instance is created, it is not possible
to ssh to it. This is especially true of OnMetal nodes.

To try and prevent this from affecting tests before the
tests can be migrated to use nodepool, we implement a
try/rescue block so that if either the instance creation
or the ssh connection test fails, the instance is unusable
and a second attempt is done in the fallback region.

Also:

1. The variable 'primary_region_attempts' is removed, as it
   is unused.
2. The variable 'regions_shuf' is renamed to 'regions_shuff'
   so that it is more uniform with 'fallback_regions_shuff'.
3. The difference filter used in 'fallback_regions_shuff' is
   documented, and fixed to use 'regions_shuff' as it should
   be.
4. Extra debug output is added to help diagnose issues which
   may arise with this approach.
5. If there is no valid region list the tasks will fail early
   instead of trying to execute and resulting in an obscure
   error.

Issue: [RE-1517](https://rpc-openstack.atlassian.net/browse/RE-1517)

### DIFF
--- a/playbooks/allocate_pubcloud.yml
+++ b/playbooks/allocate_pubcloud.yml
@@ -4,18 +4,15 @@
   gather_facts: False
   vars:
     keyname: "jenkins"
-    # Each region in the primary regions list will be tried this number of times.
-    primary_region_attempts: 3
     inventory_path: "{{ lookup('env', 'WORKSPACE') }}/rpc-gating/playbooks/inventory"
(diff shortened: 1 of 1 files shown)


odyssey4me:
Nice to see the shuffle working - notice that https://rpc.jenkins.cit.rackspace.net/job/RE-unit-test-slave-types/4702 builds in ORD while https://rpc.jenkins.cit.rackspace.net/job/RE-unit-test-phobos-vpn/1003 builds in DFW.

PR #940, opened 2018-07-13
RE-1655 Build Summary: Update failure types

hughsaunders:
Also:
  * When viewing a category, limit failures to the category being viewed.
  * Exclude venv dirs from lint checks
  * Scan succesful builds - this ensures non-fatal issues are included
    in stats.


scripts/build_summary/failure.py
```diff
@@ -373,6 +391,24 @@ def scan(self):
...
+    category = "C8 RE Infra"
+
+    def scan(self):
+        match_strings = [
+            "Warning: failed to create container",
+            ("This server could not verify that you are authorized to "
+             "access the document you requested"),
+        ]
+        for line in self.build.log_lines:
+            for str in match_strings:
```
git-harry: My editor hates you for shadowing the built-in, maybe `str_` instead?

scripts/build_summary/failure.py
```diff
@@ -373,6 +391,24 @@ def scan(self):
...
+                if str in line:
+                    self.matches = True
+                    self.detail = str.strip()
+                    break
+
```
git-harry: It's not breaking out of the outer loop if there is a match, is that intentional?

scripts/lint_jjb.py
```diff
@@ -318,6 +318,14 @@ def check_timed_trigger(data, in_file):
     rc = 0
     for _dir in args.dirs.split(','):
         for root, dirs, files in os.walk(_dir):
+            skip_dir = False
+            excludes = map(lambda s: s.strip(), args.exclude_files.split(','))
+            for exclude in excludes:
+                if "{s}{ex}".format(s=os.sep, ex=exclude) in root:
```
git-harry: This could potentially catch things that have similar names to excluded directories.

scripts/lint_jjb.py
```diff
@@ -318,6 +318,14 @@ def check_timed_trigger(data, in_file):
     rc = 0
     for _dir in args.dirs.split(','):
         for root, dirs, files in os.walk(_dir):
+            skip_dir = False
+            excludes = map(lambda s: s.strip(), args.exclude_files.split(','))
```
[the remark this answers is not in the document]
hughsaunders: Ahh, methodcaller allows you to map a method, nice thanks!

scripts/build_summary/failure.py
hughsaunders: Replaced double loop with product.

scripts/build_summary/failure.py
hughsaunders: Done

hughsaunders:
@git-harry generated http://rpc-repo.rackspace.com/rpcgating/buildsummarybeta/index.html#/ from this code.

hughsaunders:
recheck_unit_tests - Failed on clone :(

scripts/lint_jjb.py
```diff
@@ -316,16 +317,18 @@ def check_timed_trigger(data, in_file):
...
-        for root, dirs, files in os.walk(_dir):
-            skip_dir = False
-            excludes = map(lambda s: s.strip(), args.exclude_files.split(','))
-            for exclude in excludes:
-                if "{s}{ex}".format(s=os.sep, ex=exclude) in root:
-                    skip_dir = True
-                    break
-            if skip_dir:
-                continue
+        walk_list = list(os.walk(_dir))
```
git-harry: By constructing the list of all files and directories here, it removes the benefit of modifying the directory list and no longer skips files in excluded directories. It's now also only trying to skip them at the top level, is that intentional? I've added a little example below which I think shows the difference.

```
$ ls -1R foo
a
b
c

foo/a:
a.txt

foo/b:
b.txt

foo/c:
c.txt
```

```python
import os                                                                          
                                                                                   
excludes = ["b"]                                                                   
_dir = "/tmp/foo"                                                                  
                                                                                   
walk_list = list(os.walk(_dir))                                                    
_, dirs, _ = walk_list[0]                                                          
for exclude in excludes:                                                           
    try:                                                                           
        dirs.remove(exclude)                                                       
    except ValueError:                                                             
        pass                                                                       
                                                                                   
print("What this code is doing:")                                                  
print("\n".join(str(item) for item in walk_list))                                  
                                                                                   
                                                                                   
new_walk_list = []                                                                 
for root, dirs, files in os.walk(_dir):                                            
    for exclude in excludes:                                                       
        try:                                                                       
            dirs.remove(exclude)                                                   
        except ValueError:                                                         
            pass                                                                   
    new_walk_list.append((root, dirs, files))                                      
                                                                                   
print("\nWhat this code should be doing:")                                         
print("\n".join(str(item) for item in new_walk_list))
```

```
What this code is doing:
('/tmp/foo', ['a', 'c'], [])
('/tmp/foo/a', [], ['a.txt'])
('/tmp/foo/c', [], ['c.txt'])
('/tmp/foo/b', [], ['b.txt'])

What this code should be doing:
('/tmp/foo', ['a', 'c'], [])
('/tmp/foo/a', [], ['a.txt'])
('/tmp/foo/c', [], ['c.txt'])
```

hughsaunders: @git-harry thanks for the clear explanation, I see that the dirs objects in each tuple are independent so my method of only removing it once won't work. I was assuming some reference object, but thats not the case with a string.

hughsaunders:
Addressed @git-harry comment, tided commits, updated beta page again. Ready for review.

PR #954, opened 2018-07-18
RE-1757 Increase Phobos NodePool Max Servers Value

dealako:
- Increased value of phobos max-servers to 5 after VPN/Phobos network setup/configuration validation

JIRA: RE-1757

### DIFF
--- a/nodepool/files/nodepool.yml
+++ b/nodepool/files/nodepool.yml
@@ -136,7 +136,7 @@ providers:
     diskimages: *provider_diskimage_block
     pools:
       - name: main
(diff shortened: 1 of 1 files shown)


dealako:
@mattt416 yes, except for addressing RE-1756.

dealako:
RE-1756 merged, @mattt416 @odyssey4me think we can merge this one as well?

odyssey4me:
@dealako That patch has merged, but the fix has not been applied to the jenkins master yet. Once that's done we should be able to merge this.

PR #961, opened 2018-07-20
Reconfigure apt sources for OnMetal hosts

odyssey4me:
The OnMetal hosts have their images prepared by Ubuntu
and therefore include the default use of Ubuntu mirror
for apt packages. These have turned out to be a bit
flaky so we have opted to reconfigure the hosts with
the Rackspace mirror instead, which we've found in
all our tests to be reliable.

This patch implements that change in configuration, but
also ensures that this is only done on a host which does
not already have python installed. This prevents any
unnecessary execution of the reconfiguration tasks.

Issue: [RE-1758](https://rpc-openstack.atlassian.net/browse/RE-1758)

### DIFF
--- a/playbooks/allocate_pubcloud.yml
+++ b/playbooks/allocate_pubcloud.yml
@@ -127,54 +127,104 @@
       debug:
         msg: "Generated inventory: {{ lookup('file', inventory_path+'/hosts')}}"
 
-    - name: Wait for startup tasks to finish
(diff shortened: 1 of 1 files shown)


odyssey4me:
On a standard cloud instance, all the tasks pertaining to adding the python packages are skipped as they are not necessary:
https://rpc.jenkins.cit.rackspace.net/job/RE-unit-test-slave-types/5061/

PR #964, opened 2018-07-23
RE-1767 Prevent Instance Leak

hughsaunders:
**Ensure retries cleanup**
  The allocate_pubcloud playbook will try in a second region if creating an   
  instance in the first region fails, or if the instance in the first region
  can't be contacted via ssh.

  This commit ensures that any instance created in the first region is
  removed before attempting to create an instance in a second region.

  This commit also ensures that writing the inventory file is the first task
  after each instance creation task so that the cleanup stage has the highest
  chance of success.

  This commit also switches back from wait_for_connection to wait_for
  as wait_for_connection ignores host and port parameters, only checking
  connectivity to the host the task is targeted at.

**Ensure inventory stash happens**
  If an error happens during the playbook that allocates instances, then
  groovy throws an exception and the stash task doesn't happen. If the stash
  doesn't happen then cleanup fails, as cleanup relies on retrieving an
  inventory file from a stash.

  So if the playbook succesfully allocates an instance, then fails, the
  instance will be leaked because cleaup will fail.

  This change adds try/finally to ensure that stash is always attempted
  regardless of playbook success.

Jira:  RE-1767

Issue: [RE-1767](https://rpc-openstack.atlassian.net/browse/RE-1767)

### DIFF
--- a/playbooks/allocate_pubcloud.yml
+++ b/playbooks/allocate_pubcloud.yml
@@ -210,10 +210,10 @@
               exit 1
             fi
             cat << EOF >/etc/apt/sources.list
-            deb ${DISTRO_MIRROR} ${DISTRO__RELEASE} ${DISTRO_COMPONENTS//,/ }
-            deb ${DISTRO_MIRROR} ${DISTRO__RELEASE}-updates ${DISTRO_COMPONENTS//,/ }
-            deb ${DISTRO_MIRROR} ${DISTRO__RELEASE}-backports ${DISTRO_COMPONENTS//,/ }
-            deb ${DISTRO_MIRROR} ${DISTRO__RELEASE}-security ${DISTRO_COMPONENTS//,/ }
+            deb ${DISTRO_MIRROR} ${DISTRO_RELEASE} ${DISTRO_COMPONENTS//,/ }
+            deb ${DISTRO_MIRROR} ${DISTRO_RELEASE}-updates ${DISTRO_COMPONENTS//,/ }
+            deb ${DISTRO_MIRROR} ${DISTRO_RELEASE}-backports ${DISTRO_COMPONENTS//,/ }
+            deb ${DISTRO_MIRROR} ${DISTRO_RELEASE}-security ${DISTRO_COMPONENTS//,/ }
             EOF
           args:
             executable: /bin/bash
--- a/playbooks/allocate_pubcloud.yml
+++ b/playbooks/allocate_pubcloud.yml
@@ -166,9 +166,9 @@
     - name: Check whether python is present on the host
       raw: |
         if [[ -e /usr/bin/python ]]; then
-          echo yes
+          echo -n yes
         else
-          echo no
+          echo -n no
         fi
       register: _python_check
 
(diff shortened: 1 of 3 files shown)


playbooks/allocate_pubcloud.yml
```diff
@@ -59,13 +64,41 @@
...
       rescue:
+        - name: Ensure first attempt instance is removed.
+          os_server:
+            name: "{{ instance_name }}"
+            flavor: "{{ flavor }}"
+            state: absent
+            cloud: "{{ cloud_name }}"
+            region_name: "{{ regions_shuff[0] }}"
+            image: "{{ image }}"
+            key_name: "{{ keyname }}"
```
git-harry: You can probably get rid of a lot of these pairs because they're only needed when creating the instance.

git-harry: Maybe call the clean up playbook here instead of having a task?

hughsaunders:
Removed some keypairs as suggested by @git-harry, left the inventory writes till afterwards as the task is already duplicated twice, and writing it before would require four very similar tasks.

playbooks/allocate_pubcloud.yml
```diff
@@ -178,7 +178,7 @@
 
     - name: Install python onto the host if it is not already present
       when:
-        - "_python_check.stdout | bool"
+        - "_python_check.stdout | trim | bool"
```
odyssey4me: It's always safer to ensure precedence, just in case: ``(_python_check.stdout | trim) | bool``

hughsaunders:
New test build: https://rpc.jenkins.cit.rackspace.net/view/mnaio/job/PM_rpc-openstack-master-xenial_mnaio_no_artifacts-swift-deploy/253/console been sleeping for 1 minute, for 20 minutes... ▶️

mancdaz:
phobos vpn check unit test:

```
TASK [Reconfigure the host to use Rackspace Mirrors] ***************************
changed: [singleuseslave]

TASK [Update apt cache] ********************************************************
FAILED - RETRYING: Update apt cache (3 retries left).
FAILED - RETRYING: Update apt cache (2 retries left).
FAILED - RETRYING: Update apt cache (1 retries left).
fatal: [singleuseslave]: FAILED! => {"attempts": 3, "changed": true, "failed": true, "rc": 100, "stderr": "Shared connection to 104.130.216.141 closed.\r\n", "stdout": "E: Malformed line 5 in source list /etc/apt/sources.list (type)\r\nE: The list of sources could not be read.\r\n", "stdout_lines": ["E: Malformed line 5 in source list /etc/apt/sources.list (type)", "E: The list of sources could not be read."]}
	to retry, use: --limit @/var/lib/jenkins/workspace/RE-unit-test-phobos-vpn/rpc-gating/playbooks/allocate_pubcloud.retry

localhost                  : ok=11   changed=4    unreachable=0    failed=0   
singleuseslave             : ok=4    changed=3    unreachable=0    failed=1   
```

odyssey4me:
recheck_all

odyssey4me:
OK, https://rpc.jenkins.cit.rackspace.net/job/RE-unit-test-slave-types/5146/console shows this working for an onmetal host, so we're good to go.

odyssey4me:
recheck_all

PR #970, opened 2018-07-25
RE-1490 Remove unneeded rpco snapshot jobs

mattt416:
We're currently building newton/newton-rc snapshots for
trusty/trusty_loose_artifacts and xenial/xenial_loose_artifacts but
as only 'loose' artifact mode is being used in production we should
force all jobs to use the 'loose' image and skip building the 'strict'
images.

This commit adds a separate project for newton/newton-rc for snapshots
so we can be specific about which images get built.

Additionally, we bump all existing jobs using snapshots to use the
'loose' newton image, and bump all other snapshots to the latest
released version. Note that for queens, we actually fix the image to
rpc-r17 rather than rpc-r18.

Issue: [RE-1490](https://rpc-openstack.atlassian.net/browse/RE-1490)

### DIFF
--- a/rpc_jobs/rpc_openstack.yml
+++ b/rpc_jobs/rpc_openstack.yml
@@ -453,13 +453,30 @@
     image:
       - xenial:
           IMAGE: "Ubuntu 16.04.2 LTS prepared for RPC deployment"
+      - trusty:
+          IMAGE: "Ubuntu 14.04.5 LTS prepared for RPC deployment"
+    scenario:
+      - "swift"
+    action:
+      - deploy:
+          FLAVOR: "7"
+    jobs:
+      - 'PM_{repo_name}-{branch}-{image}-{scenario}-{action}'
+
+# We create a separate snapshot project since we don't need all newton-aio
+# jobs triggering snapshot builds.
+- project:
+    name: "rpc-openstack-newton-aio-postmerge-snapshot"
+    repo_name: "rpc-openstack"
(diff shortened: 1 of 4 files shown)


mattt416:
recheck_all

odyssey4me:
@mattt416 I think we may have to do this the other way around. Apparently, besides Phobos, every other installation in production is using the 'loose' artifact mode. We should therefore switch all the image creation and consumption to use that mode instead. There is an epic in our backlog which I registered to revise the newton artifact implementation to get rid of the 'strict' mode given it's not used - but it will be some time before that's done.

mattt416:
Thanks for the info @odyssey4me . I've revised the PR with your feedback in mind.

rpc_jobs/rpc_openstack.yml
```diff
@@ -453,13 +453,27 @@
...
+      - "swift"
+    action:
+      - deploy:
+          FLAVOR: "7"
+    jobs:
+      - 'PM_{repo_name}-{branch}-{image}-{scenario}-{action}'
+
+# We create a separate snapshot project since we don't need all newton-aio
+# jobs triggering snapshot builds.
+- project:
```
git-harry: The `CRON: "@monthly"` parameter has been dropped, why was that? It looks like it was there to reduce the frequency of the periodic jobs created for use by the push jobs.

mattt416: Re-added -- it was removed on my first edit but only partially restored after my 2nd edit. Thanks for catching!

PR #972, opened 2018-07-27
RE-1788 Add rpc-upgrades release jobs

mattt416:
This commit adds some rpc-upgrades release jobs, and fine-tunes
skip_pattern for PR jobs.

Issue: [RE-1788](https://rpc-openstack.atlassian.net/browse/RE-1788)

### DIFF
--- a/rpc_jobs/rpc_upgrades.yml
+++ b/rpc_jobs/rpc_upgrades.yml
@@ -10,6 +10,8 @@
       | \.rst$
       | ^releasenotes/
       | ^gating/generate_release_notes/
+      | ^gating/periodic/
+      | ^gating/release/
     image:
       - trusty_aio:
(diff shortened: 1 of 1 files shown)


mattt416:
Hey @antonym @rexredinger, I basically duplicated your functional PR jobs as RELEASE jobs -- the jobs that will get run before a release is cut. Can you confirm if you're happy with this -- or if you'd rather just run a single job, an MNAIO job, or anything else that makes sense for rpc-upgrades?

mattt416:
@antonym @rexredinger Actually, should I exclude kilo_to_newton_leap since that's not a required PR job ?

antonym:
Yeah, I'd drop the Kilo job... the bits are so old it breaks pretty often, thanks!

mattt416:
@antonym sounds good -- fixed !

PR #979, opened 2018-08-02
RE-1325 Create pluggable artifact upload hook

odyssey4me:
In this PR, we create a pluggable artifact upload hook which currently implements two types: 'log' and 'file'.

This new hook makes use of the in-repo ``component_metadata.yml`` file to discover its configuration.

The 'log' type works in exactly the same way as the current job artifact publishing, and has a shim for the ``RE_HOOK_ARTIFACT_DIR`` environment variable for backwards compatibility.

The 'file' type provides the capability to upload any file or recursively upload any folder to public cloud's cloudfiles, in a container name which matches the repository running the test.

We also switch the current log archiving mechanism to use the 'log' type.

There is follow up work in [RE-1813](https://rpc-openstack.atlassian.net/browse/RE-1813) to automatically publish the URL where these files can be accessed to some place that's easy to find.

Issue: [RE-1325](https://rpc-openstack.atlassian.net/browse/RE-1325)
Issue: [RE-1812](https://rpc-openstack.atlassian.net/browse/RE-1812)

### DIFF
--- a/docs/debugging.md
+++ b/docs/debugging.md
@@ -23,27 +23,37 @@ operations such as creating a single use instance for a project's jobs to
 execute on. As the console includes logging for steps not defined by a project,
 it's inefficient to wade through it when looking for the cause of a failure.
 
-## Artifacts
+## Log Artifacts
 While command outputs as viewed in blue ocean or the console are useful, they
-don't tell the whole story. Jobs can produce artifacts which are stored
-for a [set time](https://github.com/rcbops/rpc-gating/blob/master/playbooks/upload_to_swift.yml) after the completion of a job. Any job can produce artifacts
-and use the provided functions to publish those. Jobs artifacts should include
-everything required to debug a build.
+don't tell the whole story. Jobs can be configured to publish log artifacts to
+a Rackspace Cloud Files container, which automatically has CDN and web browsing
+enabled.
 
-To view artifacts, click "Build Artifact Links" in the sidebar from the standard
-Jenkins build page.
+Artifact gathering and publishing is configured by creating a file named
+``component_metadata.yml`` in the root of the git repository which has the
+following format:
 
-### Artifact Example: RPCO
+``` yaml
+"artifacts": [
+  {
+     "type": "log",
+     "source": "/foo"
+  }
+]
+```
+
+The above configuration will upload ``log`` files from every job, sourced from
+the ``/foo`` directory. These will be published and set to expire after 30 days.
+The link to where they are published will be available and displayed
+prominently in the Jenkins job result.
+
+### Log Artifact Example: RPCO
 
 RPC-O builds archive the etc and log dirs of each container and the host.
 This includes the openstack_deploy directory, all the ansible facts, the generated
 configuration files and log files for each service. This comprehensive set of
 artifacts saves times recreating problems as usually the required information
 is stored with the failing job.
-
-
-
-
 
 For AIO jobs, the host is the deploy node, so the openstack_deploy directory
 which contains all the ansible config can be found in the hosts's etc folder.
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -301,7 +301,18 @@ String container_name(){
 def archive_artifacts(Map args = [:]){
   stage('Compress and Publish Artifacts'){
     try{
-      artifacts_dir = args.get("artifacts_dir", "${env.WORKSPACE}/artifacts")
+      // If the build fails, or this is a PR, the default
+      // artifact types to upload are 'log'. Otherwise we
+      // will attempt all types.
+      if ((currentBuild.result == "FAILURE") || (env.ghprbPullId != null)) {
+        suggested_artifact_types = "log"
+      } else {
+        suggested_artifact_types = "all"
+      }
+
+      // However, if an argument is given then use it instead.
+      artifact_types = args.get("artifact_types", suggested_artifact_types)
+
       results_dir = args.get("results_dir", "${env.WORKSPACE}/results")
 
       dir(results_dir) {
@@ -332,10 +343,7 @@ def archive_artifacts(Map args = [:]){
         junit allowEmptyResults: true, testResults: "*.xml"
       }
 
-      pubcloud.uploadToSwift(
-        container: container_name(),
-        path: artifacts_dir
-      )
+      pubcloud.uploadArtifacts("artifact_types": artifact_types)
       if(fileExists(file: "artifact_public_url")){
         artifact_public_url = readFile(file: "artifact_public_url")
         currentBuild.description = "<h2><a href='"+artifact_public_url+"'>Build Artifacts</a></h2>"
@@ -1525,7 +1533,6 @@ void stdJob(String hook_dir, String credentials, String jira_project_key, String
             // try-catch within archive_artifacts() prevents
             // this call from failing standard job builds
             archive_artifacts(
-              artifacts_dir: "${env.RE_HOOK_ARTIFACT_DIR}",
               results_dir: "${env.RE_HOOK_RESULT_DIR}"
             )
         } // try
--- a/pipeline_steps/pubcloud.groovy
+++ b/pipeline_steps/pubcloud.groovy
@@ -234,24 +234,19 @@ def runonpubcloud(Map args=[:], Closure body){
   } //outer try
 }
 
-def uploadToSwift(Map args){
-  if (fileExists(args.path)) {
-    print("Directory ${args.path} found. Uploading contents.")
-    clouds_cfg = common.writeCloudsCfg()
-    withEnv(["OS_CLIENT_CONFIG_FILE=${clouds_cfg}"]){
-      common.venvPlaybook(
-        playbooks: ["rpc-gating/playbooks/upload_to_swift.yml"],
-        vars: [
-          artifacts_dir: args.path,
-          container: args.container,
-          job_name: env.JOB_NAME,
-          build_number: env.BUILD_NUMBER,
-        ]
-      ) // venvPlaybook
-    } // withEnv
-  } else {
-    print("Directory ${args.path} not found. Skipping upload.")
-  }
+def uploadArtifacts(Map args = [:]){
+  print("Uploading artifacts.")
+  clouds_cfg = common.writeCloudsCfg()
+  withEnv(["OS_CLIENT_CONFIG_FILE=${clouds_cfg}"]){
+    common.venvPlaybook(
+      playbooks: [
+        "rpc-gating/playbooks/artifact_upload.yml"
+      ],
+      vars: [
+        artifact_type_include: args.artifact_types
+      ]
+    ) // venvPlaybook
+  } // withEnv
 }
 
 return this
--- a/playbooks/artifact_upload.yml
+++ b/playbooks/artifact_upload.yml
@@ -1,0 +1,73 @@
+---
+- name: Artifact uploader
+  hosts: localhost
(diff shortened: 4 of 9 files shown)


odyssey4me:
recheck_issue_link

docs/debugging.md
```diff
@@ -25,13 +25,50 @@ it's inefficient to wade through it when looking for the cause of a failure.
...
+deployment artifacts. The ``source`` argument can be a file or folder which
+will be uploaded recursively. If a special destination subdirectory in the
+container is desired, the ``dest`` argument can be used to define the relative
+path from the root of the container where the artifacts should be published
+into. The ``expire_after`` argument can also optionally be used to set the
+number of seconds until the artifact expires and will automatically be removed.
+
+To view ``file`` artifacts, navigate to ``<repo name>.re.rpc.rackspace.com``
+(eg: http://rpc-octavia.re.rpc.rackspace.com). If this is not available, please
+contact the RE team to create the convenience alias in DNS for the git
```
hughsaunders: Is there a procedure for this somewhere?

odyssey4me: Not currently. I will be adding it to confluence once this is accepted, and once I have a sample repo to figure out the procedure.

playbooks/artifact_upload.yml
```diff
@@ -0,0 +1,54 @@
...
+        component_artifacts: >-
+          {%- if (_component_metadata_file.stat.exists | bool) and (component_metadata.artifacts is defined) %}
+          {%-   if 'log' in component_metadata.artifacts | json_query('[*].type') %}
+          {%-     set _result = component_metadata.artifacts %}
+          {%-   else %}
+          {%-     set _result = component_metadata.artifacts + legacy_artifact_log_type %}
+          {%-   endif %}
+          {%- else %}
+          {%-   set _result = legacy_artifact_log_type %}
+          {%- endif %}
```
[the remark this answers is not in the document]
odyssey4me: I tried to keep the specifics (uncluding checking if the source exists) of each type in the task files for that type, rather than here. It may be slightly less efficient, but it makes it more obvious why an artifact did not get uploaded in the resulting console output because you're able to see that the type config was picked up (or not), but that it got skipped because the source did not exist. I figure that this is more useful for diagnosing issues than skipping it earlier where the cause of skipping would be less obvious.

odyssey4me:
@hughsaunders The 'file' type artifact is something of a work-in-progress, whereas the 'log' type is an adaptation of the well established existing process (and is tested by the unit tests).

I've been thinking today that perhaps instead of using a convenience CNAME in DNS (which is a manual task by a restricted set of people), we could automatically push up a PR to the release repo to add the CDN URL's once an artifact has been published. This would provide much more transparency and do away with the need for the CNAME.

Regardless, the primary intent of this card is to establish something to publish artifacts more easily. I'd like to get on with the Epic's intent (to publish MNAIO images), then perhaps circle back in the same epic to make the 'file' type publishing more visible and robust.

odyssey4me:
recheck_all

odyssey4me:
Successful build with new job configuration:
https://rpc.jenkins.cit.rackspace.net/job/RE-unit-test-artefact-publication-odyssey4me/6
CDN Container public URL for rpc-gating: https://3532b781e8c3821b0e1d-056fbf9cb875ed1243d1c670da7bbdc4.ssl.cf1.rackcdn.com/

playbooks/tasks/artifact_upload/upload_file.yml
```diff
@@ -0,0 +1,67 @@
+---
+- name: Check that the artifacts source path has been provides
```
hughsaunders: provided

playbooks/tasks/artifact_upload/upload_file.yml
```diff
@@ -0,0 +1,67 @@
...
+    - include: cloudfiles_create_container.yml
+
+    # The ansible os_object module does not currently support setting
+    # the object expiration header field, nor does it do threaded uploads
+    # (which make this much faster), so we use the swift client instead.
+    - name: Upload Artifacts to Cloud Files
+      command: >-
+        swift upload {{ object_store_container_name }} {{ artifact['source'] | basename }}
+        --object-threads 100
+        {{ artifact.expire_after is defined | ternary("--header 'X-Delete-After:" ~ artifact.get('expire_after') ~ "'", "") }}
```
hughsaunders: why `~` instead of `+` ?

playbooks/artifact_upload.yml
```diff
@@ -0,0 +1,73 @@
+---
+- name: Artifact uploader
+  hosts: localhost
+  connection: local
+  gather_facts: False
+  vars:
+    region: "DFW"
+    cloud_name: "public_cloud"
+    component_metadata_file_path: "{{ lookup('env', 'WORKSPACE') }}/component_metadata.yml"
```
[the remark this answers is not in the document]
odyssey4me: It probably should. I know we have ``RE_JOB_REPO_NAME``, but is there perhaps a better way to get the repo name/path?

playbooks/tasks/artifact_upload/upload_file.yml
odyssey4me: ``~`` is for concatenating strings. See http://jinja.pocoo.org/docs/2.10/templates/#math

rpc_jobs/unit/artefact_publish.yml
```diff
@@ -27,6 +27,22 @@
...
+            artifacts: [
+              [
+                type: "file",
+                source: "${env.WORKSPACE}/artifacts",
+                dest: "${env.JOB_NAME}/${env.BUILD_NUMBER}/",
+                expire_after: 86400
+              ]
+            ]
+          ]
+          writeYaml file: 'component_metadata.yml', data: artifacts_data
```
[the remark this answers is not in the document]
odyssey4me: Correct, will fix that - well spotted!

docs/debugging.md
```diff
@@ -23,28 +23,38 @@ operations such as creating a single use instance for a project's jobs to
...
+
+Artifact gathering and publishing is configured by creating a file named
+``component_metadata.yml`` in the root of the git repository which has the
+following format:
+
+``` yaml
+"artifacts": [
+  {
+      "type": "log",
+      "source": "/foo"
```
mattt416: Total nit, but weird indentation here.

odyssey4me: Ah yeah, looks like I used one too many spaces?

pipeline_steps/common.groovy
```diff
@@ -301,7 +301,18 @@ String container_name(){
 def archive_artifacts(Map args = [:]){
   stage('Compress and Publish Artifacts'){
     try{
-      artifacts_dir = args.get("artifacts_dir", "${env.WORKSPACE}/artifacts")
+      // If the build fails, or this is a PR, the default
+      // artifact types to upload are 'log'. Otherwise we
+      // will attempt all types.
+      if ((currentBuild.result == "FAILURE") || (env.'ghprbPullId') == '') {
```
mattt416: I don't think this comment matches the actual code here.

playbooks/artifact_upload.yml
```diff
@@ -0,0 +1,73 @@
...
+    legacy_artifact_log_type:
+      - "type": "log"
+        "source": "{{ lookup('env', 'RE_HOOK_ARTIFACT_DIR') | default(lookup('env', 'WORKSPACE') ~ '/artifacts', true) }}"
+    artifact_type_include: "{{ (lookup('env', 'ghprbPullId') == '') | ternary('all', 'log') }}"
```
mattt416: Can you not use `artifact_type_include` here ?

pipeline_steps/common.groovy
odyssey4me: It looks like it does to me?

playbooks/artifact_upload.yml
odyssey4me: Well yes, I could just leave this out and not define it here - expecting it to be passed in. However, as @hughsaunders noted having it here is useful for testing outside of jenkins.

pipeline_steps/common.groovy
mattt416: If it's a PR, wouldn't `env. ghprbPullId != ''` ?

odyssey4me: Ah, great catch!

playbooks/artifact_upload.yml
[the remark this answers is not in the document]
hughsaunders: I'd use `"${env.WORKSPACE}/${env.RE_JOB_REPO_NAME}"` which is what the clone uses, so you'll definitely end up in the same place..

playbooks/artifact_upload.yml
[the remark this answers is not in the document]
hughsaunders: That would work, simpler and means we only check the trigger in one place.

pipeline_steps/pubcloud.groovy
```diff
@@ -234,24 +234,19 @@ def runonpubcloud(Map args=[:], Closure body){
...
-        ]
-      ) // venvPlaybook
-    } // withEnv
-  } else {
-    print("Directory ${args.path} not found. Skipping upload.")
-  }
+def uploadArtifacts(Map args = [:]){
+  print("Uploading artifacts.")
+  clouds_cfg = common.writeCloudsCfg()
+  withEnv(["OS_CLIENT_CONFIG_FILE=${clouds_cfg}"]){
```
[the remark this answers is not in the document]
hughsaunders: Ah sorry, I should have spotted it was a moved block, no pressure to update it.

playbooks/artifact_upload.yml
mattt416: OK, makes sense. Probably no point passing it in from groovy then?

odyssey4me: We pass it in from groovy for two reasons:

1. To allow our PR unit test to be able to upload more than just logs.
2. To allow any pipeline to set the type they want to upload as they wish, much like we want to for the unit test. :)

rpc_jobs/unit/artefact_publish.yml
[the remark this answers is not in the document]
mattt416: Good catch -- and explains why `component_metadata_file_path: "{{ lookup('env', 'WORKSPACE') }}/component_metadata.yml"` is working in `playbooks/artifact_upload.yml`.

Side note, may be worth trying to test this on a live job to suss out any errors like this.

odyssey4me:
Updated job result: https://rpc.jenkins.cit.rackspace.net/job/RE-unit-test-artefact-publication-odyssey4me/10
Updated CDN result, in the right destination this time: https://3532b781e8c3821b0e1d-056fbf9cb875ed1243d1c670da7bbdc4.ssl.cf1.rackcdn.com/RE-unit-test-artefact-publication-odyssey4me/10/

pipeline_steps/common.groovy
```diff
@@ -301,7 +301,18 @@ String container_name(){
 def archive_artifacts(Map args = [:]){
   stage('Compress and Publish Artifacts'){
     try{
-      artifacts_dir = args.get("artifacts_dir", "${env.WORKSPACE}/artifacts")
+      // If the build fails, or this is a PR, the default
+      // artifact types to upload are 'log'. Otherwise we
+      // will attempt all types.
+      if ((currentBuild.result == "FAILURE") || (env.'ghprbPullId') != '') {
```
mattt416: I think you should actually be checking that `env.'ghprbPullId') != null` ... sorry, didn't catch that the first time around!

PR #981, opened 2018-08-02
RLM-1425 Set logging to be more verbose by default

rexredinger:
Title says it all.

Somewhat related to discussion here:

Issue: [RLM-1425](https://rpc-openstack.atlassian.net/browse/RLM-1425)

### DIFF
--- a/scripts/release.py
+++ b/scripts/release.py
@@ -10,7 +10,8 @@ import git
 import ghutils
 from notifications import mail, mailgun, try_context
 
-logger = logging.getLogger("release")
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger()
 
 
 # release notes


rexredinger:
recheck_issue_link

rexredinger:
recheck_issue_link

hughsaunders:
Many of the scripts that call this supply `--debug` so this won't have too much effect. 

However I think this is a better approach, we should probably remove all the invocations with `--debug` and set level at the top of each script.

scripts/release.py
```diff
@@ -10,7 +10,8 @@
 import ghutils
 from notifications import mail, mailgun, try_context
 
-logger = logging.getLogger("release")
+logging.basicConfig(level=logging.INFO)
```
git-harry: This appears to break the debug logging due to the ordering of `logging.basicConfig` calls between it and ghutils, I believe it only has an effect the first time it is called.

scripts/release.py
```diff
@@ -10,7 +10,8 @@
...
+logger = logging.getLogger()
```
git-harry: Removing the name gives an unhelpful `root` in the log line, `release` is much more descriptive.

PR #984, opened 2018-08-08
Extend release to publish details to Confluence

git-harry:
The purpose of this change is to automate the publishing of releases to
the Confluence wiki.

`release.py` is extended to include the new sub-command
`publish_release_to_wiki`. This updates the table on the page to include
an entry for the new release. If the release already exists on the page,
its details, such as comment, are updated.

If a release is done within 7 days of the first working day of the month
it is considered a scheduled release, if it is outside of that window it
is considered an asynchronous release. The wiki page "Product Releases"
details the differences between these two types of release.

JIRA: RE-1534

Issue: [RE-1534](https://rpc-openstack.atlassian.net/browse/RE-1534)

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -4,6 +4,7 @@ appdirs==1.4.3
 asn1crypto==0.22.0
 Babel==2.4.0
 backports.ssl-match-hostname==3.5.0.1
+beautifulsoup4==4.6.1
 cffi==1.10.0
 click==6.7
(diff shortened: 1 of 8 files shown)


mattt416:
Can you run an async release against a dummy component with this change in place?

git-harry:
@mattt416 https://github.com/rcbops/releases/pull/56

PR #999, opened 2018-08-13
RE-730 Make jenkins label consistent for JIRA cards

jlpogue3:
The original task asked to add a label for cards that were created
automatically.  At the time I began this task, the label "jenkins" had
been added to almost every automated PR/card, but after further
examination, some recent features do not include the "jenkins" label. So
this task now involves simply adding the "jenkins" label to any missing
automated card - for the sake of consistency.

JIRA: RE-730

Issue: [RE-730](https://rpc-openstack.atlassian.net/browse/RE-730)

### DIFF
--- a/scripts/add_component_gate_trigger_job.sh
+++ b/scripts/add_component_gate_trigger_job.sh
@@ -24,6 +24,7 @@ EOF
           --description "${issue_message}" \
           --label COMPONENT_GATE_TRIGGER \
           --label "${COMPONENT_NAME}" \
+          --label jenkins
   )
   echo "Issue: ${issue}"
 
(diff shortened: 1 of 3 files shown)


jlpogue3:
Looks like a failure occurred but not due to my changes.  recheck_unit_tests

scripts/add_component_gate_trigger_job.sh
```diff
@@ -24,6 +24,7 @@ EOF
           --description "${issue_message}" \
           --label COMPONENT_GATE_TRIGGER \
           --label "${COMPONENT_NAME}" \
+          --label jenkins \
```
hughsaunders: Last line of the command shouldn't have `\` appended as the next line is not a continuation.

hughsaunders: Still shouldn't be there IMO.

PR #1002, opened 2018-08-14
RE-1800 Add playbooks/roles/ to .gitignore

mattt416:
The rpc-gating venv contains openstack.logrotate, but it's not present
in .gitignore so anything attempting to do a `git add -A` in
rpc-gating will include this ansible role.

This commit updates .gitignore by adding playbookes/roles and removing
all individual roles specified in playbooks/roles/.

Issue: [RE-1800](https://rpc-openstack.atlassian.net/browse/RE-1800)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,10 +4,7 @@
 .lintvenv
 **/.?venv/
 **/.?lintvenv*/
-playbooks/roles/geerlingguy.nginx/
-playbooks/roles/jnv.unattended-upgrades/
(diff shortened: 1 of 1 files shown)


mattt416:
@odyssey4me I thought about that, but we have our own roles in that dir which we want to keep in git.

mattt416:
@odyssey4me oh wait, that's not true ... looking.

mattt416:
@odyssey4me updated!

PR #1006, opened 2018-08-16
Implement our own ansible.cfg at the highest precedence

odyssey4me:
The various tests which run may include implementing an ansible.cfg
file in the default search paths that ansible uses, resulting in
our playbooks trying to load/use things which are not meant to be.

To combat this, we ensure that we have our own ansible.cfg file and
that we use the environment variable to inform ansible where it should
find it so that it does not search other paths.

We also move ANSIBLE_HOST_KEY_CHECKING and ANSIBLE_FORCE_COLOR into
ansible.cfg so that the settings are easier to find in one place.

Issue: [RE-1870](https://rpc-openstack.atlassian.net/browse/RE-1870)


pipeline_steps/common.groovy
```diff
@@ -106,7 +106,7 @@ def venvPlaybook(Map args){
       sh """#!/bin/bash -x
         which scl && source /opt/rh/python27/enable
         set +x; . ${env.WORKSPACE}/.venv/bin/activate; set -x
-        export ANSIBLE_HOST_KEY_CHECKING=False
+        export ANSIBLE_CONFIG=${env.WORKSPACE}/rpc-gating/ansible.cfg
```
hughsaunders: This path doesn't match the location of the file. Either move the file to root of the repo or add `/playbooks/` here.

odyssey4me:
recheck_unit_tests

PR #1018, opened 2018-08-22
RE-1624 Add ability to comment on new PRs

mattt416:

1. Renames the rpc_jobs/pull_request_issue_link.yml file to
rpc_dating/pull_request_whisperer.yml to make it more generic for
functions that need to run when a PR is created. We also add a call to
github.add_comment_to_pr() here so we can post a mini howto on a PR
when a PR is created. Lastly we add some template variables that allow
different bits of functionality to run depending on the needs of each
project.

2. Moves out the projects defined in rpc_jobs/pull_request_whisperer.yml
to their respective project files (ie. rpc_gating/rpc_openstack.yml).

3. Adds a new function common.createPrWhisperer() which is used
to create whisperer projects when a new component is added to
rcbops/releases.

4. Adds a `failure-status` to the `github-pull-request` trigger in
standard_job_premerge.yml so that a failed job will have its failure
status message set to indicate how to re-run the failed job.

Issue: [RE-1624](https://rpc-openstack.atlassian.net/browse/RE-1624)


mattt416:
Hey @hughsaunders , any thoughts on this approach?

scripts/ghutils.py
```diff
@@ -74,6 +74,42 @@ def create_issue(repo, tag, link, label):
...
+    required=True,
+)
+@click.option(
+    '--body',
+    help="The body of the comment to post to the issue",
+    required=True,
+)
+@click.option(
+    '--allow-duplicates',
+    help='Retry the merge on failure.',
```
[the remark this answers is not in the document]
mattt416: Oops -- fixing now.

scripts/ghutils.py
```diff
@@ -74,6 +74,42 @@ def create_issue(repo, tag, link, label):
...
+    help="Issue to update",
+    required=True,
+)
+@click.option(
+    '--body',
+    help="The body of the comment to post to the issue",
+    required=True,
+)
+@click.option(
+    '--allow-duplicates',
```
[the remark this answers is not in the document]
mattt416: Good catch. Here allow_duplicates always gets set to a string, so this isn't right. Fixing now.

scripts/ghutils.py
```diff
@@ -74,6 +74,42 @@ def create_issue(repo, tag, link, label):
...
+                found = True
+                break
+
+    if allow_duplicates or not found:
+        success = issue.create_comment(body)
+        if success:
+            click.echo("Pull request updated with issue reference.")
+        else:
+            click.echo("There was a failure commenting on the issue.")
+            raise SystemExit(1)
```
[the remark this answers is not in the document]
mattt416: Agreed -- but noticed that seemed to be what was being done in ghutils.py already, so wasn't sure if I should add an additional import for sys. Thoughts?

scripts/ghutils.py
```diff
@@ -74,6 +74,42 @@ def create_issue(repo, tag, link, label):
...
+    if not allow_duplicates:
+        for comment in issue.iter_comments():
+            if comment.body == body:
+                found = True
+                break
+
+    if allow_duplicates or not found:
+        success = issue.create_comment(body)
+        if success:
+            click.echo("Pull request updated with issue reference.")
```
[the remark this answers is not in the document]
mattt416: Copy-paste fail, fixing.

scripts/ghutils.py
```diff
@@ -74,6 +74,42 @@ def create_issue(repo, tag, link, label):
...
+        else:
+            click.echo("There was a failure commenting on the issue.")
+            raise SystemExit(1)
+
```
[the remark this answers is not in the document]
mattt416: Done.

hughsaunders:
As I said in my review, I'm not too keen on adding extra projects to JJB, as its going to become unwieldy, I don't have a strong opinion about whether this gets merged into issue link (probably with a rename) or into the standard jobs with some extra logic for deduplication.

mattt416:
@hughsaunders @git-harry I'll have a look at merging this w/ the issue link job -- I'm less keen on putting this into the premerge standard job as it'd mean repos would have to have check jobs for any of this to actually work.

mattt416:
@git-harry @hughsaunders updated by merging this commenter code into the issue-linker. I've still got a touch more testing to do, but can you give it a once-over and confirm you're happier with this approach?

rpc_jobs/pull_request_whisperer.yml
```diff
@@ -0,0 +1,78 @@
+- project:
+    name: 'Pull-Request-Whisperer-Jobs'
+    series:
+      - all_branches:
+          branches: ".*"
+    repo:
+      - rpc_openstack:
```
git-harry: It would probably make sense to separate this into the project files.

rpc_jobs/pull_request_whisperer.yml
```diff
@@ -0,0 +1,78 @@
...
+          repo_url: "https://github.com/rcbops/rpc-gating"
+          run_issue_linker: true
+      - releases:
+          repo_url: "https://github.com/rcbops/releases"
+          run_issue_linker: true
+          comment_body: |
+            Thank you for pull request to `rcbops/releases`!
+
+            Once all `check` jobs have passed and the pull request has received a
+            sufficient amount of reviews, the repository's `release` jobs can be triggered by
```
git-harry: "Amount" sounds a little strange to me, maybe "number"?

mattt416: Fair enough, fixing.

rpc_jobs/pull_request_whisperer.yml
mattt416: Sure, splitting these out.

mattt416:
Re-tested the component registration [here](https://rpc.jenkins.cit.rackspace.net/job/Component-Release-Trigger/169/).

rpc_jobs/pull_request_whisperer.yml
```diff
@@ -0,0 +1,51 @@
...
+    project-type: workflow
+    concurrent: true
+    run_commenter: true
+    run_issue_linker: false
+    comment_body: |
+      Thank you for submitting a pull request to this project!
+
+      Once all `check` jobs have passed and the pull request has received a
+      sufficient number of reviews, the repository's `gate` jobs can be triggered by
+      adding a comment of `check_component_gate` to this pull request.
```
hughsaunders: I think this will trigger the gate job. `trigger-phrase: '(re)?check_component_gate'`

git-harry: @hughsaunders the code appears to [match not find](https://github.com/jenkinsci/ghprb-plugin/blob/4e86ed47a96a01eeaa51a479ff604252109635f6/src/main/java/org/jenkinsci/plugins/ghprb/Ghprb.java#L102) and so won't trigger the job. I've tested this to confirm.

PR #1023, opened 2018-08-24
Segment large file artifacts when uploading them

odyssey4me:
Cloud Files only supports file sizes up to 5GB [1]. Anything larger
must be segmented. When serving these files via the CDN address
they are still delivered as a single file, so segmentation provides
the advantage of a faster upload and downloads still work as one
would expect by delivering a single large file. This ensures that
we are able to upload large files (such as VM images) and consume
them.

[1] https://support.rackspace.com/how-to/cloud-files-uploading-large-files/

Issue: [RE-1561](https://rpc-openstack.atlassian.net/browse/RE-1561)

### DIFF
--- a/playbooks/tasks/artifact_upload/upload_file.yml
+++ b/playbooks/tasks/artifact_upload/upload_file.yml
@@ -65,12 +65,16 @@
 
     # The ansible os_object module does not currently support setting
     # the object expiration header field, nor does it do threaded uploads
-    # (which make this much faster), so we use the swift client instead.
+    # (which make this much faster), nor does it support setting a segment
+    # size (for large files), so we use the swift client instead. We use
+    # a segment size of 1GB to ensure that large files are segmented to
+    # less than the 5GB limit implemented by Cloud Files.
     - name: Upload Artifacts to Cloud Files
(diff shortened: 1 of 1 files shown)


playbooks/tasks/artifact_upload/upload_file.yml
```diff
@@ -65,12 +65,16 @@
...
+    # (which make this much faster), nor does it support setting a segment
+    # size (for large files), so we use the swift client instead. We use
+    # a segment size of 1GB to ensure that large files are segmented to
+    # less than the 5GB limit implemented by Cloud Files.
     - name: Upload Artifacts to Cloud Files
       command: >-
         swift upload {{ object_store_container_name }} {{ artifact_basename }}
         --object-threads 100
         --skip-identical
+        --segment-size 1073741824
```
hughsaunders: I'd make that 4.5GB to reduce the number of chunks, not critical though.

odyssey4me: I could do that, although that may make downloading the swift client slower. We'll have to see how well this works in practice.

PR #1024, opened 2018-08-27
RE-1905 Revert multi hook dir code

mattt416:
We've removed the deprecated hook dirs, so this code is no longer
necessary.

PR #1025 will need to merge for this PR to gate successfully.

Issue: [RE-1905](https://rpc-openstack.atlassian.net/browse/RE-1905)


mattt416:
recheck_xenial_lint_test

PR #1030, opened 2018-08-28
Resolve artifact upload error when no dest is specified

odyssey4me:
When no destination is provided, the set_fact fails to resolve
because the split function is still evaluated by Ansible at run
time. In this patch we do the following:

1. We switch to using a regex_replace filter instead, which
   should only get evaluated if the dest is defined. The regex
   will result in '/foo/bar' or 'foo/bar' returning 'foo'.
2. We add the configuration to the unit test to verify that
   this works.
3. We resolve a bug where if the source was 'foo' and the dest
   was 'bar/baz', the resulting upload would go into 'bar/baz/foo'
   which is not what was intended. With the change in the task
   which handles the move of content, it will now go into the
   destination path 'bar/baz' as intended.

Issue: [RE-1901](https://rpc-openstack.atlassian.net/browse/RE-1901)

### DIFF
--- a/playbooks/tasks/artifact_upload/upload_file.yml
+++ b/playbooks/tasks/artifact_upload/upload_file.yml
@@ -46,18 +46,26 @@
 
     - name: Create the destination path and move the source file(s) into it
       shell: |
-        mkdir -p {{ artifact['dest'] }}
-        mv {{ artifact['source'] | basename }} {{ artifact['dest'] }}/
+        mkdir -p {{ artifact['dest'] | dirname }}
+        mv -v {{ artifact['source'] | basename }} {{ artifact['dest'] }}
       args:
         warn: no
         executable: /bin/bash
         chdir: "{{ artifact['source'] | dirname }}"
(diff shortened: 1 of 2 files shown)


odyssey4me:
Reconfigured unit test showing that it all succeeds: https://rpc.jenkins.cit.rackspace.net/job/RE-unit-test-artefact-publication/1646/

odyssey4me:
Reconfigured artifact publication job showing everything working: https://rpc.jenkins.cit.rackspace.net/job/RE-unit-test-artefact-publication/1650/
The resulting artifacts with the destination set: https://3532b781e8c3821b0e1d-056fbf9cb875ed1243d1c670da7bbdc4.ssl.cf1.rackcdn.com/RE-unit-test-artefact-publication/1650/
The resulting artifacts with no destination set: https://3532b781e8c3821b0e1d-056fbf9cb875ed1243d1c670da7bbdc4.ssl.cf1.rackcdn.com/file_artifacts_no_dest/

PR #1040, opened 2018-08-30
RE-1494 Platform Maintenance Window

hughsaunders:
* Add groovy functions for checking maintenance window overlap
* Add notifications for cancelled jobs
    * jira - for RE to collect all cancelled jobs that relate
      to a specific maintenance window
    * PR - to explain why the job was cancelled.
* Add call to abortIfMaintenanceWindowPending in common.globalWraps
* Add lint test to check that no scheduled jobs conflict with
  the maintenance window or its preceeding drain period.

Issue: [RE-1494](https://rpc-openstack.atlassian.net/browse/RE-1494)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -1,7 +1,14 @@
 import groovy.json.JsonSlurperClassic
 import groovy.json.JsonOutput
+import groovy.json.JsonException
 import org.jenkinsci.plugins.workflow.job.WorkflowJob
 import org.jenkinsci.plugins.workflow.job.WorkflowRun
+import java.time.LocalDateTime
+import java.time.DayOfWeek
+import java.time.Duration
+import java.time.ZoneOffset
+import java.time.format.DateTimeFormatter
+import java.math.MathContext
 
 void download_venv(){
   sh """#!/bin/bash -xeu
@@ -483,7 +490,7 @@ String clone_with_pr_refs(
   }
   if(ref == "origin/pr/null/merge"){
     throw new Exception(
-      "ref not supplied to common.clone_with_pr_refs or env.ghprbPullID not "\
+      "ref not supplied to common.clone_with_pr_refs or env.ghprbPullId not "\
       + "set, attempting to checkout PR for a periodic build?")
   }
   String sha
@@ -1395,6 +1402,14 @@ void globalWraps(Closure body){
     timeout(time: 10, unit: 'HOURS'){
       shared_slave(){
         wrap([$class: 'LogfilesizecheckerWrapper', 'maxLogSize': 200, 'failBuild': true, 'setOwn': true]) {
+          setTriggerVars()
+          // This test must be outside a function so that return can be used to abort the job
+          // Alternatively we could throw a custom exception and abort when catching it here.
+          if (issueExistsForNextMaintenanceWindow() && willOverlapMaintenanceWindow()){
+            common.recordAbortDueToMaintenance()
+            currentBuild.result == "ABORTED"
+            return
+          }
           print("common.globalWraps pre body")
           body()
           print("common.globalWraps post body")
@@ -1443,9 +1458,6 @@ return this
 
 void stdJob(String hook_dir, String credentials, String jira_project_key, String wrappers) {
   globalWraps(){
-    // set env.RE_JOB_TRIGGER & env.RE_JOB_TRIGGER_DETAIL
-    setTriggerVars()
-
     standard_job_slave(env.SLAVE_TYPE) {
       wrapList(stdJobWrappers(wrappers)){
         env.RE_HOOK_ARTIFACT_DIR="${env.WORKSPACE}/artifacts"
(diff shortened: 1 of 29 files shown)


scripts/lint_jjb.py
```diff
@@ -176,144 +181,230 @@ def check_retention(job, in_file):
...
+        "timed_triggers": "triggers[*].timed"
+    }
 
-    # The variable cron_value must have
-    # a value to be tested, otherwise
-    # there is no CRON parameter in the
-    # job so we pass the test.
-    if (cron_value and standard_pm_template in jobs_list):
+    schedules = []
+    for exp_label, jmes_expression in jmes_expressions.items():
```
git-harry: s/exp_label/_/ or `jmes_expressions.values()`

pipeline_steps/common.groovy
```diff
@@ -1969,3 +1974,126 @@ void updateStringParam(String name, String value, String description){
...
+}
+
+/**
+* Get date object representing the start of the next maintenance window
+*/
+LocalDateTime getNextMaintenanceWindowStart(){
+  // set start of maintenance window
+
+  DayOfWeek maintDay = DayOfWeek.MONDAY
+  Integer maintHour = 10
```
git-harry: What happens if the maintenance takes more than an hour? Have you written the process down for how this will work? Are there manual step to prevent and allow new builds?

scripts/lint_jjb.py
```diff
@@ -176,144 +181,230 @@ def check_retention(job, in_file):
...
+    maint_duration = timedelta(hours=2)
+
+    # Find the interval between builds
+    build_iter = croniter(build_schedule, base)
+    next_build = build_iter.get_next(datetime)
+    second_build = build_iter.get_next(datetime)
+    build_interval = second_build - next_build
+
+    # check the next 60 maintenance windows for conflicts
+    # this is over a year so should cover most eventualities.
```
git-harry: The variable in this case is the day of the month field. Maybe instead you could put some constraints around that field and then you don't need to be concerned about it varying with time.

rpc_jobs/standard_job_postmerge.yml
```diff
@@ -44,7 +44,7 @@
     concurrent: true
     FLAVOR: "performance1-1"
     IMAGE: "Ubuntu 16.04 LTS (Xenial Xerus) (PVHVM)"
-    CRON: "@daily"
+    CRON: "H H(13-23) * * *"
```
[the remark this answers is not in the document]
hughsaunders: The other options is 24-hours, but skipping a day (See the wiki page), can use that if preferable, depends how strictly you want to interpret "daily".

scripts/jirautils.py
```diff
@@ -53,8 +53,9 @@ def _get_or_create_issue(project, status, labels, description, summary):
 
     # Check for existing issues
     query = ("{label_terms} AND project = \"{p}\" AND status = {s}"
+             " AND summary ~\"{summary}\""
```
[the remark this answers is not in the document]
hughsaunders: yep

pipeline_steps/common.groovy
```diff
@@ -1969,3 +1974,126 @@ void updateStringParam(String name, String value, String description){
...
+
+  LocalDateTime completion = getPredictedCompletionTime()
+  LocalDateTime maintWindowStart = getNextMaintenanceWindowStart()
+  LocalDateTime prBufferStart = maintWindowStart.minus(prBuffer)
+
+  return prBufferStart.isBefore(completion)
+}
+
+void abortIfMaintenanceWindowPending(){
+  if (willOverlapMaintenanceWindow()){
```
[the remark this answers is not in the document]
hughsaunders: I would like to avoid scheduled jobs being aborted, as we want to increase reliability and trust. Thats my reasoning for linting the schedules. 

However I do like the idea of using an issue to signify that we have something to do in a future maintenance window. Then we can avoid aborting PR jobs if theres nothing to do in a particular week.

scripts/lint_jjb.py
```diff
@@ -176,144 +181,230 @@ def check_retention(job, in_file):
...
-def check_timed_trigger_list(data, in_file):
-
-    timed_trigger_list = jmespath.search('triggers[*].timed', data)
+    if not isinstance(schedule, str):
+        return 1
 
-    return_value = 0
-    for trigger in timed_trigger_list or []:
+    # A comment can't be invalid.
+    if is_comment(schedule):
```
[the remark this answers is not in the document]
hughsaunders: Jenkins is fine with "#foo" as cron value - it just ignores it. So it depends if yaml or jenkins is interpreting the #.

pipeline_steps/common.groovy
hughsaunders: The maintenance window is 2 hours, and we'll have to enable quiet mode manually.

pipeline_steps/common.groovy
```diff
@@ -1969,3 +1974,126 @@ void updateStringParam(String name, String value, String description){
...
 }
+
+/**
+* Get the predicted duration for the current build in milliseconds.
+* Prediction is based on up to 6 recent builds.
+* See: https://javadoc.jenkins-ci.org/hudson/model/Job.html#getEstimatedDurationCandidates--
+* NonCPS because build is not serialisable
+*/
+@NonCPS
+Duration getPredictedDuration(){
```
dealako: I like, it's clean.  But would be more 'functional' or re-usable if we passed in the parent job as an input parameter. Meh, either way works.  This is clean enough.  Just a thought. Or perhaps renaming to getCurrentBuildPredictedDuration() to properly describe what it does, but that's too long of a method name.  Keep it as-is - just thinking out loud.

pipeline_steps/common.groovy
hughsaunders: I looked into creating a job for, this but it seems quiet mode currently has problems with pipeline: https://issues.jenkins-ci.org/browse/JENKINS-34256

scripts/lint_jjb.py
hughsaunders: Done

scripts/lint_jjb.py
```diff
@@ -176,144 +181,230 @@ def check_retention(job, in_file):
...
 
-    return_value = 0
-    if check_timed_trigger_value(cron_value) == 1:
-        sys.stderr.write(error_message)
-        return_value = 1
+    return schedule
 
-    return return_value
 
+def is_valid_cron_statement(schedule, name):
```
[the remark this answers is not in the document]
hughsaunders: Done

scripts/lint_jjb.py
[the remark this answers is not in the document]
hughsaunders: Thinking about it further, if YAML interpreted the `#` then at this point the value would be `None`, so to get a `#` this far it must have been quoted or escaped.

pipeline_steps/common.groovy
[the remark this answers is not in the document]
hughsaunders: Added `issueExistsForNextMaintenanceWindow()` and will only abort if that is true, great suggestion thanks. Also added a job for creating the maintenance window that has no triggers so we can run it manually when we need a window.

pipeline_steps/common.groovy
```diff
@@ -1969,3 +1974,144 @@ void updateStringParam(String name, String value, String description){
...
+    "BACKLOG",
+    "${nextWindowDateString} Maintenance Window",
+    "Jira issue to collect notes relating to a maintenance window, for more information please see https://rpc-openstack.atlassian.net/wiki/spaces/RE/pages/469794817/RE+Infrastructure+Maintenance+Window",
+    ["jenkins", "re", "maintenance"]
+  )
+  return jiraIssue
+}
+
+void abortIfMaintenanceWindowPending(){
+  if (issueExistsForNextMaintenanceWindow() && willOverlapMaintenanceWindow()){
```
git-harry: It would be nice if this allowed an ad-hoc maintenance to be scheduled outside of the normal window.

scripts/lint_jjb.py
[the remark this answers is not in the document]
git-harry: Why not just leave this to croniter, it seems unnecessary to try to determine if the expression is actually commented out when either way it's an invalid value. It also seems a strange way to try to add a comment, why would the whole line be commented out.

rpc_jobs/standard_job_postmerge.yml
```diff
@@ -44,7 +44,7 @@
     concurrent: true
     FLAVOR: "performance1-1"
     IMAGE: "Ubuntu 16.04 LTS (Xenial Xerus) (PVHVM)"
-    CRON: "@daily"
+    CRON: "{CRON_DAILY}"
```
git-harry: For some reason I couldn't respond to your comment. It would be nice to restrict the impact to Mondays only. Could two expressions be supplied, one for limiting the hours on Monday and the other allowing any time of the day?Although if it is a problem, having it still impact work on a Monday might still be a bad thing. Maybe it's easier to just see if this causes an issue and address it then

pipeline_steps/common.groovy
```diff
@@ -1969,3 +1974,144 @@ void updateStringParam(String name, String value, String description){
...
+
+    if(env.RE_JOB_TRIGGER == "PULL"){
+      // Build triggered by a PR
+
+      String comment = "Build ${env.BUILD_URL} was aborted as it's estimated completion time overlapped an RE maintenance window. See Jira Issue: ${jiraLink}"
+      github.add_comment_to_pr(comment)
+
+      // Set pull link string that will be embedded in a Jira comment,
+      // so that RE team members reviewing the maintenance ticket can
+      // easily find affected PRs.
```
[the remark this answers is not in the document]
hughsaunders: Good thinking, lets do that.

rpc_jobs/defaults.yml
```diff
@@ -24,3 +24,6 @@
     JENKINS_NODE_EXECUTORS: "2"
     jenkins_node_exclusive: true
     allow_jenkins_sudo: true
+    CRON_DAILY: "H H * * 2-7"
+    CRON_WEEKLY: "{CRON_WEEKLY}"
+    CRON_MONTHLY: "{CRON_MONTHLY}"
```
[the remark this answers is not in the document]
hughsaunders: Find and replace fail, oops.

pipeline_steps/common.groovy
```diff
@@ -1969,3 +1974,144 @@ void updateStringParam(String name, String value, String description){
...
+    "${nextWindowDateString} Maintenance Window",
+    "Jira issue to collect notes relating to a maintenance window, for more information please see https://rpc-openstack.atlassian.net/wiki/spaces/RE/pages/469794817/RE+Infrastructure+Maintenance+Window",
+    ["jenkins", "re", "maintenance"]
+  )
+  return jiraIssue
+}
+
+void abortIfMaintenanceWindowPending(){
+  if (issueExistsForNextMaintenanceWindow() && willOverlapMaintenanceWindow()){
+    // This doesn't seem to have any effect when using error()
```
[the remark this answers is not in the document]
hughsaunders: yes :(

pipeline_steps/common.groovy
```diff
@@ -1969,3 +1974,144 @@ void updateStringParam(String name, String value, String description){
...
+    // point if a matching issue doesn't exist.
+    String jiraIssue = getOrCreateIssueForNextMaintenanceWindow()
+
+    print("Jira Issue Key: ${jiraIssue}")
+    String jiraLink="https://rpc-openstack.atlassian.net/browse/${jiraIssue}"
+
+    if(env.RE_JOB_TRIGGER == "PULL"){
+      // Build triggered by a PR
+
+      String comment = "Build ${env.BUILD_URL} was aborted as it's estimated completion time overlapped an RE maintenance window. See Jira Issue: ${jiraLink}"
```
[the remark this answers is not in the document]

rpc_jobs/bump_snapshots.yml
```diff
@@ -8,7 +8,7 @@
     parameters:
       - rpc_gating_params
     triggers:
-      - timed: "@daily"
+      - timed: "H H * * 2-7" # @daily // can't use {CRON_DAILY} as this isn't a template
```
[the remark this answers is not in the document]
hughsaunders: Done

rpc_jobs/standard_job_postmerge.yml
hughsaunders: Two expression can't be supplied within the macro, but each job could have:
```
 triggers:
   - timed: "{CRON_DAILY}"
   - timed: "{CRON_MONDAY}"
```

Or something like that. However I think that daily meaning the job runs 6/7 days will be fine for most jobs. Any that do have a problem can use the other pattern as [explained on the maintenance confluence page](https://rpc-openstack.atlassian.net/wiki/spaces/RE/pages/469794817/RE+Infrastructure+Maintenance+Window).

pipeline_steps/common.groovy
hughsaunders: Created RE-1958 for this

hughsaunders:
Tested the abort and resume functions with the following pipeline: 

```
library "rpc-gating@origin/pr/1040/merge"
common.globalWraps(){
    // set PR vars (these would be set already if the job was PR triggered)
    env.ghprbGhRepository =  "rcbops/rpc-gating"
    env.ghprbPullId = "1070"
    env.ghprbPullLink = "https://github.com/rcbops/rpc-gating/pull/1070"
    env.RE_JOB_TRIGGER = "PULL"
    common.recordAbortDueToMaintenance()
    common.restartAbortedPRBuilds()
}
```

Related PR: https://github.com/rcbops/rpc-gating/pull/1070
Related Issue: https://rpc-openstack.atlassian.net/browse/RE-1959
Successful Build: https://rpc.jenkins.cit.rackspace.net/job/scratchpipeline/1143/

pipeline_steps/github.groovy
```diff
@@ -70,12 +70,33 @@ void add_issue_url_to_pr(){
...
-  List org_repo = env.ghprbGhRepository.split("/")
+/**
+ * If called with add_comment_to_pr("comment"), then comment will be added to
+ * the PR that triggered this job (using ghprb env vars to discover the pr).
+ *
+ * To add comments to an alternate pr, supply repo in the form "org/repo" and
+ * prid as "id" where id is the pull request number.
+ */
+void add_comment_to_pr(String body,
+                       String repo_url="use_env",
```
git-harry: It might be simpler to do something like `String repo_url=env.ghprbGhRepository,`, it would then get set to `null` if it wasn't a PR job and you could test them for `null`.

pipeline_steps/github.groovy
```diff
@@ -70,12 +70,33 @@ void add_issue_url_to_pr(){
...
+                       String prid="use_env",
+                       Boolean allowDuplicates=false){
```
git-harry: I like how you've added three parameters, one in snake_case, one in alllowercase and one in camelCase.

pipeline_steps/github.groovy
```diff
@@ -70,12 +70,33 @@ void add_issue_url_to_pr(){
...
+  if ( repo_url == "use_env"){
+    repo_url = env.ghprbGhRepository
+  }
+  if (prid == "use_env"){
+    prid = env.ghprbPullId
+  }
```
git-harry: Having them independently able to use the ghprb values allows for bugs, for example if only the repo URL was overridden on a pull request job, what would that mean? Pull request IDs are not shared so it is almost certain to reference the wrong pull request.

PR #1043, opened 2018-08-31
RE-1920 RPC OSP Credentials

hughsaunders:
These credentials are already stored in the jenkins credentials
store, this commit exposes them via the RE JJB mechanism for requesting
credentials.


hughsaunders:
Unit test failed on jira, as the api returned a 500: 
```jira.exceptions.JIRAError: JiraError HTTP 500 url: https://rpc-openstack.atlassian.net/rest/api/2/field
	text: <!DOCTYPE html><html lang="en"><head><meta charset="utf-8"><meta http-equiv="X-UA-Compatible" content="IE=edge"><title>Oops - an error has occurred</title>```

recheck_unit_tests

PR #1044, opened 2018-08-31
Cryptography version update

hughsaunders:
This commit bumps cryptography as there are problems with the wheels
for the version we were using previously.

It also adds OS detection to the lint script and sets appropriate
vars so that c libs from homebrew can be located when building
wheels. This shouldn't affect builds in CI (due to the conditional) but
will enable developers to build venvs (and run lint) on local
macOS environments.

Issue: [RE-1923](https://rpc-openstack.atlassian.net/browse/RE-1923)

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -12,7 +12,7 @@ cmd2==0.8.7
 configparser==3.5.0
 contextlib2==0.5.5
 croniter==0.3.20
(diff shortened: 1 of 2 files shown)


hughsaunders:
recheck_unit_tests failed on library git clone in nodepool job

PR #1061, opened 2018-09-04
Added in a gating test of building the release notes

letterj:
The job scenario "build_releasenotes" is run.  This does not install
Ceph or do anything other than try and build the release notes.

### DIFF
--- a/rpc_jobs/rpc_ceph.yml
+++ b/rpc_jobs/rpc_ceph.yml
@@ -207,3 +207,24 @@
     jobs:
       - 'Pull-Request-Whisperer_{repo}'
 
+- project:
+    name: "rpc-ceph-pre-merge-releasenotes"
+
+    repo_name: "rpc-ceph"
+    repo_url: "https://github.com/rcbops/rpc-ceph"
+
+    branches:
+      - "master"
+
+    image:
+      - container:
+          SLAVE_TYPE: "container"
+
+    scenario:
+      - "build_releasenotes"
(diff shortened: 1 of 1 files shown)


rpc_jobs/rpc_ceph.yml
```diff
@@ -207,3 +207,24 @@
     jobs:
       - 'Pull-Request-Whisperer_{repo}'
 
+- project:
```
git-harry: I believe this should already be addressed by "rpc-ceph-re-release-pr" which is defined above. It creates the job https://rpc.jenkins.cit.rackspace.net/job/RE-Release-PR_rpc-ceph-master/ which includes testing the release notes part of the release process. Is that not sufficient?

letterj: @git-harry I did not add this in.   So Does it need to be removed?     If so I can remove it.   Right now I'm trying to figure out why the Jira unit test failed

letterj:
recheck_CIT_unit_tests

letterj:
recheck_unit_tests

rpc_jobs/rpc_ceph.yml
git-harry: Why would you want to remove "rpc-ceph-re-release-pr"? What is the need for this pull request if "rpc-ceph-re-release-pr" already exists?

The unit test failure looked transient to me, it should go away with a recheck.

PR #1065, opened 2018-09-06
RE-1906 Bump flask version

hughsaunders:
The version of flask used in webhooktranslator is vulnerable
to a denial of service attack (CVE-2018-1000656).

This commit bumps flask to the oldest non-vulnerable version.

Issue: [RE-1906](https://rpc-openstack.atlassian.net/browse/RE-1906)


hughsaunders:
It would need to be manually deployed, but as its not in use at the moment, manual decommissioning would be better.

PR #1066, opened 2018-09-06
Implement Standard Job for RE Infrastructure Deployments

dealako:
- Converted Graphite, Grafana, Nodepool, and Statsd separate jobs to a single job-template.
- Deleted separate job definitions

JIRA: RE-1472

Issue: [RE-1472](https://rpc-openstack.atlassian.net/browse/RE-1472)


rpc_jobs/setup_infra.yml
```diff
@@ -0,0 +1,137 @@
...
+
+- project:
+    name: 'Setup-Infra'
+    repo_name: "rpc-gating"
+    repo_url: "https://github.com/rcbops/rpc-gating"
+    branch: "master"
+    jira_project_key: "RE"
+    CRON: "@daily"
+    RPC_GATING_BRANCH: "master"
+    apps:
```
hughsaunders: I would prefer app, as we've used the singular for the other JJB axes.

rpc_jobs/setup_infra.yml
```diff
@@ -0,0 +1,137 @@
...
+      - Grafana:
+          name: "Grafana"
```
hughsaunders: I admire your dedication to capitalisation ;-)

rpc_jobs/setup_infra.yml
```diff
@@ -0,0 +1,137 @@
...
+        }}
+      }} // globalWraps
+
+- project:
+    name: 'Setup-Infra'
+    repo_name: "rpc-gating"
+    repo_url: "https://github.com/rcbops/rpc-gating"
+    branch: "master"
+    jira_project_key: "RE"
+    CRON: "@daily"
```
hughsaunders: Could you use `H H * * 2-7` instead of @daily? This will avoid a conflict with #1040, see https://rpc-openstack.atlassian.net/wiki/spaces/RE/pages/469794817/RE+Infrastructure+Maintenance+Window for more info.

hughsaunders: Actually don't worry about changing this, we'll see which merges first then the other can rebase. Then you can use `{CRON_DAILY}` which won't exist until #1040 merges.

rpc_jobs/setup_infra.yml
```diff
@@ -0,0 +1,137 @@
...
+    name: "Setup-{apps}"
+    project-type: pipeline
+    concurrent: false
+    properties:
+      - build-discarder:
+          num-to-keep: 30
+    parameters:
+      - rpc_gating_params
+    triggers:
+      - timed: "{{CRON}}"
```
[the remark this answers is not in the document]
dealako: but this is a template...so `{{` is required???

rpc_jobs/setup_infra.yml
```diff
@@ -0,0 +1,137 @@
...
+          print(e)
+          throw(e)
+        }} finally {{
+          common.archive_artifacts()
+        }}
+      }} // globalWraps
+
+- project:
+    name: 'Setup-Infra'
+    repo_name: "rpc-gating"
```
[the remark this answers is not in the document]
dealako: removed extra parameters

rpc_jobs/setup_infra.yml
```diff
@@ -0,0 +1,133 @@
...
+        }}
+      }} // globalWraps
+
+- project:
+    name: 'Setup-Infra'
+    CRON: "@daily"
+    RPC_GATING_BRANCH: "master"
+    app:
+      - Grafana:
+          name: "Grafana"
```
git-harry: I don't believe `name: "Grafana"` is necessary, in the template you should just need to use `{app}` instead of `{name}`.

Groovy strings have `toLowerCase`. I think it would be preferable to calculate `name_lower` that have a to define a separate key value pair.

I think this could just be written as:

```yaml
app:
  - Grafana:
      credentials: "id_rsa_cloud10_jenkins_file,RE_GRAFANA_ADMIN_PASSWORD,RE_GRAFANA_GRAFYAML_API_KEY"
```

dealako: Indeed - updated to use {app} instead of {name} which removes the name parameter.  Nice cleanup.

Also, sadly the templating engine doesn't appear to allow functions (unlike helm where you could easily do `{{ mystring | lower }}`. If I'm wrong, please show me the docs.  Yes, we could use toLowerCase in the groovy DSL block on the strings, but I'd have to introduce string concatenation where ${{env.WORKSPACE}} was used - if only to protect against lowering the workspace directory name with mixed case. I see what you are suggesting, but it's too much cleverness which adds complexity to reduce the obvious parameter. In the present form, it's clear and concise.

git-harry: Why not do something like `name_lower = "{app}".toLowerCase()` and then `foo = "bar_${{name_lower}}_baz"` wherever you want to use it.

hughsaunders: JJB doesn't use jinja, it does its own macro interpolation 😠, so you can't do translation within jjb. However name_lower is only used within the dsl block, so as @git-harry mentioned you can use groovy string manipulation. (I should have thought of that).

hughsaunders:
Failed on lint due to maintenance window conflict, see https://rpc-openstack.atlassian.net/wiki/spaces/RE/pages/469794817/RE+Infrastructure+Maintenance+Window but the short answer is use `{CRON_DAILY}` instead of `@daily`

hughsaunders:
recheck_lint

hughsaunders:
recheck_xenial_lint_test

PR #1067, opened 2018-09-07
RE-1842 Hold nodepool nodes on failure

hughsaunders:
This commit holds nodepool nodes when an exception is
caught so that failures can be investigated before
the node is cleaned up.

Issue: [RE-1842](https://rpc-openstack.atlassian.net/browse/RE-1842)


hughsaunders:
This is not useful until the nodepool agents plugin support hold-until, which is being added in [RE-1533|https://rpc-openstack.atlassian.net/browse/RE-1533]. 

Once that has merged, https://github.com/jenkinsci/nodepool-agents-plugin/pull/23 will need to be updated to support timeout, then this can be updated and merged.

hughsaunders:
Should not be merged until https://github.com/jenkinsci/nodepool-agents-plugin/pull/23 has merged, released and been deployed.

pipeline_steps/common.groovy
```diff
@@ -968,7 +968,12 @@ void use_node(String label=null, body){
       print "${env.NODE_NAME} preparation complete, now ready for use."
       body()
     } catch (e){
-      print "Caught exception on ${env.NODE_NAME}: ${e}"
+      errString = "Caught exception on ${env.NODE_NAME}: ${e} Build: ${env.BUILD_URL}"
+      print errString
+
+      if (env.NODE_NAME =~ /^nodepool-/ && env.HOLD_ON_ERROR.asBoolean() == true){
```
git-harry: Given that `HOLD_ON_ERROR` will be a string, I think you want to use [toBoolean](http://docs.groovy-lang.org/latest/html/api/org/codehaus/groovy/runtime/StringGroovyMethods.html#toBoolean-java.lang.String-) instead of [asBoolean](http://docs.groovy-lang.org/latest/html/api/org/codehaus/groovy/runtime/StringGroovyMethods.html#asBoolean-java.lang.CharSequence-).

I wonder if we could be doing a better job with this type of data. Some variables in env are only required by pipeline_steps. Maybe we should defined global variables for these things at the beginning and delete them from `env` so they can be appropriately typed and used without requiring special understanding.

rpc_jobs/re_release_manual.yml
```diff
@@ -20,6 +20,8 @@
           SLAVE_CONTAINER_DOCKERFILE_REPO: "RE"
           SLAVE_CONTAINER_DOCKERFILE_PATH: "./Dockerfile.standard_job"
           SLAVE_CONTAINER_DOCKERFILE_BUILD_ARGS: "BASE_IMAGE=ubuntu:16.04"
+          hold_on_error: false
+          hold_on_error_duration: "nohold"
```
git-harry: If `hold_on_error` is `true` and `hold_on_error_duration` is `"nohold"`, what happens?

Does `"nohold"` have a special significance? I can't find it defined anywhere. Why not just use `0` or maybe `null`?

I'm inclined to suggest that either `hold_on_error_duration` should always have a valid time or `hold_on_error` should be ditched and `hold_on_error_duration` only be used to determine what happens.

pipeline_steps/common.groovy
git-harry: `isNodepoolNode` might be useful to add as a separate function.

rpc_jobs/re_release_manual.yml
hughsaunders: I agree, collapsed into a single variable.

pipeline_steps/common.groovy

hughsaunders: Good catch, removed the boolean var so its no longer an issue. I agree that typing is suboptimal, especially as JJB has a bool type, but stores the result as a string (As env vars are untyped). Worth revisiting this as part of a transition away from jjb.

odyssey4me:
recheck_unit_tests

PR #1069, opened 2018-09-11
Update releases jobs to give clearer information

git-harry:
To improve the clarity of the information provided by whisperer in
rcbops/releases this change:
- updates the status context prefix used by the check job to be prefixed
  with `check` to make it easier to reference by whisperer
- removes the use of CIT from the status context for the release job
- modifies the whisperer comment body to better explain and reference
  the check and release jobs.

JIRA RE-1945

Issue: [RE-1945](https://rpc-openstack.atlassian.net/browse/RE-1945)

### DIFF
--- a/rpc_jobs/releases.yml
+++ b/rpc_jobs/releases.yml
@@ -18,7 +18,7 @@
       - 'test'
 
     jira_project_key: 'RE'
-
+    status_context_prefix: "check"
     jobs:
       - 'PR_{repo_name}-{series}-{image}-{scenario}-{action}'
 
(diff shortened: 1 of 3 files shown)


git-harry:
recheck_all

rpc_jobs/releases.yml
```diff
@@ -32,17 +32,17 @@
...
+            Once all required `check` jobs, prefixed with `check/`, have passed and the pull
+            request has received a sufficient number of reviews, the repository's `release` job
+            can be triggered by adding a comment of `:shipit:` to this pull request.
 
-            If the requested release has one or more failing tests and there are valid reasons
-            to ignore the failure, a release can be forced through by triggering the process
-            a with comment of `:shipit: skip validation`.
+            If the requested release has one or more failing project tests and there are valid
+            reasons to ignore the failure(s), a release can be forced through by triggering the
+            process a with comment of `:shipit: skip validation`.
```
mancdaz: Perhaps it might benefit from the addition of:

'this option should only be used in exceptional circumstances. The first course of action should always be to fix the failing job(s)'

or similar. I just want to make it absolutely clear that it's a last resort rather than a 'oh it's release day and we haven't got time to fix anything, I'll just force it and fix it later (never)'

git-harry: `:shipit: yes I know that it's failing tests but I still really really really want to release it anyway!`

PR #1072, opened 2018-09-17
RE-1954 RE-Enable Phobos nodepool region

hughsaunders:
This commit enabled phobos as a nodepool region for labels
that aren't used by RPC-MaaS jobs. MaaS jobs are incompatible
with phobos as they expected MaaS entities to be pre-created
for the node that they run on.

This commit also adds bionic nodes to phobos and adds ubuntu-bionic
to the disk images config that was missing previously.

Issue: [RE-1954](https://rpc-openstack.atlassian.net/browse/RE-1954)

PR #1075, opened 2018-09-18
Switch to using bionic hosts for RPC-O MNAIO master jobs

odyssey4me:
In order to prepare for using the MNAIO images, we need to
switch to using bionic hosts. They're required to support
the use of the later virt-tools which perform the magic
in the images prior to booting the VM's to ensure that they
work when starting up without having to run any extra
playbooks.

JIRA: RE-1562

Issue: [RE-1562](https://rpc-openstack.atlassian.net/browse/RE-1562)


hughsaunders:
This shouldn't be merged before #1072 as the current nodepool bionic configuration is incomplete.

PR #1077, opened 2018-09-18
Add Queens Check to rpc-maas

dealako:
- Updated job templates to include queens

JIRA: RE-1969

Issue: [RE-1969](https://rpc-openstack.atlassian.net/browse/RE-1969)


chundu:
How about the release job? https://github.com/rcbops/rpc-gating/blob/15616e7892f1a1cae4cf3f1123b47adf26a7cebb/rpc_jobs/rpc_maas.yml#L151-L154

PR #1078, opened 2018-09-19
Add Fallback Region for Snapshot Jobs

dealako:
- Added DFW as the fallback region to allow snapshot jobs to retry within same region

JIRA: RE-1973

Issue: [RE-1973](https://rpc-openstack.atlassian.net/browse/RE-1973)


odyssey4me:
recheck_xenial_lint_test

PR #1079, opened 2018-09-19
RE-1974 Remove Grape Usage

hughsaunders:
Grape is the dependency manager for groovy. Dependency downloads have
been unreliable, this commit removes the one usage of grape.

Replace use of groovy grape with an internal call. Much less code,
though the script approval system gives dire warnings against
this approach:
"1 dangerous signatures previously approved which ought not have been."

Issue: [RE-1974](https://rpc-openstack.atlassian.net/browse/RE-1974)


hughsaunders:
While this commit reduces code, and removes the grape dependency, it requires the `Jenkins.get()` method to be approved for use in pipelines. This allows any job to control the Jenkins instance (shut it down etc). 

screen shot 2018-09-19 at 22 08 24

hughsaunders:
As explained in the updated commit message, we can avoid approving `Jenkins.get()` for all jobs, by disabling the groovy sandbox for the specific job that requires access, which in this case is the RE-Maintenance-Start job.

hughsaunders:
Also removed the cancel quiet down stage & functionality as a Job can't cancel quiet down mode, because no jobs are started while in quietdown mode 🐐

PR #1080, opened 2018-09-20
RE-1976 Fix dual schedules

hughsaunders:
Two timer triggers were added to some jobs in order to avoid
the maintenance window. However only one makes it into the
Jenkins configuration. This commit switches from two timer
triggers to a single timer trigger that contains two cron
expressions.

Issue: [RE-1976](https://rpc-openstack.atlassian.net/browse/RE-1976)

### DIFF
--- a/scripts/extract_dsl.py
+++ b/scripts/extract_dsl.py
@@ -29,7 +29,6 @@ def extract_dsl(jjbfile, outdir):
                 item=item[key]['name'])
             with open(outfile, "w") as outf:
                 outf.write(dsl)
-            print("Written {outfile}".format(outfile=outfile))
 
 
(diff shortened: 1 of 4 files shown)


scripts/extract_dsl.py
```diff
@@ -29,7 +29,6 @@ def extract_dsl(jjbfile, outdir):
                 item=item[key]['name'])
             with open(outfile, "w") as outf:
                 outf.write(dsl)
-            print("Written {outfile}".format(outfile=outfile))
```
odyssey4me: Why was this removed?

hughsaunders: log spam

PR #1081, opened 2018-09-20
Add RPCO snapshot support to Nodepool

git-harry:
nodepool.yml is updated to support three new types of node. Replicating
how the snapshots are currently used, each node type is flavour 7. The
nodes are labelled based on the major series to which they belong and
not the specific version, this simplifies the management of the
configuration but does mean if a project requires use of a specific
version it will need to be done manually.

bump_snapshots.yml has been updated so that Nodepool is automatically
updated when new snapshots become available.

JIRA: RE-1465

Issue: [RE-1465](https://rpc-openstack.atlassian.net/browse/RE-1465)

### DIFF
--- a/nodepool/files/nodepool.yml
+++ b/nodepool/files/nodepool.yml
@@ -24,6 +24,17 @@ labels:
     min-ready: 0
     max-ready-age: 3600
 
+  # rpco snapshots
+  - name: rpc-r14-xenial_loose_artifacts-swift
+    min-ready: 5
+    max-ready-age: 86400
+  - name: rpc-r16-xenial_no_artifacts-swift
+    min-ready: 5
+    max-ready-age: 86400
+  - name: rpc-r17-xenial_no_artifacts-swift
+    min-ready: 5
+    max-ready-age: 86400
+
   #g1-8
   - name: ubuntu-bionic-g1-8
     min-ready: 3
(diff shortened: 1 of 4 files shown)


nodepool/files/nodepool.yml
```diff
@@ -192,6 +203,27 @@ providers:
...
         labels: *provider_pools_onmetal_labels_block
+      - name: snapshots
+        # Ignore the quota given by public cloud and only respect what we provide as max-servers
+        ignore-provider-quota: true
+        max-servers: 10
+        availability-zones: []
+        labels:
+          - name: rpc-r14-xenial_loose_artifacts-swift-7-15
+            cloud-image: rpc-r14.17.0-xenial_loose_artifacts-swift
+            flavor-name: 7
```
odyssey4me: Nodepool (or I think it's shade actually - can't recall) rejects a single letter flavor name, so you'll have to use ``flavor-name: "15GB Standard Instance"`` here.

git-harry:
Tests will need a recheck once Jenkins maintenance is over.

git-harry:
recheck_all

odyssey4me:
recheck_unit_tests
recheck_whisperer

nodepool/files/nodepool.yml
```diff
@@ -24,6 +24,17 @@ labels:
     min-ready: 0
     max-ready-age: 3600
 
+  # rpco snapshots
+  - name: rpc-r14-xenial_loose_artifacts-swift-7-15
```
odyssey4me: nit: These labels are very long, and the ``-7-15`` could be dropped, I think, given that this label can never use another flavor.

nodepool/files/nodepool.yml
```diff
@@ -192,6 +204,27 @@ providers:
...
         availability-zones: []
         labels: *provider_pools_onmetal_labels_block
+      - name: snapshots
+        # Ignore the quota given by public cloud and only respect what we provide as max-servers
+        ignore-provider-quota: true
+        max-servers: 25
+        availability-zones: []
+        labels:
+          - name: rpc-r14-xenial_loose_artifacts-swift
+            cloud-image: rpc-r14.17.0-xenial_loose_artifacts-swift
```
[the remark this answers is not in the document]
odyssey4me: Oh yes. This provider will need to have its own cloud images block - it can't use the anchor. The block for this region will have to have a copy of the IAD one, but with the additional images that these labels will use.

nodepool/files/nodepool.yml
```diff
@@ -174,9 +185,20 @@ providers:
...
+    boot-timeout: 1800
     hostname-format: nodepool-{label.name}-{provider.name}-{node.id}
-    cloud-images: *provider_cloudimage_block
+    cloud-images:
+      - name: ubuntu-bionic-onmetal
+        image-name: "OnMetal - Ubuntu 18.04 LTS (Bionic Beaver)"
+      - name: ubuntu-xenial-onmetal
+        image-name: "OnMetal - Ubuntu 16.04 LTS (Xenial Xerus)"
+      - name: rpc-r14-xenial_loose_artifacts-swift-image
+        image-name: rpc-r14.17.0-xenial_loose_artifacts-swift
```
odyssey4me: I have no idea whether ``config-drive: true`` needs to be set for each of these new images. We do currently make use of config drive when creating them for jenkins. Perhaps it's safer to set it for now?

PR #1083, opened 2018-09-21
Switch rpc-appformix repo URL to use git@

odyssey4me:
The repo is private, so to allow jenkins to authenticate via
its key we need to use the URL git@github.com:rcbops/rpc-appformix
instead.

Issue: [RE-1985](https://rpc-openstack.atlassian.net/browse/RE-1985)


odyssey4me:
recheck_unit_tests

PR #1084, opened 2018-09-21
Allow Single Region to be in both Primary and Fallback

dealako:
- Updated cloud region logic: if there is only one item in the region list, use it - even if it's in the first list, otherwise remove previously attempted region

JIRA: RE-1987

Issue: [RE-1987](https://rpc-openstack.atlassian.net/browse/RE-1987)


dealako:
recheck_unit_tests

PR #1086, opened 2018-09-21
Reset python in rpc-gating venv at run time

odyssey4me:
After extracting the venv on a target host, the python version
on the host may be different enough to cause the scripts used
in the venv to fail. This patch ensures that the python version
in the venv is replaced with whatever is on the host at runtime.

Resetting python in the venv cannot be done by simply doing a
python binary copy - it results in a broken venv which is missing
some of the standard libraries. Due to that, we need to use the
'virtualenv' tool for the reset. Given the range of options
available over time, we have to do a fair amount of logic to
figure out the options available to use.

To help prove that this works, unit tests are added for the
following:

1. Get a nodepool bionic node.
2. Publish artifacts using a xenial container (instead of using
   the long-lived slave), and a xenial nodepool node.
3. Publish artifacts using a bionic container, and a bionic
   nodepool node.
4. Verify that at least one script is working using the venv when
   building any single-use slaves (container, nodepool, and
   jenkins-built).

The current standard job Dockerfile installs a lot of content into
it, so the environment is not pristine. The package 'groovy2' is
also not available in Ubuntu Bionic. As part of this patch, we
reduce what the Dockerfile installs to a minimal set to support
running tests and executing the rpc-gating artifact uploading.

Note that the use of the rpc-gating scripts in a container is not
guaranteed to work. The artifact uploads work, but the jira-utils
script does not. The issue is that the venv python replacement is
done on the long-running-slave (Xenial), but then trying to use it
in the Bionic (or other) platform in the container fails because
the standard libraries aren't right for that distro. As such, I've
switched the Container Custom Args test to just use Xenial.

Issue: [RE-1977](https://rpc-openstack.atlassian.net/browse/RE-1977)

### DIFF
--- a/rpc_jobs/unit/artefact_publish.yml
+++ b/rpc_jobs/unit/artefact_publish.yml
@@ -7,11 +7,22 @@
           num-to-keep: 30
     parameters:
       - rpc_gating_params
+      - standard_job_params:
+          SLAVE_TYPE: "container"
+          SLAVE_CONTAINER_DOCKERFILE_REPO: "RE"
+          SLAVE_CONTAINER_DOCKERFILE_PATH: "./Dockerfile.standard_job"
+          SLAVE_CONTAINER_DOCKERFILE_BUILD_ARGS: "BASE_IMAGE=ubuntu:16.04"
+          hold_on_error: "0"
     dsl: |
       library "rpc-gating@${RPC_GATING_BRANCH}"
       common.globalWraps(){
-        // Do something that creates an artifact
-        stage("Build"){
+        // NOTE(mattt): These vars are required in common.standard_job_slave()
+        // when this job is testing a custom dockerfile.
+        env.REPO_URL = "https://github.com/rcbops/rpc-gating"
+        env.BRANCH = "${RPC_GATING_BRANCH}"
+        env.RE_JOB_REPO_NAME = "rpc-gating"
+
+        common.standard_job_slave(env.SLAVE_TYPE){
           // ARA placeholder included to test that an ARA
           // link is added to the sidebar of the index page
           // when an ARA report is present.
@@ -30,9 +41,6 @@
             cp -r artifacts file_artifacts_no_dest
           """
 
-          // Set the RE_JOB_REPO_NAME (normally done by standard jobs)
-          env.RE_JOB_REPO_NAME = "rpc-gating"
-
           // Create a temporary component_metadata file
           def artifacts_data = [
             artifacts: [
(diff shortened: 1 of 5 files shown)


odyssey4me:
recheck_unit_tests

odyssey4me:
Verifying whether this resolves the artifact upload issues in https://rpc.jenkins.cit.rackspace.net/view/RPC-O/job/PM_rpc-openstack-master-xenial_mnaio_no_artifacts-swift-deploy/322/

odyssey4me:
recheck_unit_tests

odyssey4me:
recheck_unit_tests

odyssey4me:
recheck_unit_tests

odyssey4me:
recheck_unit_tests

odyssey4me:
recheck_unit_tests

odyssey4me:
recheck_unit_tests

odyssey4me:
recheck_xenial_lint_test

odyssey4me:
recheck_unit_tests

odyssey4me:
recheck_unit_tests

odyssey4me:
recheck_unit_tests

odyssey4me:
recheck_unit_tests

odyssey4me:
recheck_unit_tests

odyssey4me:
recheck_unit_tests

odyssey4me:
Tested with the jobs built using the changes from this PR: https://rpc.jenkins.cit.rackspace.net/job/RPC-Gating-Unit-Tests/1948/

rpc_jobs/unit/artefact_publish.yml
```diff
@@ -6,12 +6,28 @@
       - build-discarder:
           num-to-keep: 30
     parameters:
+      - instance_params:
+          IMAGE: "Ubuntu 16.04 LTS (Xenial Xerus) (PVHVM)"
```
hughsaunders: Where are these used? `common.globalWraps()` will allocate a long running slave, `common.standard_job_slave()` will create a container on that long running slave.

odyssey4me: I just grabbed them from the single_use_slave unit test because I thought they were necessary.

odyssey4me:
recheck_unit_tests

odyssey4me:
recheck_unit_tests

unrelated failure:

```
TASK [Start vpnc service] ******************************************************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "Unable to start service vpnc.service: Job for vpnc.service failed because the control process exited with error code. See \"systemctl status vpnc.service\" and \"journalctl -xe\" for details.\n"}
	to retry, use: --limit @/var/lib/jenkins/workspace/RE-unit-test-wrappers/rpc-gating/playbooks/****_vpn.retry

localhost                  : ok=4    changed=3    unreachable=0    failed=1  
```

odyssey4me:
🔥

PR #1090, opened 2018-09-25
RE-1965 Maintenance Process Improvements

hughsaunders:
The first RE maintenance window occurred on 2018-09-24, and there were some
    teething problems:
    * Builds aborted due to the maintenance window had result set to
      SUCCESS instead of ABORTED.
    * Schedule was incorrect for the maintenance start job
    * Maintenance start job was aborted as it overlapped the maintenance
      window
    * Maintenance start job failed on missing common. prefix for lib func.
    * Maintenance was completed quickly but there was no mechanism for
      resuming jobs before the end of the predefined window.

    All these issues are fixed by this commit.

 e5adbe0 (Hugh Saunders, 23 hours ago)
    Revert "RE-1965 Disable Maintenance aborts"

    This reverts commit 0b3748ac576a5a89f3f9fc17d17457192ef33cde. Maintenance
    aborts were disabled temporarily while bugs are fixed This pr fixes the
    bugs and re-enables the abort check.

Issue: [RE-1965](https://rpc-openstack.atlassian.net/browse/RE-1965)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -3,12 +3,21 @@ import groovy.json.JsonOutput
 import groovy.json.JsonException
 import org.jenkinsci.plugins.workflow.job.WorkflowJob
 import org.jenkinsci.plugins.workflow.job.WorkflowRun
+import java.time.LocalTime
 import java.time.LocalDateTime
 import java.time.DayOfWeek
 import java.time.Duration
 import java.time.ZoneOffset
 import java.time.format.DateTimeFormatter
 import java.math.MathContext
+import groovy.transform.Field
+
+// Constants Governing the Weekly RE-Maintenance Window
+@Field DayOfWeek maintDay = DayOfWeek.MONDAY
+@Field Integer maintHour = 10
+@Field Duration maintDuration = Duration.ofHours(2)
+@Field Integer daysInWeek = 7
+
 
 void download_venv(){
   sh """#!/bin/bash -xeu
@@ -1397,6 +1406,7 @@ void wrapList(List wrappers, Closure body){
   }
 }
 
+
 // add wrappers that should be used for all jobs.
 // max log size is in MB
 void globalWraps(Closure body){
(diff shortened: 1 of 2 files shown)


pipeline_steps/common.groovy
```diff
@@ -2131,14 +2135,84 @@ Boolean willOverlapMaintenanceWindow(){
...
+    /RPC-Gating-Unit-Tests/,
+    /scatchpipeline/
+  ]
+
+  for (pattern in patterns){
+    if (env.JOB_NAME.contains(pattern)){
+      return true
+    }
+  }
+  print("This Job is not allowed to build during maintenance windows")
```
git-harry: You've added print statements to a number of these `Boolean` type functions for one value, why not add logging for the inverse result? Do the logs make it obvious so it's not required?

hughsaunders: The prints all show things that contribute to a build being aborted. A test that allows the build to continue doesn't need to log IMO.

pipeline_steps/common.groovy
```diff
@@ -2131,14 +2135,84 @@ Boolean willOverlapMaintenanceWindow(){
...
+  return false
+}
+
+Boolean maintenanceComplete(){
+
+  // 1. Maintenance is complete if the window has expired
+  LocalDateTime now = LocalDateTime.now()
+  if(now.getHour() > maintHour + maintDuration){
```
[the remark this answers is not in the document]
hughsaunders: Refactored so that math is only done on datetime objects rather than ints.

mancdaz:
recheck_unit_tests

odyssey4me:
recheck_unit_tests I suspect the unit tests may fail until https://github.com/rcbops/rpc-gating/pull/1093 is finalised.

pipeline_steps/common.groovy
```diff
@@ -2131,14 +2127,75 @@ Boolean willOverlapMaintenanceWindow(){
...
+  String jiraIssue = ""
+  try {
+    jiraIssue = maintenanceIssueForDate("today")
+  } catch (Exception e){
+    // No maintenance issue for today so no maintenance is in progress.
+    return false
+  }
+
+  Boolean afterStartTime =  LocalTime.now() > LocalTime.of(maintHour, 0)
+  query = "key=\"${jiraIssue}\" and STATUS in (Doing)"
```
git-harry: I think using doing column leaves this open to mistakes, there are potentially reasons that a maintenance card could go into one of the columns between doing and done without the maintenance being over. Using the done column is unambiguous.

What happens if the maintenance window has started but the issue is still in the backlog? Is it intentional that build should be able to run during this time or is there some other mechanism that prevents them from starting?

mancdaz: > there are potentially reasons that a maintenance card could go into one of the columns between doing and done without the maintenance being over. Using the done column is unambiguous

You could use ```statusCategory``` rather than a hardcoded ```status```. 
```'statusCategory = "In Progress"'```would  cover any status in between ```Backlog``` and ```Done```.

mancdaz:
recheck_unit_tests

PR #1093, opened 2018-09-26
Phobos VPN: Add debugging and conditional connecting

hughsaunders:
Phobos VPN
* Add debug output to playbook
* VPN into Phobos only if outside Phobos
* Successfully test Phobos connectivity as exit criteria

Issue: [RE-2003](https://rpc-openstack.atlassian.net/browse/RE-2003)

### DIFF
--- a/playbooks/phobos_vpn.yml
+++ b/playbooks/phobos_vpn.yml
@@ -27,9 +27,36 @@
         src: vpnc.service.j2
         dest: /lib/systemd/system/vpnc.service
 
-    - name: Start vpnc service
-      systemd:
-        enabled: yes
-        service: vpnc.service
-        daemon_reload: yes
-        state: restarted
+    - name: Start VPN
+      block:
+        - name: Start vpnc service
+          systemd:
+            enabled: yes
+            service: vpnc.service
+            daemon_reload: yes
+            state: restarted
+          register: vpnc_restart
+
+      rescue:
+        - name: Output failed task information
+          debug: var=vpnc_restart
+
+        - name: Get vpnc service status
+          command: systemctl status vpnc.service
+          register: vpnc_status
+          failed_when: false
+          changed_when: false
+
+        - name: Output vpnc service status
+          debug: var=vpnc_status
+
+        - name: Get system log
+          command: journalctl -xe
+          register: system_log
+          changed_when: false
+
+        - name: Output system log
+          debug: var=system_log
+
+        - name: Exit with failure
+          fail: msg="vpnc service failed to start"
--- a/playbooks/phobos_vpn.yml
+++ b/playbooks/phobos_vpn.yml
@@ -1,62 +1,90 @@
 - name: Configure Phobos VPN Tunnel (should only be run on single use slaves)
   hosts: "{{ target_hosts | default('localhost') }}"
   tasks:
-    # Use shell because python-apt isn't available so the apt module fails.
-    # I don't want to add python-apt to one of the more general slave setup
-    # scripts as it depends on python. We want to reduce the number of deps
-    # installed on general slaves to the bare minimum.
-    - name: Install apt packages
+    - name: Install ping
+      # Don't use Ansible's apt module, but instead apt-get "manually", because
+      # its dependencies, ie python-apt and Python, are probably not installed.
       shell: |
         apt-get update
-        apt-get install -y vpnc
-      register: install_packages
-      until: install_packages | success
+        apt-get install -y iputils-ping
+      register: install_ping
+      until: install_ping | success
       retries: 5
       delay: 2
       tags:
         - skip_ansible_lint
 
+    - name: Test connectivity to VPN server
+      # NOTE(coreywright): The Phobos VPN server cannot be pinged from within
+      # the Phobos environment, so use this task as a method of determining
+      # whether we are in the Phobos environment or not.
+      command: ping -c 3 {{ gateway }}
+      ignore_errors: true
+      changed_when: false
+      register: ping_vpn
 
-    - name: Template vpnc config
-      template:
-        src: vpnc.conf.j2
-        dest: /etc/vpnc/phobos.conf
-
-    - name: Template vpnc systemd service
-      template:
-        src: vpnc.service.j2
-        dest: /lib/systemd/system/vpnc.service
-
-    - name: Start VPN
+    - name: Configure and start VPN client
+      # NOTE(coreywright): If we are not in the Phobos environment, ie we can
+      # ping the Phobos VPN server, then we need to connect to it by way of
+      # VPN.
+      when: ping_vpn | succeeded
       block:
-        - name: Start vpnc service
-          systemd:
-            enabled: yes
-            service: vpnc.service
-            daemon_reload: yes
-            state: restarted
-          register: vpnc_restart
+        # Use shell because python-apt isn't available so the apt module fails.
+        # I don't want to add python-apt to one of the more general slave setup
+        # scripts as it depends on python. We want to reduce the number of deps
+        # installed on general slaves to the bare minimum.
+        - name: Install apt packages
+          shell: |
+            apt-get update
+            apt-get install -y vpnc
+          register: install_packages
+          until: install_packages | success
+          retries: 5
+          delay: 2
+          tags:
+            - skip_ansible_lint
 
-      rescue:
-        - name: Output failed task information
-          debug: var=vpnc_restart
 
-        - name: Get vpnc service status
-          command: systemctl status vpnc.service
-          register: vpnc_status
-          failed_when: false
-          changed_when: false
+        - name: Template vpnc config
+          template:
+            src: vpnc.conf.j2
+            dest: /etc/vpnc/phobos.conf
 
-        - name: Output vpnc service status
-          debug: var=vpnc_status
+        - name: Template vpnc systemd service
+          template:
+            src: vpnc.service.j2
+            dest: /lib/systemd/system/vpnc.service
 
-        - name: Get system log
-          command: journalctl -xe
-          register: system_log
-          changed_when: false
+        - name: Start VPN
+          block:
+            - name: Start vpnc service
+              systemd:
+                enabled: yes
+                service: vpnc.service
+                daemon_reload: yes
+                state: restarted
+              register: vpnc_restart
 
-        - name: Output system log
-          debug: var=system_log
+          rescue:
+            - name: Output failed task information
+              debug: var=vpnc_restart
 
-        - name: Exit with failure
-          fail: msg="vpnc service failed to start"
+            - name: Get vpnc service status
+              command: systemctl status vpnc.service
+              register: vpnc_status
+              failed_when: false
+              changed_when: false
+
+            - name: Output vpnc service status
+              debug: var=vpnc_status
+
+            - name: Get system log
+              command: journalctl -xe
+              register: system_log
+              changed_when: false
+
+            - name: Output system log
+              debug: var=system_log
+
+            - name: Exit with failure
+              fail: msg="vpnc service failed to start"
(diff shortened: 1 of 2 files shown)


playbooks/phobos_vpn.yml
```diff
@@ -33,3 +33,27 @@
         service: vpnc.service
         daemon_reload: yes
         state: restarted
+      failed_when: false
```
odyssey4me: Instead of using failed_when: false, rather use a try/rescue/always block. You could make the diagnostic output only show in the rescue block, so it'll only show when it fails the try block which is where the service restart should be.

coreywright: yeah, i stumbled upon that yesterday (while reminding myself the syntax of how to put a conditional on a block): https://docs.ansible.com/ansible/latest/user_guide/playbooks_blocks.html#error-handling.  i just didn't have the time at that moment to learning something new and experiment with it (not knowing how long the then-current vpnc failures was going to last).

all that is not to say that's it's not the better approach, and i'll implement it when i have the time (and the opportunity to test it with vpnc failing), but merely to justify the less-then-optimal change in this PR.

coreywright:
pro-tip: ignoring whitespace more easily reveals this PR's changes (and github now has an option to do just that).

i first tried to use ping to test phobos VPN connectivity, but ping wasn't installed, so i installed it... and then figured if i'm going to install something, then i might as well install netcat and use it to test for connectivity to both the VPN server (before) and phobos APIs (after).  i originally wasn't going to use the phobos APIs, but it's a great "exit criteria" proving that the rest of the playbook performed correctly (whether to start the VPN or not).

coreywright:
testing of the current version of this PR started with https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-eng-ops-master-xenial-embedded-deploynoclean/631 (though technically that used commit 2714f187ea98559f2cdfe556e25a116ce553131c, but that only varies from the current version by a comment in the phobos_vpn.yml playbook).

```diff
user@localhost:/rpc-gating$ git diff 2714f187ea98559f2cdfe556e25a116ce553131c f5836afb99d922ef64f96bb4f0bb79658a222318
diff --git a/playbooks/phobos_vpn.yml b/playbooks/phobos_vpn.yml
index 351d863..57b081e 100644
--- a/playbooks/phobos_vpn.yml
+++ b/playbooks/phobos_vpn.yml
@@ -89,7 +89,7 @@
 
     - name: Test connectivity to Phobos API endpoint
       # NOTE(coreywright): Whether we are in the Phobos environment or not (and
-      # correspondingly just made a VPN connect to it) prove Phobos
+      # correspondingly just made a VPN connection to it) prove Phobos
       # connectivity by connecting to its OpenStack API endpoint.
       command: netcat -v -z phobos.rpc.rackspace.com 5000
       changed_when: false
```

playbooks/phobos_vpn.yml
```diff
@@ -1,35 +1,95 @@
...
-    # scripts as it depends on python. We want to reduce the number of deps
-    # installed on general slaves to the bare minimum.
-    - name: Install apt packages
+    - name: Install netcat
       shell: |
         apt-get update
-        apt-get install -y vpnc
-      register: install_packages
-      until: install_packages | success
+        apt-get install -y netcat-openbsd
```
odyssey4me: Rather than install netcat, you could just use wait_for to test connectivity. As noted in the comment that's been removed here, we try not to install things on target hosts as far as we possibly can.

playbooks/phobos_vpn.yml
```diff
@@ -1,35 +1,95 @@
...
+              register: system_log
+              changed_when: false
+
+            - name: Output system log
+              debug: var=system_log
+
+            - name: Exit with failure
+              fail: msg="vpnc service failed to start"
+
+    - name: Test connectivity to Phobos API endpoint
```
odyssey4me: wait_for can also be used here rather than netcat

playbooks/phobos_vpn.yml
coreywright: 1. the comment wasn't removed, but instead moved along with the associated task (and placed within a `block:`, ie indented, which misled the diff algorithm, which is why i recommended diff-ing with the "hide whitespace changes" option, whether here in github or in git with `-w`).
1. i interpreted the comment to refer to packages installed on the slave *before* a job (as part of slave setup), not *during* a job, especially if easily worked around, not needed for all jobs, and increases "base" slave resources (eg setup time, image size if making snapshots or creating custom images).
1. `wait_for` doesn't appear to support udp which is what vpnc uses (ie udp port 4500).

all that said, it was my mistake to use netcat for udp port detection (`UDP port scans using the -uz combination of flags will always report success irrespective of the target machine's state.` - netcat man page; i thought its algorithm was better, more like nmap's where it factors in ICMP responses, eg "destination port unreachable" if the port is closed, but firewalls and IP stack configurations make this unreliable), so i'm going back to:

1. install ping
1. use ping to detect VPN server reachability

playbooks/phobos_vpn.yml
coreywright: yeah, i'll use `wait_for` here.  netcat isn't working in the UDP-for-VPN case, so i'm reverting to ping there and `wait_for` suffices for TCP-for-API here.

thanks for the `wait_for` reminder (as i've primarily only used it to detect if a host is up and ready to continue with Ansible after a reboot by `wait_for` TCP port 22, ie SSH, which is what Ansible specifically needs to "wait for" before continuing).

odyssey4me: Assuming this endpoint is a web service, you can also use the ``uri`` module instead, which allows you to just get the header and you can tell it what HTTP response to expect.

playbooks/phobos_vpn.yml
odyssey4me: Ah, fair enough. Thanks for covering the bases - I think using ping might be the simplest option here then.

coreywright:
tested all day in https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-eng-ops-master-xenial-embedded-deploynoclean (while testing rpc-eng-ops embed-env changes and trying to recreate an intermittent failure), but only ever ran on public cloud slaves, never ran on a phobos slave (ie vpn-not-needed conditions never tested).

coreywright:
tl;dr playbook seemed to succeed for the slave-in-phobos use-case, but overall job failed, so still investigating and testing.

finally presumably got a slave in phobos (ie pinging vpn server failed and was able to connect to phobos API endpoint though skipping starting vpn), but strange failure later on:
```
13:55:58 TASK [Install Rally system requirements] ***************************************
14:00:09  [WARNING]: sftp transfer mechanism failed on [172.20.41.86]. Use
14:00:09 ANSIBLE_DEBUG=1 to see detailed information
14:15:42  [WARNING]: scp transfer mechanism failed on [172.20.41.86]. Use
14:15:42 ANSIBLE_DEBUG=1 to see detailed information
14:19:37 fatal: [deploy_vm]: UNREACHABLE! => {"changed": false, "msg": "SSH Error: data could not be sent to remote host \"172.20.41.86\". Make sure this host can be reached over ssh", "unreachable": true}
14:19:40 	to retry, use: --limit @/var/lib/jenkins/workspace/PM_rpc-eng-ops-master-xenial-embedded-deploynoclean/rpc-eng-ops/embedded-environment/rally.retry
14:19:40 
14:19:40 PLAY RECAP *********************************************************************
14:19:40 
deploy_vm                  : ok=15   changed=0    unreachable=1    failed=0   
```

this is strange because the phobos API was exercised successfully many times, so the slave is truly in phobos, but it hiccuped once which i haven't seen (or maybe i just haven't noticed, especially since it retries):
```
13:53:26 + openstack --insecure configuration show
13:55:35 Failed to discover available identity versions when contacting https://phobos.rpc.rackspace.com:5000/. Attempting to parse version from URL.
13:55:35 Could not find versioned identity endpoints when attempting to authenticate. Please check that your auth_url is correct. SSL exception connecting to https://phobos.rpc.rackspace.com:5000/: HTTPSConnectionPool(host='phobos.rpc.rackspace.com', port=5000): Max retries exceeded with url: / (Caused by SSLError(SSLError("bad handshake: SysCallError(110, 'ETIMEDOUT')",),))
[Pipeline] }
13:55:36 ERROR: script returned exit code 1
13:55:36 Retrying
```

coreywright:
Ansible playbook and initial Phobos API connectivity failures because of [REO-365](https://rpc-openstack.atlassian.net/browse/REO-365) and work-around commit added to PR.

coreywright:
Converted `wait_for` usage to `uri` because HTTP(S) request, as compared to simple-stupid TCP data-less connection, seems to test REO-365 work-around in addition to general Phobos connectivity (two birds, one stone).

playbooks/phobos_vpn.yml
```diff
@@ -1,35 +1,111 @@
...
+              changed_when: false
+
+            - name: Output system log
+              debug: var=system_log
+
+            - name: Exit with failure
+              fail: msg="vpnc service failed to start"
+
+    - name: Set MTU to 1450
+      # NOTE(coreywright): If we are in the Phobos environment, ie we can ping
```
hughsaunders: Should be `can't` ?

playbooks/phobos_vpn.yml
```diff
@@ -1,35 +1,111 @@
...
+              register: system_log
+              changed_when: false
+
+            - name: Output system log
+              debug: var=system_log
+
+            - name: Exit with failure
+              fail: msg="vpnc service failed to start"
+
+    - name: Set MTU to 1450
```
hughsaunders: If I understand correctly, all phobos nodes need the MTU tweak, but not all jobs that may run in phobos include this playbook. 

So this should probably be in a separate playbook or a common.groovy function. 

eg:
```
void phobosSetMTU(){
  sh """
	which ping || { apt-get update; apt-get install -y iputils-ping}
	ping -c 3 ${gateway} || ip link set mtu 1450 dev ens3
  """
```
Then add a call to that after `install_ansible()` in [common.use_node](https://github.com/rcbops/rpc-gating/blob/master/pipeline_steps/common.groovy#L1008)

Is `ens3` the correct interface name across virtualised and bare metal instances?

coreywright: tl;dr this task is/was just a *temporary* work-around so i can/could continue testing and using this playbook until REO-365 was resolved.  sorry for not using the word "temporary", "short-term", etc and/or adding the "status-dont-merge" tag.

REO-365 has been resolved.  i was 99% sure of the resolution and the speed of its implementation, but deferred on doing it myself, so i implemented this work-around (and took great advantage of it in the intervening time).

coreywright:
Removed commit/work-around related to REO-365 now that it has been resolved, but left in the `wait_for` → `uri` change because:
1. it would have helped catch/expose REO-365 (and did help test it) and can help catch future problems.
1. it's no more onerous than `wait_for`.

btw, thanks @odyssey4me for the original `uri` suggestion!

odyssey4me:
recheck_unit_tests

odyssey4me:
recheck_unit_tests

playbooks/phobos_vpn.yml
```diff
@@ -1,35 +1,101 @@
...
+      # VPN.
+      when: ping_vpn | succeeded
+      block:
+        # Use shell because python-apt isn't available so the apt module fails.
+        # I don't want to add python-apt to one of the more general slave setup
+        # scripts as it depends on python. We want to reduce the number of deps
+        # installed on general slaves to the bare minimum.
+        - name: Install apt packages
+          shell: |
+            apt-get update
```
odyssey4me: apt-get update here isn't necessary, given that it was done in the "Install ping" task

coreywright: yeah, sorry, stupid me: mindless cut-paste-minimal-edit error.

PR #1095, opened 2018-09-26
Fix thaw file detection

git-harry:
When using the Groovy File class, the thaw file does not get found. This
can switches to using Bash.

JIRA: RE-1465

Issue: [RE-1465](https://rpc-openstack.atlassian.net/browse/RE-1465)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -2306,14 +2306,14 @@ void restartAbortedPRBuilds(String maint_date="today"){
  */
 void runThawIfSnapshot(){
   String thawFileName = "/gating/thaw/run"
-  def thawFile = new File(thawFileName)
-  if (thawFile.exists()){
-    stage('Execute snapshot thaw'){
-      sh """#!/bin/bash -xeu
+  stage("Execute snapshot thaw if ${thawFileName} present"){
+    sh """#!/bin/bash -eu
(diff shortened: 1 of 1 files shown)


pipeline_steps/common.groovy
```diff
@@ -2306,14 +2306,14 @@ void restartAbortedPRBuilds(String maint_date="today"){
  */
 void runThawIfSnapshot(){
   String thawFileName = "/gating/thaw/run"
-  def thawFile = new File(thawFileName)
-  if (thawFile.exists()){
-    stage('Execute snapshot thaw'){
-      sh """#!/bin/bash -xeu
+  stage('Execute snapshot thaw if file present'){
```
odyssey4me: To be clearer, perhaps replace 'file' with '/gating/thaw/run' here?

pipeline_steps/common.groovy
```diff
@@ -2306,14 +2306,14 @@ void restartAbortedPRBuilds(String maint_date="today"){
...
+    sh """#!/bin/bash -xeu
```
odyssey4me: This is a simple enough script that I think we can forgo setting -x here to cut down on some log noise.

git-harry:
@odyssey4me I had to update because I forgot to switch to `"` from `'` so interpolation wasn't working.

PR #1097, opened 2018-09-26
Expose the build result to the post stage

odyssey4me:
To ensure that standard jobs are able to recognise whether the run
stage passed/failed, we export a new environment variable before
running the post script. This makes it possible to adjust the post
script to skip the artifact configuration which causes broken images
to get uploaded after the job.

Issue: [RE-1989](https://rpc-openstack.atlassian.net/browse/RE-1989)


odyssey4me:
recheck_unit_tests

PR #1098, opened 2018-09-26
RI-489 Adds Periodic Testing of osp-mnaio

antonym:
Adds Periodic Testing of osp-mnaio

Issue: [RI-489](https://rpc-openstack.atlassian.net/browse/RI-489)

### DIFF
--- a/rpc_jobs/osp_mnaio.yml
+++ b/rpc_jobs/osp_mnaio.yml
@@ -41,6 +41,40 @@
       - 'PR_{repo_name}-{series}-{image}-{scenario}-{action}'
 
 - project:
+    name: "osp-mnaio-periodic-daily"
+    repo_name: "osp-mnaio"
+    repo_url: "https://github.com/rcbops/osp-mnaio"
+    image:
+      - xenial_mnaio:
+          SLAVE_TYPE: "nodepool-ubuntu-xenial-om-io2"
+    scenario:
+      - "3ctlr_2comp_3ceph"
+    action:
+      - "osp_13_deploy-smoke-tests"
+    jira_project_key: "RLM"
+    credentials: "rpc_osp"
+    jobs:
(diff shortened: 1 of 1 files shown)


rpc_jobs/osp_mnaio.yml
```diff
@@ -40,6 +40,23 @@
...
+    name: "osp-mnaio-periodic"
+    repo_name: "osp-mnaio"
+    repo_url: "https://github.com/rcbops/osp-mnaio"
+    image:
+      - xenial_mnaio:
+          SLAVE_TYPE: "nodepool-ubuntu-xenial-om-io2"
+    scenario:
+      - "3ctlr_2comp_3ceph"
+    action:
+      - "osp_13_deploy"
```
coreywright: can we change this to `osp_13_test`, `osp_13_deploy-and-test`, etc, so we can pivot off of the action name (eg `RE_JOB_ACTION`, `OSP_JOB_ACTION`) within the gating scripts (ie `osp-mnaio/gating/periodic/`) to determine that we should modify the tests ran (ie `TEMPEST_TESTS`).

i'm envisioning doing a full test run weekly and a smoke test run nightly with that communicated to osp-mnaio's gating scripts by way of the action (eg `osp_13_test-full` & `osp_13_test-smoke`, respectively).

what's currently proposed has no difference from the PR gating job that i can tell except for the full name (ie `RE_JOB_NAME`; `PM_{repo_name}-{branch}-{image}-{scenario}-{action}` vs `PR_{repo_name}-{series}-{image}-{scenario}-{action}`) which is not exactly explicit nor intuitive about how testing differs or that the testing even differs at all.

antonym:
@coreywright pushed up an update to run full weekly and smoke daily.  I figure we probably don't just need just a no test deploy and can lump it in with smoke.

PR #1100, opened 2018-09-26
Reduce max-age for nodepool nodes to reduce churn

odyssey4me:
Our current max-age for nodepool nodes for most labels
is 1 hour. This was initially implemented during the
testing phase for nodepool to ensure that changes in
the image were implemented quickly. However, the images
are now largely stable, so all it's doing now is causing
unnecessary churn.

This patch changes the max-age to 1 day to reduce churn.

The snapshot images are not yet being used by any jobs,
but consuming a large amount of quota. Even when they
do start getting used, the frequency of their usage is
quite low, so a min-ready setting of 1 should be fine.

To ensure that our VM usage for DFW does not exceed our
actual quota of 512GB RAM, we also change the max-servers
for the two pools to ensure that neither can ever use more
than half of it. We set ORD to max-server 30 to match the
quota, and reduce IAD to 20 to leave headroom as that's
the region where we run most of our infrastructure.

Issue: [RE-2006](https://rpc-openstack.atlassian.net/browse/RE-2006)

### DIFF
--- a/nodepool/files/nodepool.yml
+++ b/nodepool/files/nodepool.yml
@@ -16,62 +16,62 @@ labels:
   # maas specific labels are introduced that draw from all regions except phobos.
   - name: maas-ubuntu-xenial-g1-8
     min-ready: 0
-    max-ready-age: 3600
+    max-ready-age: 86400
   - name: maas-ubuntu-trusty-g1-8
     min-ready: 0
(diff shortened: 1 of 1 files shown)


odyssey4me:
recheck_unit_tests

PR #1107, opened 2018-10-01
RI-499 Use snapshots on incremental upgrades

antonym:
Use version snapshots to avoid inital deploy and
just focus on upgrading.

Issue: [RI-499](https://rpc-openstack.atlassian.net/browse/RI-499)

### DIFF
--- a/rpc_jobs/rpc_upgrades.yml
+++ b/rpc_jobs/rpc_upgrades.yml
@@ -249,6 +249,24 @@
     CRON: "{CRON_WEEKLY}"
 
 - project:
+    name: "rpc-upgrades-aio-snapshot-incremental"
+    repo_name: "rpc-upgrades"
+    repo_url: "https://github.com/rcbops/rpc-upgrades"
+    image:
+      - xenial_snapshot
+    scenario:
+      - swift
+    action:
+      - newton_to_queens_incremental:
+          SLAVE_TYPE: "rpc-r14-xenial_loose_artifacts-swift"
+      - pike_to_queens_incremental:
+          SLAVE_TYPE: "rpc-r16-xenial_no_artifacts-swift"
+    jira_project_key: "RLM"
+    jobs:
+      - 'PM_{repo_name}-{branch}-{image}-{scenario}-{action}'
+    CRON: "{CRON_WEEKLY}"
+
+- project:
     name: "rpc-upgrades"
     repo_name: "rpc-upgrades"
     repo_url: "https://github.com/rcbops/rpc-upgrades"


rpc_jobs/rpc_upgrades.yml
```diff
@@ -248,6 +248,26 @@
...
+      - xenial_snapshot:
+          REGIONS: "DFW"
+          BOOT_TIMEOUT: 1500
+    scenario:
+      - newton:
+          IMAGE: "rpc-r14-xenial_loose_artifacts-swift"
+      - pike:
+          IMAGE: "rpc-r16-xenial_no_artifacts-swift"
+    action:
+      - "newton_to_queens_incremental"
```
odyssey4me: The resulting job name doesn't make much sense. Are you sure this is what you want? It seems to me that the scenario/action axes setup you have for the mnaio job above makes a bit more sense?

rpc_jobs/rpc_upgrades.yml
```diff
@@ -248,6 +248,26 @@
...
+    name: "rpc-upgrades-aio-newton-snapshot-incremental"
+    repo_name: "rpc-upgrades"
+    repo_url: "https://github.com/rcbops/rpc-upgrades"
+    image:
+      - xenial_snapshot:
+          REGIONS: "DFW"
+          BOOT_TIMEOUT: 1500
+    scenario:
+      - newton:
+          IMAGE: "rpc-r14-xenial_loose_artifacts-swift"
```
odyssey4me: `rpc-r14-xenial_loose_artifacts-swift` is not an available IMAGE, however it is an available SLAVE_TYPE. Using this will make use of nodepool, which is good - just ensure that the right key word is used.

rpc_jobs/rpc_upgrades.yml
```diff
@@ -248,6 +248,26 @@
...
+      - pike:
+          IMAGE: "rpc-r16-xenial_no_artifacts-swift"
```
odyssey4me: `rpc-r16-xenial_no_artifacts-swift` is not an available IMAGE, however it is an available SLAVE_TYPE. Using this will make use of nodepool, which is good - just ensure that the right key word is used.

rpc_jobs/rpc_upgrades.yml
antonym: Yeah, the scenario is a bit redundant with the action, but changing the action would break how we parse it currently.  I'm mainly doing this as a test to see if it works and if we benefit from doing snapshots, I'll probably refactor how the scenarios/actions work.

rpc_jobs/rpc_upgrades.yml
antonym: Yep, makes sense, adjusted to use SLAVE_TYPE for both.

rpc_jobs/rpc_upgrades.yml
antonym: Yeah, that makes more sense, I forgot you can move those variables anywhere, updated.

PR #1109, opened 2018-10-02
Publishing file artefacts

git-harry:
JIRA: RE-1813

Issue: [RE-1813](https://rpc-openstack.atlassian.net/browse/RE-1813)

Depends on https://github.com/rcbops/rpc_component/pull/10

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -372,9 +372,17 @@ def archive_artifacts(Map args = [:]){
       }
 
       pubcloud.uploadArtifacts("artifact_types": artifact_types)
-      if(fileExists(file: "artifact_public_url")){
-        artifact_public_url = readFile(file: "artifact_public_url")
-        currentBuild.description = "<h2><a href='"+artifact_public_url+"'>Build Artifacts</a></h2>"
+      def buildArtifacts
+      if (fileExists("build_artifacts.yml")){
+        buildArtifacts = readYaml file: "build_artifacts.yml"
+      } else{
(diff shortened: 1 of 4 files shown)


git-harry:
recheck_unit_tests

git-harry:
recheck_all

git-harry:
recheck_all

git-harry:
recheck_all

pipeline_steps/common.groovy
```diff
@@ -372,9 +372,23 @@ def archive_artifacts(Map args = [:]){
...
+      if (fileExists("build_artifacts.yml")){
+        buildArtifacts = readYaml file: "build_artifacts.yml"
+      } else{
+        buildArtifacts = ["artifacts": []]
+      }
+      println "Uploaded artefact details:\n${buildArtifacts}"
+      buildArtifacts.artifacts.each {k, v ->
+        // rpc-gating is skipped because it is not a component and so that unit tests can continue
+        if (k == "file" && v.container_name != "rpc-gating"){
+          addArtifactTypeToComponent(v.container_name, v.name, k, v.public_url, "RE")
```
hughsaunders: Does the public url change with every build? Could we check if the existing value is the same and shortcut this call?

git-harry: The assumption I've made is that only new and unique artefact stores are added given that's all that seems to be required. It can then be extended as the need arises. That said, I did forget to check if it was added so I'll fix that.

git-harry:
recheck_xenial_lint_test

PR #1137, opened 2018-10-15
Only configure fallback DNS for systemd-resolved

odyssey4me:
In https://github.com/rcbops/rpc-gating/pull/1135 we configured
systemd-resolved in nodepool bionic images to ensure that they
are able to resolve DNS until glean is fixed to configured the
systemd-networkd interfaces with DNS as well.

To ensure that the glean configuration will take preference when
a fix is included in the release, we switch the DNS configuration
here to be a fallback only. Any systemd-networkd will then take
preference, but fall back to this configuration if that does not
work.

Issue: [RI-514](https://rpc-openstack.atlassian.net/browse/RI-514)


hughsaunders:
recheck_all

hughsaunders:
recheck_all

PR #1138, opened 2018-10-15
Update hardening tests to use MNAIO images

odyssey4me:
In order to ensure that the hardening tests make use of
MNAIO images rather than do the deployment from scratch,
we rename the 'action' to 'hardening' rather than 'deploy'.

We also update the nodepool node to use bionic.

Finally we change the 'image' axis to be 'xenial_mnaio'
for pike, queens rocky and master rather than make mention
of 'loose_artifacts' given that these RPC-O branches do not
have any notion of artifacts. We restrict that naming to
only newton. We also add bionic VM tests for rocky and
master.

Issue: [RE-2053](https://rpc-openstack.atlassian.net/browse/RE-2053)

### DIFF
--- a/rpc_jobs/hardening.yml
+++ b/rpc_jobs/hardening.yml
@@ -1,20 +1,54 @@
 - project:
-    name: "hardening-repo-rpc-openstack-mnaio-postmerge"
+    name: "hardening-rpc-openstack-bionic"
+    repo_name: "hardening"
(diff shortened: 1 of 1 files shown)


odyssey4me:
@git-harry The hardening repo doesn't need any changes. It just clones the rpc-openstack repo and runs the pre/run/post stages from it.

PR #1140, opened 2018-10-15
Ensure hardening job image names match RPC-O jobs

odyssey4me:
To allow the hardening jobs to simply use the RPc-O MNAIO
images, the 'image' axis names need to match. This patch
ensures that is the case.

Issue: [RE-2053](https://rpc-openstack.atlassian.net/browse/RE-2053)


odyssey4me:
recheck_all

PR #1147, opened 2018-10-19
Allow commit_and_pull_request.sh to update branch

git-harry:
If the branch already exists, the `git push` command fails. This change
forces the push to ensure that the script can always update an existing
pull request.

JIRA: RE-2100

Issue: [RE-2100](https://rpc-openstack.atlassian.net/browse/RE-2100)


hughsaunders:
recheck_unit_tests

PR #1157, opened 2018-10-31
Switch non-critical jobs to run weekly and add LON region to nodepool

odyssey4me:
In order to cope with a resource shortage in public
cloud's OnMetal service, we reduce the frequency
for some non-critical jobs to run only weekly.

This can be reverted once the resource constraints
have been lifted.

In order to provide additional capacity for tests using
OnMetal nodes, we add the LON region to nodepool and
configure it only for those specific labels. We also add
the graphing for the new region to grafana.

Given that the 'ubuntu-xenial-om-io2' label is no longer
in use, we also reduce the min-ready for that node type
to 0.

Finally, the LON region uses a different authentication
endpoint so we add a new 'cloud' to the clouds.yaml file
which provides the appropriate account credentials and
the right endpoint.


nodepool/files/nodepool.yml
```diff
@@ -284,6 +284,21 @@ providers:
...
 
+  - name: pubcloud-lon
+    region-name: 'LON'
+    cloud: pubcloud_uk
+    boot-timeout: 600
+    hostname-format: nodepool-{label.name}-{provider.name}-{node.id}
+    cloud-images: *provider_cloudimage_block
+    image-name-format: nodepool-{image_name}-{timestamp}
+    pools:
+      - name: main
```
hughsaunders:
```suggestion
      - name: onmetal
```

grafana/grafyaml/dashboard-nodepool.yaml
```diff
@@ -262,9 +270,26 @@ dashboard:
...
+            - target: alias(stats.gauges.nodepool.provider.pubcloud-lon.nodes.used, 'Used')
+            - target: alias(stats.gauges.nodepool.provider.pubcloud-lon.nodes.deleting, 'Deleting')
+            - target: alias(stats.gauges.nodepool.provider.pubcloud-lon.max_servers, 'Max')
+          seriesOverrides:
+            - alias: Max
+              stack: False
         - title: Test Nodes (ORD)
           type: graph
-          span: 4
+          span: 3
```
hughsaunders: Whats `span` here?

odyssey4me: It's the width of the graph, spanning x columns. Given that we're adding a new one they need to be reduced to fit into the same space.

hughsaunders: Makes sense, thanks 👍

odyssey4me:
recheck_unit_tests

odyssey4me:
recheck_all

hughsaunders:
recheck_all

PR #1162, opened 2018-11-01
Add test node set for phobos ironic nodes and add Phobos into Grafana

odyssey4me:
In order to verify the capability to use Ironic nodes
in Phobos, we add a new set of labels for them. Once
it's verified that they work we can add them to the
general pool of ubuntu-*-om-io2 nodes.

To provide visibility into what's happening in Phobos
we add the appropriate dashboard configuration.

To accommodate the new graph on the bottom row we
reduce the span size to ensure it still fits in.

Issue: [RE-2051](https://rpc-openstack.atlassian.net/browse/RE-2051)

### DIFF
--- a/nodepool/templates/nodepool.yml.j2
+++ b/nodepool/templates/nodepool.yml.j2
@@ -81,12 +81,25 @@ labels:
     min-ready: 1
     max-ready-age: 86400
 
+  # onmetal io2
+  - name: diag-ubuntu-bionic-om-io2
+    min-ready: 0
+    max-ready-age: 86400
+  - name: diag-ubuntu-xenial-om-io2
+    min-ready: 0
+    max-ready-age: 86400
+
 providers:
   - name: phobos-nodepool
     region-name: "RegionOne"
     cloud: phobos_nodepool
-    boot-timeout: 120
+    boot-timeout: 600
     hostname-format: nodepool-{label.name}-{provider.name}-{node.id}
+    cloud-images:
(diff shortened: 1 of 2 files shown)


nodepool/templates/nodepool.yml.j2
```diff
@@ -159,6 +172,22 @@ providers:
             flavor-name: releng-7
             key-name: jenkins
             console-log: true
+      - name: onmetal
+        max-servers: 10
```
coreywright: what exactly does `max-servers` mean for a pool?

i mean, i presume that there will never be any more than 10 `ironic-storage-perf` nodes used from phobos at any time, but what is nodepool's allocation algorithm:
* allocate all 10 right away?
* allocate one at a time as needed, up to 10, and maintaining all 10 forevermore?
* allocate one at a time as needed, up to 10, but then releasing nodes after a period of lack of demand?

i ask because we only have 13 storage-perf nodes in phobos, with 9 of those currently in use, and i want to be prepared to explain when there are none available why jenkins is holding 10 of them but potentially not using them (eg "it's only been an hour or two since all the nightly jobs finished so the nodepool release timeout hasn't occurred, but the majority should be freed up in another hour or two.")

odyssey4me: @coreywright It means that nodepool will never use more than 10. It only allocates nodes when they are asked for by a test, or to satisfy the `min-ready` for the node type - but the `min-ready` can be satisfied by any of the regions.

PR #1167, opened 2018-11-02
RE-2085 Fix PR aborts post maintenance

hughsaunders:
Previously, there was no check for the status of the maintenance ticket,
so builds were still aborted even when the ticket was progressed
past doing. This commit adds a check so that once a maintenance
ticket is moved past doing it won't count towards future mainteanances.

Issue: [RE-2085](https://rpc-openstack.atlassian.net/browse/RE-2085)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -2336,7 +2336,7 @@ Boolean willOverlapMaintenanceWindow(){
 Boolean issueExistsForNextMaintenanceWindow(project="RE"){
   LocalDateTime nextWindow = getNextMaintenanceWindowStart()
   String nextWindowDateString = nextWindow.format(DateTimeFormatter.ISO_LOCAL_DATE)
-  List issues = jira_query("project=\"${project}\" AND summary ~ '${nextWindowDateString} Maintenance Window'")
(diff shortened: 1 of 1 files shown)


pipeline_steps/common.groovy
```diff
@@ -2336,7 +2336,7 @@ Boolean willOverlapMaintenanceWindow(){
 Boolean issueExistsForNextMaintenanceWindow(project="RE"){
   LocalDateTime nextWindow = getNextMaintenanceWindowStart()
   String nextWindowDateString = nextWindow.format(DateTimeFormatter.ISO_LOCAL_DATE)
-  List issues = jira_query("project=\"${project}\" AND summary ~ '${nextWindowDateString} Maintenance Window'")
+  List issues = jira_query("project=\"${project}\" AND summary ~ '${nextWindowDateString} Maintenance Window' and STATUS not in (Finished)")
```
hughsaunders: Note the same clause was used previously: https://github.com/rcbops/rpc-gating/blob/8e1bd5bb739b7fa2816ebf2f082dad9fe6cde3c0/pipeline_steps/common.groovy#L2382

PR #1170, opened 2018-11-02
Adds temporary job to capture older versions

antonym:
Captures MNAIO environment after the initial deployment of a
version has ran.  Leverages the workarounds to install an older
version so that we can create a library of older versions and
focus on just testing upgrades rather than deployment of an older
verision.

Uses snapcap as the action which skips the upgrade part of
the job and captures the image on success. Testing out with one
image to see how it works.  

The upgrade_to is ignored but the format is left as is to avoid making a lot of changes to the existing process.

Relies on https://github.com/rcbops/rpc-upgrades/pull/276

### DIFF
--- a/rpc_jobs/rpc_upgrades.yml
+++ b/rpc_jobs/rpc_upgrades.yml
@@ -261,6 +261,22 @@
     CRON: "{CRON_WEEKLY}"
 
 - project:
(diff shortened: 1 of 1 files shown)


antonym:
recheck_unit_tests

PR #1172, opened 2018-11-06
RE-2000 Quote JQL terms that may contain - or _

hughsaunders:
Jira processes string search terms including translating underscores
to spaces. To prevent this, much quoting is required.

Issue: [RE-2000](https://rpc-openstack.atlassian.net/browse/RE-2000)

### DIFF
--- a/scripts/jirautils.py
+++ b/scripts/jirautils.py
@@ -52,8 +52,8 @@ def _get_or_create_issue(project, status, labels, description, summary):
     issue = None
 
     # Check for existing issues
-    query = ("{label_terms} AND project = \"{p}\" AND status = {s}"
-             " AND summary ~\"{summary}\""
+    query = (r'{label_terms} AND project = "{p}" AND status = {s}'
+             r' AND summary ~"\"{summary}\""'
              .format(label_terms=get_label_query_terms(labels),
                      p=project, s=status, summary=summary))
     issues = issues_for_query(query)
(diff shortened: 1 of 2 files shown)


hughsaunders:
recheck_whisperer

scripts/jirautils.py
```diff
@@ -70,6 +70,9 @@ def _get_or_create_issue(project, status, labels, description, summary):
                          il=",".join(i.key for i in issues),
                          i=issue))
     else:
+        LOGGER.debug("Query: {q} Returned 0 issues. "
```
git-harry: When no issue or more than one issue is found, the query is logged, but not when there is only one. Maybe it would be better to log the query after is is defined and not in a subset of the sections of this if block.

hughsaunders: The logs are for unusual situations or side effects. When there are 0 issues, we create a new issue and log that. When there are >1 then something has gone wrong as there shouldn't be duplicate issues matching the criteria, so we log that as an anomaly. When exactly one issue is found no further action is required, we just return the issue key, this is the most "normal" path and IMO doesn't require a log entry.

git-harry: Logs are information, what you describe is one use case for them. The existing calls are being logged at the debug level which suggests they are for when there is something to debug, for example what happens if one issue is returned but it is not the one that was expected? How would one effectively debug that without knowing the query that was made?

PR #1176, opened 2018-11-08
Adds jobs to capture older versions of RPC-O

antonym:
Adds older versions of jobs we need to capture for
testing, set at a monthly cadence but really can
be removed once the image has been captured.

Also adds an initial testing job for the image we've
captured already.

### DIFF
--- a/rpc_jobs/rpc_upgrades.yml
+++ b/rpc_jobs/rpc_upgrades.yml
@@ -270,8 +270,38 @@
     scenario:
       - swift
     action:
+      - mitaka_to_r14.current_snapcap
+      - r13.0.0_to_r14.current_snapcap
+      - r13.0.2_to_r14.current_snapcap
+      - r13.1.2_to_r14.current_snapcap
       - r13.1.3_to_r14.current_snapcap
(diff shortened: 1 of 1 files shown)


rpc_jobs/rpc_upgrades.yml
```diff
@@ -270,8 +270,38 @@
...
+      - liberty_to_r14.current_snapcap
+      - r12.2.8_to_r14.current_snapcap
+      - r12.2.5_to_r14.current_snapcap
+      - r12.2.2_to_r14.current_snapcap
+      - r12.1.2_to_r14.current_snapcap
+      - kilo_to_r14.current_snapcap
     jira_project_key: "RLM"
+    jobs:
+      - 'PM_{repo_name}-{branch}-{image}-{scenario}-{action}'
+    CRON: "{CRON_MONTHLY}"
```
odyssey4me: If these are for one-off image generation, then CRON can be set to an empty string to make these jobs all manually executed. That way you can build them on demand, but leave them in place.

antonym:
recheck_unit_tests

PR #1183, opened 2018-11-14
RE-2184 Switch MK8s Checkmarx jobs to ssh clones

hughsaunders:
This allows keys to be used to access private repos.

Also set thresholds to 0 not 1, so that any vulnerability causes a scan to fail. 

Issue: [RE-2184](https://rpc-openstack.atlassian.net/browse/RE-2184)


dealako:
recheck_xenial_standard_job_pre_merge_test

dealako:
recheck_all

PR #1202, opened 2018-11-27
Reduce retention and leave logs uncompressed for 5 days

odyssey4me:
A change [1] in the workflow-job-plugin has resulted in jenkins
no longer being able to render compressed job logs. As such, we
need to disable compression. We are not able to expand our
available disk space for logs and cannot meet our current retention
without compression, so we have to reduce the retention and disable
the compression for anything younger than 5 days.

[1] https://github.com/jenkinsci/workflow-job-plugin/commit/46bf528f5965ecb94180d2a01d91bf7cb05da9f5
JIRA: RE-2215

Issue: [RE-2215](https://rpc-openstack.atlassian.net/browse/RE-2215)

### DIFF
--- a/rpc_jobs/build_summary.yml
+++ b/rpc_jobs/build_summary.yml
@@ -11,7 +11,7 @@
       - rpc_gating_params
     properties:
       - build-discarder:
-          days-to-keep: 30
+          days-to-keep: 10
     dsl: |
       library "rpc-gating@${RPC_GATING_BRANCH}"
       common.globalWraps(){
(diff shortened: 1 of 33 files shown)


hughsaunders:
Thanks for this @odyssey4me, I'm working on the plugin downgrade in RE-2215, if thats unsuccessful, we'll merge this.

hughsaunders:
Thanks for this @odyssey4me, we didn't need it in the end as the plugin downgrade worked.

PR #1206, opened 2018-11-27
ASC-1251 Add asc checkmarx jobs

chundu:
This PR will add Checkmarx scan jobs against ASC team repos. A new job config file 'rpc_asc.yml' was added.

Issue: [ASC-1251](https://rpc-openstack.atlassian.net/browse/ASC-1251)

### DIFF
--- a/rpc_jobs/rpc_asc.yml
+++ b/rpc_jobs/rpc_asc.yml
@@ -1,0 +1,35 @@
+- project:
+    name: 'asc-checkmarx'
+    jira_project_key: "ASC"
+    scan_type:
+      - all
+      - pci
+    trigger:
+      - PM
+    repo_name:
+      - zigzag:
+          repo_url: "git@github.com:rcbops/zigzag"
+          branch: master
+      - pytest-zigzag:
+          repo_url: "git@github.com:rcbops/pytest-zigzag"
+          branch: master
+      - pytest-rpc:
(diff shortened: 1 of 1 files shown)


rpc_jobs/rpc_asc.yml
```diff
@@ -0,0 +1,35 @@
...
+    name: 'asc-checkmarx'
+    jira_project_key: "ASC"
+    scan_type:
+      - all
+      - pci
+    trigger:
+      - PM
+    repo_name:
+      - zigzag:
+          repo_url: "git@github.com:rcbops/zigzag"
```
johnduarte: I don't believe that Jenkins is using SSH to connect to github. Therefore all of the `repo_url` strings need to use the `https` endpoint. All of the url values should be changed to the following format.

```
https://github.com/rcbops/<repo-name>
```
For `zigzag` the string would be as follows:
```
https://github.com/rcbops/zigzag
```

hughsaunders: Currently SSH urls must be used for private repos, as Jenkins uses ssh key based authentication for clones. However SSH urls cannot be used for projects that include PR triggered jobs. At some point we will need to PR tests for a private repo and will have to fix that conflict. Currently checkmarx jobs are not triggered by PRs so using ssh URLs is fine in this case.

PR #1212, opened 2018-11-29
RE-2217 Reflect sub build status in job result

hughsaunders:
The bump snapshots job builds new images by kicking off a new Jenkins
buid for each iamge. If those sub jobs fail, image update is skipped
for that image, but the stage shows green in the UI and the parent job
ends as successful.

This commit changes the error handling so that builds that fail show
up as failed parallel stages in blue ocean. It also fails the parent job
if any of the sub jobs fail. Note that a Jira issue is not created for
failure unless a sub job has failed 7 times consecutively.

### DIFF
--- a/rpc_jobs/bump_snapshots.yml
+++ b/rpc_jobs/bump_snapshots.yml
@@ -14,158 +14,171 @@
 
       library "rpc-gating@${RPC_GATING_BRANCH}"
       common.globalWraps(){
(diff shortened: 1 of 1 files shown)


hughsaunders:
Two sub builds failed in both the following examples:

Before: 
screen shot 2018-11-28 at 14 18 10

After:
screen shot 2018-11-29 at 16 33 23

PR #1215, opened 2018-12-03
[RI-520] Remove Pike Octavia PM Job

d34dh0r53:
Per the comments in issue RI-520 we are removing the Pike PM job for 
Octavia as we do not deploy Octavia on Pike.

Issue: [RI-520](https://rpc-openstack.atlassian.net/browse/RI-520)

### DIFF
--- a/rpc_jobs/rpc_octavia.yml
+++ b/rpc_jobs/rpc_octavia.yml
@@ -32,40 +32,6 @@
     jobs:
       - 'PR_{repo_name}-{series}-{image}-{scenario}-{action}'
 
-#- project:
-#    name: "rpc-octavia-pike-pre-merge"
-#
-#    repo_name: "rpc-octavia"
-#    # URL of the rpc-octavia repository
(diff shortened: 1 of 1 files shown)


d34dh0r53:
Good catch @hughsaunders, I've removed the pre-merge job as well.

rpc_jobs/rpc_octavia.yml
```diff
@@ -32,39 +32,39 @@
...
-    # Octavia ignores that setting for now
-    action:
-      - "test"
-
-    jira_project_key: "K8S"
-
-    # Link to the standard pre-merge-template
-    jobs:
-      - 'PR_{repo_name}-{series}-{image}-{scenario}-{action}'
+#- project:
```
odyssey4me: I'm not sure I understand the purpose of commenting these out. It results in noise which has no purpose. If the intent is to restore this one day, then a look run git history provides what it was. I would recommend simply removing it.

PR #1224, opened 2018-12-05
Change job scenario and action for osp13 gates

hmakkapati:


hmakkapati:
This blocks on https://github.com/rcbops/rpc-maas/pull/598

PR #1228, opened 2018-12-06
Shortens incremental action to fit char limit

antonym:
* Shortens the incremental action to give a bit more breathing room
  on the job character limit.
* Adds 16.0.0-beta.1 for customer upgrade test
* Adds Newton to Rocky, Pike to R, Queens to R for initial testing
* Removes hold_on_error until things are more stable

Depends on https://github.com/rcbops/rpc-upgrades/pull/291


antonym:
recheck_xenial_lint_test

PR #1233, opened 2018-12-07
RE-2238 Fix confluence wiki page update

hughsaunders:
Previously only the release notes column was run through the
urlize filter, this meant that a stray ampersand could make it into
the comments filter, and cause confluence to return 400.

Issue: [RE-2238](https://rpc-openstack.atlassian.net/browse/RE-2238)


hughsaunders:
Alternative to #1229

PR #1237, opened 2018-12-10
RE-2062 Provide ssh key for lint

hughsaunders:
JJB definitions can be stored in project repos, some of which are private. An
ssh-agent is provided to the lint script so it can clone all the necessary
repos for testing JJB definitions.

Issue: [RE-2062](https://rpc-openstack.atlassian.net/browse/RE-2062)


hughsaunders:
Lint failed because the job hasn't been rebuilt to include the new credential. I've manually added the credential to the job, and will retry. recheck_xenial_lint_test

hughsaunders:
Issue was incorrect, should have been RE-2262

PR #1238, opened 2018-12-10
ASC-1371 Move sdqc action from PM jobs to PR jobs

johnduarte:
This commit moves the sdqc action from the PM jobs to the PR jobs on the master, rocky, and queens branches.


chundu:
I had a question in the JIRA ticket.

PR #1241, opened 2018-12-11
Upgrade from Newton release for Genentech gating job

rexredinger:
This gate is intended to provide some testing that approximates the Genentech upgrade scenario.  Initially, the understanding was that they would be upgrading from r16.0.0-beta, but it turns out they are actually running r14.6.0.

### DIFF
--- a/rpc_jobs/rpc_upgrades.yml
+++ b/rpc_jobs/rpc_upgrades.yml
@@ -245,7 +245,7 @@
       - "pike_to_queens_inc"
       - "pike_to_rocky_inc"
       - "queens_to_rocky_inc"
-      - "r16.0.0-beta.1_to_queens_inc"
+      - "r14.6.0_to_queens_inc"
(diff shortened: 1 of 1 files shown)


rpc_jobs/rpc_upgrades.yml
```diff
@@ -245,7 +245,7 @@
       - "pike_to_queens_inc"
       - "pike_to_rocky_inc"
       - "queens_to_rocky_inc"
-      - "r16.0.0-beta.1_to_queens_inc"
```
odyssey4me: Are we to assume that https://github.com/rcbops/rpc-gating/commit/a93bc86c3c13c1e806bbf5481998119ac76a7c1d added the wrong one and that this fixes it? There's nothing in the PR message or commit msg to explain the replacement.

rexredinger: Updated my initial comment to help answer your question.

PR #1245, opened 2018-12-13
JJB Lint fixes

hughsaunders:
* Remove unnecessary call to configure_git
* Prevent jjb-path-setup from cloning job source repos that
  are cloned by other mechanisms. This includes rpc-gating
  which is provided by common.use_node and the repo currently
  being tested, as we want to use the version under test.
* Source venv before calling jjb-path-setup as it requires library
  from the venv
* Call the lint script from within the rpc-gating directory, as
  it expects paths relative to rpc-gating.
* Add the --python command to lint.sh when creating venvs to ensure
  the correct interpreter is installed in each venv
* Remove .git from the end of any ssh urls found in config. The suffix
  is not necessary and consistency is required.
* Add error message for when a componenet metdata file can't be found
  in a job source repo.

### DIFF
--- a/job_dsl/standard_job_lint_jenkins.groovy
+++ b/job_dsl/standard_job_lint_jenkins.groovy
@@ -2,11 +2,7 @@ library "rpc-gating-master"
 common.globalWraps(){
   common.standard_job_slave(env.SLAVE_TYPE) {
     try {
-      stage("Configure Git"){
-        common.configure_git()
-      }
-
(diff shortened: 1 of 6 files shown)


hughsaunders:
@dealako test run: https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/Lint-Jenkins_mk8s-master/detail/Lint-Jenkins_mk8s-master/22/pipeline

PR #1250, opened 2018-12-13
Fix Client Remote File Inclusion Vulnerability

dealako:
* Updated to use local CSS/JS libraries/files instead of remoting
downloading from CDN per Checkmarx recommendation: "Where possible, host
all script files locally, rather than remotely. Ensure that locally
hosted 3rd party script files are constantly updated and maintained."
* Added Rackspace favicon

JIRA: RE-2205

Issue: [RE-2205](https://rpc-openstack.atlassian.net/browse/RE-2205)


dealako:
recheck_unit_tests

PR #1257, opened 2019-01-03
RE-2265 Overrideable default for folder excludes

hughsaunders:
Previously a param was added to allow checkmarx jobs to exclude folders.
However that param didn't have a default value that used a jjb macro
so couldn't be overidden from a jjb project. This commit adds the
required macro and a default value.

Issue: [RE-2265](https://rpc-openstack.atlassian.net/browse/RE-2265)


hughsaunders:
This is an addition to https://github.com/rcbops/rpc-gating/pull/1256

PR #1263, opened 2019-01-03
RE-2298 Wiki update still failing

hughsaunders:
#1233 added urlize to a second field, but left two unfiltered. This commit filters the remaining two.

Issue: [RE-2298](https://rpc-openstack.atlassian.net/browse/RE-2298)


hughsaunders:
This doesn't actually solve the problem :(

PR #1272, opened 2019-01-07
Externalize Standard Job Timeout

dealako:
* Jobs can now override the job standard timeout

JIRA: RE-2307

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -1240,19 +1240,19 @@ void internal_slave(Closure body){
 }
 
 void standard_job_slave(String slave_type, Closure body){
-  timeout(time: 8, unit: 'HOURS'){
+  timeout(time: env.NODE_TIMEOUT_HR, unit: 'HOURS'){
     if (slave_type == "instance"){
       pubcloud.runonpubcloud(){
         body()
       }
     } else if (slave_type.startsWith("nodepool-")){
       use_node(slave_type){
-        body();
+        body()
       }
     }else if (slave_type == "shared"){
       // don't need to wrap body in shared_slave here
       // as globalWraps will have already allocated a shared slave executor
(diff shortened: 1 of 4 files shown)


pipeline_steps/common.groovy
```diff
@@ -1240,19 +1240,19 @@ void internal_slave(Closure body){
 }
 
 void standard_job_slave(String slave_type, Closure body){
-  timeout(time: 8, unit: 'HOURS'){
+  timeout(time: env.JOB_TIMEOUT_HR, unit: 'HOURS'){
```
hughsaunders: This isn't really a job timeout, this is a slave timeout. A job may use multiple slaves, in parallel or serial, and its not unreasonable for the job as a whole to take longer than any one slave is used for. 

I guess what I'm saying is that if these are going to be tweak-able they should probably be tweak-able independently.

Both should have a default, so most users will only need to set `JOB_TIMEOUT` but `AGENT_TIMEOUT` could be useful in some situations.

rpc_jobs/unit/job_timeout.yml
```diff
@@ -0,0 +1,37 @@
+- job:
+    name: RE-unit-test-job-timeout
```
hughsaunders: I think this should be a lint check.

pipeline_steps/common.groovy
dealako: Refactored to accomodate additional parameter.

rpc_jobs/unit/job_timeout.yml
dealako: Removed - moved logic to jjb_lint.py

dealako:
Closing this PR as we need these changes to be introduced as two separate commits to allow our existing pre-merge checks to pass.

PR #1273, opened 2019-01-07
RI-613: Add Bionic Job for Designate

leader716:
Issue: [RI-627](https://rpc-openstack.atlassian.net/browse/RI-627)

Issue: [RI-613](https://rpc-openstack.atlassian.net/browse/RI-613)


leader716:
recheck_all

PR #1275, opened 2019-01-09
Externalize Standard Job Timeout

dealako:
* Removed node timeout - will rely on job timeout only
* Leveraged BUILD_TIMEOUT_HR for global wraps timeout
* Added lint test for timeout value/range

JIRA: RE-2307

Issue: [RE-2307](https://rpc-openstack.atlassian.net/browse/RE-2307)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -1240,56 +1240,53 @@ void internal_slave(Closure body){
 }
 
 void standard_job_slave(String slave_type, Closure body){
-  timeout(time: 8, unit: 'HOURS'){
-    if (slave_type == "instance"){
-      pubcloud.runonpubcloud(){
+  if (slave_type == "instance") {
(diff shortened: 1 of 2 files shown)


pipeline_steps/common.groovy
```diff
@@ -1672,7 +1669,7 @@ void globalWraps(Closure body){
   // global timeout is long, so individual jobs can set shorter timeouts and
   // still have to cleanup, archive atefacts etc.
   timestamps{
-    timeout(time: 10, unit: 'HOURS'){
+    timeout(time: env.BUILD_TIMEOUT_HRS, unit: 'HOURS'){
```
hughsaunders: You could get this to parse gate by setting a default for when the env var isn't available. 
```suggestion
    timeout(time: env.BUILD_TIMEOUT_HRS ?: 3, unit: 'HOURS'){
```

PR #1276, opened 2019-01-10
Checkmarx Plugin Retry Enhancements

dealako:
* Increased the number of Checkmarx retries
* Added exponential backoff for the retry interval

JIRA: RE-2312

### DIFF
--- a/pipeline_steps/checkmarx.groovy
+++ b/pipeline_steps/checkmarx.groovy
@@ -24,7 +24,8 @@ def scan(String scan_type, String repo_name, String exclude_folders){
             throw new Exception("Invalid scan type: ${scan_type}, should be default or pci")
         }
         // This step has a habit of throwing NPEs, retry it. RE
-        retry(3) {
+        waitTime = 30
+        retry(10) {
             // Try within retry so that sleep can be added on failure.
             // This may help if the issue is at the remote end.
             try {
@@ -68,7 +69,9 @@ def scan(String scan_type, String repo_name, String exclude_folders){
                 )
             } catch (Exception e){
                 print ("Caught exception while running checkmarx scan: "+e)
-                sleep(time: 30, unit: "SECONDS")
+                sleep(time: waitTime, unit: "SECONDS")
+                // Exponential backoff - double the wait for each retry
+                waitTime *= 2
                 // exception must propagate back to the retry call
                 throw e
             } //try


pipeline_steps/checkmarx.groovy
```diff
@@ -24,7 +24,8 @@ def scan(String scan_type, String repo_name, String exclude_folders){
             throw new Exception("Invalid scan type: ${scan_type}, should be default or pci")
         }
         // This step has a habit of throwing NPEs, retry it. RE
-        retry(3) {
+        waitTime = 30
```
hughsaunders: This could potentially timeout the job, as the max wait time is 8.5 hrs.

I suggest starting at 8s, then the max wait time is ~ 2.25 hours. 

```
In [19]: def exwait(wait, tries):
    ...:     total=0
    ...:     wait /= 2
    ...:     for _ in range(tries):
    ...:         wait *= 2
    ...:         total += wait
    ...:     return total
    ...:
    ...:

In [20]: exwait(30,10) / (60*60.0)
Out[20]: 8.525

In [21]: exwait(8,10) / (60*60.0)
Out[21]: 2.2733333333333334

In [22]: exwait(4,10) / (60*60.0)
Out[22]: 1.1366666666666667

```

```suggestion
        waitTime = 8
```

pipeline_steps/checkmarx.groovy
```diff
@@ -68,7 +69,9 @@ def scan(String scan_type, String repo_name, String exclude_folders){
                 )
             } catch (Exception e){
                 print ("Caught exception while running checkmarx scan: "+e)
-                sleep(time: 30, unit: "SECONDS")
+                sleep(time: waitTime, unit: "SECONDS")
```
hughsaunders: If a whole load of jobs hit checkmarx, then start backing off, they'll all use the same backoff pattern and may hit the same contention issue when retrying. This could be mitigated by adding a small random value to the wait time. 

```suggestion
                sleep(time: waitTime + random.randrange(0, 60), unit: "SECONDS")
```

Probably have to import random for that to work.

pipeline_steps/checkmarx.groovy
```diff
@@ -68,7 +73,10 @@ def scan(String scan_type, String repo_name, String exclude_folders){
...
```
jlpogue3: I think this would re-introduce the problem that Hugh mentioned before, so the waitTime would get to be too large.

dealako: I believe I did - the initial value is 8 seconds (instead of 30 seconds) - we ramp up from there.

PR #1278, opened 2019-01-11
Build RPC-O newton deploy artifacts as part of MNAIO job

odyssey4me:
RPC-O has a deployment artifact build job which must be executed
before the MNAIO test runs. The previous mechanism of running it
periodically resulted in the MNAIO job sometimes failing due to
missing artifacts.

This patch implements a job wrapper and applies it to the MNAIO
job to make it kick off the artifact build job first, then
continue on to the MNAIO test. By doing this the artifacts will
always be present before the MNAIO test runs.

Given that the MNAIO job runs daily and it will run the artifact
build job, we remove the timer trigger for the artifact build job.
We have no need to run the same job twice a day.

As part of this we also move the artifact build job definition from
its own file into the rpc_openstack file so that all jobs for RPC-O
are present in the same file.

Issue: [RE-1608](https://rpc-openstack.atlassian.net/browse/RE-1608)


pipeline_steps/common.groovy
```diff
@@ -1595,6 +1611,7 @@ void setTriggerVars(){
 List stdJobWrappers(String wrappers){
   Map availableWrappers = [
     "phobos_vpn": {body -> connect_phobos_vpn(); body()}
```
hughsaunders:
```suggestion
    "phobos_vpn": {body -> connect_phobos_vpn(); body()},
```

PR #1280, opened 2019-01-11
Use nodepool images for AIO snapshot builds

odyssey4me:
Rather than use public cloud images, which have proven to be unreliable
due to the nova-agent not starting properly, we enable the jenkins node
build to use the images built by nodepool instead.

The playbook fetches the list of available images, then uses the latest
available one for the build.

This brings the image build process closer to what is used for all tests
and should be more reliable given that nodepool does not use nova-agent.

Issue: [RE-2278](https://rpc-openstack.atlassian.net/browse/RE-2278)

### DIFF
--- a/playbooks/allocate_pubcloud.yml
+++ b/playbooks/allocate_pubcloud.yml
@@ -39,6 +39,20 @@
           {%- set region_list = (fallback_regions | replace(',', ' ')).split() | map('trim') | map('upper') | select | shuffle %}
           {{- (region_list | length == 1) | ternary(region_list, region_list | difference([regions_shuff[0]])) }}
 
+    - name: Determine the latest image name when required
+      when:
+        - image | match('^nodepool-.*')
+      block:
+        - name: Get the list of images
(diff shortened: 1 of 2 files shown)


playbooks/allocate_pubcloud.yml
```diff
@@ -39,6 +39,19 @@
...
           {{- (region_list | length == 1) | ternary(region_list, region_list | difference([regions_shuff[0]])) }}
 
+    - name: Determine the latest image name when required
+      when:
+        - image | match('^nodepool-.*')
+      block:
+        - name: Get the list of images
+            os_image_facts:
+            cloud: "{{ cloud_name }}"
+            region_name: "{{ regions_shuff[0] }}"
```
dealako: We're seeing a syntax error on `os_image_facts:`, need to indent `cloud` and `region_name` properties?

odyssey4me: Yep, ``os_image_facts`` was incorrectly indented.

PR #1284, opened 2019-01-14
RE-2310 Idempotent Releases

hughsaunders:
This commit ensures that releases can be re-run by skipping parts
that have already succeeded.

Issue: [RE-2310](https://rpc-openstack.atlassian.net/browse/RE-2310)

### DIFF
--- a/scripts/confluenceutils.py
+++ b/scripts/confluenceutils.py
@@ -11,16 +11,13 @@ import click
 from jinja2 import Template
 import requests
 
-logging.basicConfig()
-logger = logging.getLogger("confluenceutils")
-
 
 class PageNotFound(Exception):
     pass
 
 
 class Confluence(object):
-    """https://developer.atlassian.com/cloud/confluence/rest/"""
+    """https://developer.atlassian.com/cloud/confluence/rest/ ."""
 
     def __init__(self, username, password, base_url):
         self.session = requests.Session()
@@ -53,7 +50,7 @@ class Confluence(object):
         if resp_data["size"] == 0:
             raise PageNotFound
         elif resp_data["size"] > 1:
-            logger.error(
+            logging.error(
                 json.dumps(
                     resp_data,
                     sort_keys=True,
@@ -80,7 +77,7 @@ class Confluence(object):
         resp = self.session.post(self.content_url, json=content)
         resp.raise_for_status()
         resp_data = resp.json()
-        logger.info(
+        logging.info(
             "The following page has been created: {url}".format(
                 url=resp_data["_links"]["base"] + resp_data["_links"]["webui"],
             )
@@ -105,7 +102,7 @@ class Confluence(object):
         resp = self.session.put(url, json=content)
         resp.raise_for_status()
         resp_data = resp.json()
-        logger.info(
+        logging.info(
             "The following page has been updated: {url}".format(
                 url=resp_data["_links"]["base"] + resp_data["_links"]["webui"],
             )
@@ -173,7 +170,7 @@ def is_async_release(date, scheduled_window_days=7):
     else:
         is_async = True
 
-    logger.info(
+    logging.info(
         "This release has been classified as {release_type}.".format(
             release_type=("asynchronous" if is_async else "scheduled"),
         )
@@ -182,7 +179,9 @@ def is_async_release(date, scheduled_window_days=7):
 
 
 def get_annual_release_page(c, space_key, year, product_release_page_id):
-    """ The wiki is organised in a hierarchy:
+    """Get annual release page.
+
+    The wiki is organised in a hierarchy:
         Product Releases
             {year} Monthly Releases
                 {version}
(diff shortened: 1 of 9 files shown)


hughsaunders:
Ran two releases with the same component and version to test idempotency:
https://rpc.jenkins.cit.rackspace.net/job/RE-Release/352/ ✅ 
https://rpc.jenkins.cit.rackspace.net/job/RE-Release/353/ ✅

scripts/ghutils.py
```diff
@@ -436,12 +436,17 @@ def create_release(repo, version, bodyfile):
...
     ctx_obj.version = version
     try:
-        release = repo.create_release(
-            tag_name=version,
-            name=version,
-            body=release_notes,
-        )
-        logger.info("Release {} created.".format(version))
+        try:
+            release = [r for r in repo.iter_releases() if r.name == version][0]
```
git-harry: It might be worth rerunning this to take account of situations where something has changed or at least expose any difference.

hughsaunders: OK, I'll check the body matches as well.

hughsaunders:
Fresh release: https://rpc.jenkins.cit.rackspace.net/job/RE-Release/361/ ✅ 
Repeat release: https://rpc.jenkins.cit.rackspace.net/job/RE-Release/362/ ✅ 
```2019-01-17T14:07:20+0000 INFO:root:Tag '1.0.6' already exists and the ref matches, nothing to do.```
```2019-01-17T14:07:21+0000 INFO:root:Release 1.0.6 already exists, not creating.```

File: scripts/ghutils.py (hunk 1 above)
Comment (hughsaunders) [reply scripts/ghutils.py]: Done

File: scripts/release.py [hunk 2]
```diff
@@ -164,11 +164,17 @@ def publish_tag(version, ref, clone_dir):
     clone_dir = try_context(ctx_obj, clone_dir, "clone_dir", "clone_dir")
 
     repo = git.Repo(clone_dir)
-    repo.create_tag(version, ref, message=version)
-    repo.remotes.origin.push(version)
-    logger.info(
-        "Tag '{tag}' successfully pushed to repository.".format(tag=version)
-    )
+    if version not in repo.tags:
```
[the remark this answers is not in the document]
hughsaunders: Done

scripts/ghutils.py
```diff
@@ -436,19 +435,32 @@ def create_release(repo, version, bodyfile):
     # Store version in context for use in notifications
     ctx_obj.version = version
     try:
-        release = repo.create_release(
-            tag_name=version,
-            name=version,
-            body=release_notes,
-        )
-        logger.info("Release {} created.".format(version))
+        try:
```
git-harry: Making use of `else:` prevents the exception from unintentionally being caught from the wrong line of code and, at least in my opinion, is more readable.
```suggestion
        try:
            release = [r for r in repo.iter_releases() if r.name == version][0]
        except IndexError:
            release = repo.create_release(
                tag_name=version,
                name=version,
                body=release_notes,
            )
            logging.info("Release {} created.".format(version))
        else:
            if release.body != release_notes:
                raise ReleaseFailureException(
                    "Release failed. Github release {version} already exists"
                    " but its release notes don't match the release notes"
                    " currently being supplied."
                    .format(version=version)
                )
            logging.info("Release {} already exists, not creating."
                         .format(version))
```

scripts/ghutils.py
```diff
@@ -436,19 +435,32 @@ def create_release(repo, version, bodyfile):
...
+            release = [r for r in repo.iter_releases() if r.name == version][0]
+            if release.body != release_notes:
+                raise ReleaseFailureException(
```
git-harry: If someone has attempted to release the wrong SHA and the updates it, how would that be addressed?

scripts/ghutils.py
```diff
@@ -436,19 +435,32 @@ def create_release(repo, version, bodyfile):
...
+                    "Release failed. Github release {version} already exists"
```
git-harry: At this point, is the tag guaranteed to be pointing to the expected hash or is it simply confirming that the tag names match?

scripts/ghutils.py
hughsaunders: done

scripts/ghutils.py
hughsaunders: This just checks names, the tag creation function checks the tag SHAs and recreates if necessary.

scripts/confluenceutils.py
```diff
@@ -11,16 +11,13 @@
 from jinja2 import Template
 import requests
 
-logging.basicConfig()
-logger = logging.getLogger("confluenceutils")
```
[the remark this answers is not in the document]
hughsaunders: The previous mechanism didn't work (info and below were not logged regardless of `--debug` so no data is actually lost in this transition. I'd argue that data is gained because the message we expected to be printed are now printed. 

However if you can point me at a simpler way to fix this, I'm game.

PR #1288, opened 2019-01-16
Ensure /usr/bin/python is present on target hosts

odyssey4me:
When using the setup_openstack_instances playbook, the target
hosts may not have /usr/bin/python present on them. This is
the case for Ubuntu Xenial in the new Public Cloud images.

Without /usr/bin/python present, Ansible cannot gather facts
from the target host, so we need to use the raw/script module
to get it there before any subsequent plays which use facts.

Issue: [RE-2323](https://rpc-openstack.atlassian.net/browse/RE-2323)


odyssey4me:
Tested to validate that this works: 
https://rpc.jenkins.cit.rackspace.net/job/Setup-Graphite/910/

PR #1292, opened 2019-01-23
Correct global wrap timeout

odyssey4me:
In [1] the job timeout was set to 10 hours globally, but mistakenly
set to 3 hrs in the global wrap. This is causing job timeouts for
some of the longer running jobs (like MNAIO builds).

This patch just sets the global wrap timeout to the same value.

[1] https://github.com/rcbops/rpc-gating/pull/1275

JIRA: RE-2307

Issue: [RE-2307](https://rpc-openstack.atlassian.net/browse/RE-2307)


odyssey4me:
recheck_unit_tests

PR #1293, opened 2019-01-24
RE-404 Add Kronos registry support

git-harry:
See individual commits for details

JIRA: RE-404

Issue: [RE-404](https://rpc-openstack.atlassian.net/browse/RE-404)

### DIFF
--- a/rpc_jobs/build_summary.yml
+++ b/rpc_jobs/build_summary.yml
@@ -4,8 +4,8 @@
     project-type: workflow
     parameters:
       - string:
-          name: "DOCKER_IMAGE"
-          default: "rcbops/buildsummary"
+          name: "IMAGE_NAME"
+          default: "buildsummary"
           description: |
             Name of docker image to publish
       - rpc_gating_params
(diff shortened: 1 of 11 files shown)


rpc_jobs/build_summary.yml
```diff
@@ -15,24 +15,9 @@
...
-              # add docker registry creds.
-              # using creds in jenkins global config along with image.push()
-              # doesn't seem to work :(
-              # ok to use single quotes as the \$ is interpreted by groovy not bash
-              docker login --password '${env.DOCKERHUB_PASSWORD}' --username '${env.DOCKERHUB_USER}'
-              docker build . -t "${DOCKER_IMAGE}"
-              docker push "${DOCKER_IMAGE}"
-            """
-          }
+          docker_utils.pullOrBuild(DOCKER_IMAGE)
```
hughsaunders: Why not use the repo sha as the tag here?

git-harry: I misread `DOCKER_IMAGE` and thought it included the tag. I'll need to fix this.

rpc_jobs/build_summary.yml
```diff
@@ -44,7 +29,7 @@
     parameters:
       - string:
           name: "DOCKER_IMAGE"
-          default: "rcbops/buildsummary"
+          default: "buildsummary"
           description: |
             Hours. Instances older than this will be removed.
       - rpc_gating_params
```
hughsaunders: [L54](https://github.com/rcbops/rpc-gating/pull/1293/files#diff-9d1caf1170f2e050a035c7ff42bdd282R54) below might need to be wrapped in withinternalregistry?

rpc_jobs/build_summary.yml
```diff
@@ -15,24 +15,9 @@
...
-              # add docker registry creds.
-              # using creds in jenkins global config along with image.push()
-              # doesn't seem to work :(
-              # ok to use single quotes as the \$ is interpreted by groovy not bash
-              docker login --password '${env.DOCKERHUB_PASSWORD}' --username '${env.DOCKERHUB_USER}'
-              docker build . -t "${DOCKER_IMAGE}"
-              docker push "${DOCKER_IMAGE}"
-            """
-          }
+          docker_utils.pullOrBuild(DOCKER_IMAGE, "latest")
```
hughsaunders: If I understand correctly, this will result in an image being built once, then fetched on all subsequent occasions, even if the code has changed? The image name consists of registry,repo,name & tag. None of these change so what would cause the image to not be available and get rebuilt?

Also you could delete this job, and call pullorbuild in the build-summary job.

rpc_jobs/build_summary.yml
```diff
@@ -57,17 +19,37 @@
           days-to-keep: 3
     dsl: |
       library "rpc-gating@${RPC_GATING_BRANCH}"
+      common.shared_slave{
+        env.RE_JOB_REPO_NAME = "rpc-gating"
+        dir("rpc-gating/scripts/build_summary"){
+          String sha = sh(script: "git rev-parse --verify HEAD", returnStdout: true).trim()
+          image = docker_utils.pullOrBuild(IMAGE_NAME, sha)
```
hughsaunders: This can't be run on a shared slave, as it configures the VPN, which is system wide, and may conflict with other jobs.

So either this can be run on a single use slave of some type, or we can add a Kronos VPN service to the master as we currently have for Phobos, then the whole job can run on the master.

PR #1294, opened 2019-01-24
Implement a common method for rpc-gating runtime requirements

odyssey4me:
Currently there are multiple methods for preparing a host for
tests:

1. In the jenkins nodepool plugin configuration, there is a script
   which executes on all target nodes to prepare the node for use.
   This script is used to add the jenkins user, add the ssh keys,
   add the JDK, etc. This is used for preparing OnMetal/Ironic
   hosts where we do not control the image.

2. In the setup_openstack_instances playbook, we use the raw Ansible
   module to add python to the target host so that Ansible is able
   to execute other tasks and modules on that host.
   This is used for preparing VM's where we do not control the image.

3. In the allocate_pubcloud playbook we use the raw and script
   modules to ensure that the apt sources are configured, apt debug
   logging is enabled, and that python and python-yaml are installed
   on the target host.
   This is used for preparing VM's where we do not control the image.

In this patch, we combine these all into a single script which is
used as userdata provided to cloud-init [1], which is present in all
the images for public cloud and Phobos.

In order to use this for nodepool, we also need to update the nodepool
SHA we use to include the support for userdata.

As part of this, we also make the following minor adjustments:

1. In the allocate_pubcloud playbook we use the Ansible playbook_dir
   magic variable, rather than refer to the WORKSPACE environment
   variable. This makes the playbook executable for testing without
   having to set the WORKSPACE environment variable.

2. In the user_data_publoud script, we remove the unattended upgrade
   disabling. Now that all RPC-O tests make use of nodepool images
   the issue which brought that content there is no longer relevant.

[1] https://cloud-init.io
Issue: [RE-1523](https://rpc-openstack.atlassian.net/browse/RE-1523)

### DIFF
--- a/scripts/user_data_pubcloud.sh
+++ b/scripts/user_data_pubcloud.sh
@@ -1,23 +1,123 @@
-#!/bin/bash
+#!/bin/bash -e
 
-# This file is used when creating public
-# cloud instances.
-# See playbooks/allocate_pubcloud.yml
+# This script is used as a user data file when building
+# machines which use images that are not managed by
+# nodepool. See https://cloud-init.io for more information
+# about how cloud-init uses user-data files.
+#
+# This script tries to ensure that the machine is prepared
+# in a consistent way and it has all the required packages
+# installed on it for jenkins/ansible to access it. Any
+# further preparation should be done using jenkins/ansible.
 
-source /etc/lsb-release
-
-# Delete configuration which enables automatic upgrades
-# to prevent the instance packages being upgraded before
-# the correct apt repositories have been configured.
-# ref: RE-458 / RE-473
-if [[ "${DISTRIB_CODENAME}" == "trusty" ]]; then
-    # The 'unattended-upgrades' package cannot
-    # be purged on trusty as it is a dependency
-    # for cloud-init. So instead we just remove
-    # the configuration.
-    rm -f /etc/apt/apt.conf.d/*unattended-upgrades
+# Get the distribution name
+echo -e "\n### Getting the distribution name ###\n"
+if [[ -e /etc/lsb-release ]]; then
+  source /etc/lsb-release
+  DISTRO_RELEASE=${DISTRIB_CODENAME}
+elif [[ -e /etc/os-release ]]; then
+  source /etc/os-release
+  DISTRO_RELEASE=${UBUNTU_CODENAME}
 else
-    export DEBIAN_FRONTEND=noninteractive
-    apt-get purge -y unattended-upgrades
+  echo "Unable to determine distribution due to missing lsb/os-release files."
+  exit 1
 fi
+
+# Save a backup of the original apt sources
+if [[ ! -f "/etc/apt/sources.list.original" ]]; then
+  echo -e "\n### Saving a backup of the original apt sources ###\n"
+  mv /etc/apt/sources.list /etc/apt/sources.list.original
+  echo -e "\n###Rewriting the apt sources ###\n"
+  DISTRO_MIRROR="http://mirror.rackspace.com/ubuntu"
+  DISTRO_COMPONENTS="main,universe"
+  cat << EOF >/etc/apt/sources.list
+deb ${DISTRO_MIRROR} ${DISTRO_RELEASE} ${DISTRO_COMPONENTS//,/ }
+deb ${DISTRO_MIRROR} ${DISTRO_RELEASE}-updates ${DISTRO_COMPONENTS//,/ }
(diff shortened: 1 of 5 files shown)


scripts/user_data_pubcloud.sh
```diff
@@ -1,23 +1,104 @@
...
-# to prevent the instance packages being upgraded before
-# the correct apt repositories have been configured.
-# ref: RE-458 / RE-473
-if [[ "${DISTRIB_CODENAME}" == "trusty" ]]; then
-    # The 'unattended-upgrades' package cannot
-    # be purged on trusty as it is a dependency
-    # for cloud-init. So instead we just remove
-    # the configuration.
-    rm -f /etc/apt/apt.conf.d/*unattended-upgrades
+#!/bin/bash -e
```
hughsaunders: This user data script is applied to all cloud instances allocated via ansible. If an nodepool built image is used, it will already have this config via DIB. Therefore this script needs to be idempotent.

scripts/user_data_pubcloud.sh
```diff
@@ -1,23 +1,104 @@
...
+
+# Enable debug logging for apt to make diagnosis of apt failures easier
+echo -e "\n### Enabling apt debug logging ###\n"
+echo 'Debug::Acquire::http "true";' > /etc/apt/apt.conf.d/99debug
+
+# Add jenkins user and group
+echo -e "\n### Adding the jenkins user and group ###\n"
+JENKINS_HOME="/var/lib/jenkins"
+groupadd jenkins
+useradd --gid jenkins \
```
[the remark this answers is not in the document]
hughsaunders: and the groupadd

scripts/user_data_pubcloud.sh
```diff
@@ -1,23 +1,104 @@
...
+
+# For Ubuntu Trusty add the openjdk PPA for access
+# to the openjdk8 packages
+echo -e "\n### Adding openjdk PPA for Ubuntu Trusty hosts ###\n"
+if [[ "${DISTRO_RELEASE}" == "trusty" ]]; then
+  add-apt-repository -y ppa:openjdk-r/ppa
+fi
+
+# Prepare the list of packages to install
+echo -e "\n### Preparing a list of packages to install ###\n"
```
odyssey4me: Only add the package to the install list if it's not already installed.

scripts/user_data_pubcloud.sh
```diff
@@ -1,23 +1,104 @@
...
+
+# Ensure everything has the right owner
+echo -e "\n### Ensuring jenkins owns all jenkins home files ###\n"
+chown -R jenkins:jenkins ${JENKINS_HOME}
+
+# For Ubuntu Trusty add the openjdk PPA for access
+# to the openjdk8 packages
+echo -e "\n### Adding openjdk PPA for Ubuntu Trusty hosts ###\n"
+if [[ "${DISTRO_RELEASE}" == "trusty" ]]; then
+  add-apt-repository -y ppa:openjdk-r/ppa
```
odyssey4me: What happens when it's already there?

scripts/user_data_pubcloud.sh
```diff
@@ -1,23 +1,104 @@
...
+
+# Fetch the rcbops public keys and add them to authorized_keys
+echo -e "\n### Fetching the rcbops public keys ###\n"
+SSH_PUBLIC_KEYS_URL="https://raw.githubusercontent.com/rcbops/rpc-gating/master/keys/rcb.keys"
+curl --connect-timeout 5 \
+     --retry 3 \
+     ${SSH_PUBLIC_KEYS_URL} > /tmp/ssh-public-keys
+
+echo -e "\n### Configuring authorized_keys for jenkins and root ###\n"
+for usr_home in /root ${JENKINS_HOME}; do
```
odyssey4me: If the keys are already there, what happens? If there are duplicates in authorized_keys, what happens?

scripts/user_data_pubcloud.sh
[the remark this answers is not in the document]
odyssey4me: Done.

scripts/user_data_pubcloud.sh
odyssey4me: It sill works if there are duplicates.

scripts/user_data_pubcloud.sh
odyssey4me: Done.

scripts/user_data_pubcloud.sh
odyssey4me: Done.

odyssey4me:
recheck_unit_tests

PR #1301, opened 2019-02-01
RE-2310 Fix release modifier regex matches

hughsaunders:
matcher.matches() will only return true if the pattern matches the whole
input string. The patterns in use do not cover the whole input string
only the trigger part and surrounding whitespace. That means that
using the current code .matches() won't trigger.

In a previous commit I removed .matches() and it caused release
jobs to fail because Jenkins CPS groovy does not cooerce matcher
objects to Boolean when matching method signatures, but vanilla
groovy does.

Consider the following snippet:
```
def func(Boolean f){
  print f
}

Boolean match(){
    return ("foo" =~ /oo/)
}
func(match())
```
This snippet returns true in groovy and:
java.lang.NoSuchMethodError: No such DSL method 'func'

in Pipeline/CPS groovy, because theres no method func(Matcher f).

This commit removes the .matches() call, but changes the operator to ==~
which requires an exact match, and always returns a Boolean.
To accomodate this, the patterns have been expanded to allow any
text before and after the required pattern.

Test code:
```
def func (Boolean f){
    print f
}

m = "before foo  after" ==~ /.*oo.*/
assert m instanceof Boolean
func(m)
```

This outputs "true" in both groovy and cps groovy.

Issue: [RE-2310](https://rpc-openstack.atlassian.net/browse/RE-2310)

### DIFF
--- a/pipeline_steps/common.groovy
+++ b/pipeline_steps/common.groovy
@@ -1976,7 +1976,7 @@ void runReleasesPullRequestWorkflow(String baseBranch, String prBranch, String j
 }
 
 Boolean skipPullRequestTests(String triggerPhrase){
-  return (ghprbCommentBody =~ /\s*${triggerPhrase}\s*/).matches()
+  return (ghprbCommentBody ==~ /.*${triggerPhrase}.*/)
 }
 
 /**
@@ -1986,7 +1986,7 @@ Boolean skipPullRequestTests(String triggerPhrase){
 * will be removed in order for the release to proceed
 */
 Boolean shouldReRelease(String reReleaseTriggerPhrase){
-  return (ghprbCommentBody =~ /\s*${reReleaseTriggerPhrase}\s*/).matches()
+  return (ghprbCommentBody ==~ /.*${reReleaseTriggerPhrase}.*/)
 }
 
 List getComponentChange(String baseBranch, String prBranch){


dealako:
@git-harry are you suggesting the `shouldReRelease()` function regex be something like: `/^${reReleaseTriggerPhrase}$/)` - basically the trigger phrase without any other text? (possibly stripping white space)

git-harry:
@dealako I misinterpreted the 'Re' in `shouldReRelease` as release engineering and not 're-' and so thought it was affecting whether or not the release was run.

Looking at this in a bit more detail I see that these don't impact whether or not the build is triggered and therefore I'm not really sure it really makes any difference. These patterns can only be matched if the build is triggered and the build is triggered by `'\s*:shipit:( skip validation| re release)*\s*'` which only allows whitespace.

The use of the find operator with matches appears equivalent to the use of the match operator and so I'm not sure it really makes any difference but the use of the operator by itself looks nicer.

Overall I don't know what this is solving so I'll dismiss my review for now.

hughsaunders:
@git-harry 	The regex being changed is not the trigger regex, that is https://github.com/rcbops/rpc-gating/blob/master/rpc_jobs/standard_job_release.yml#L10:
```
          trigger-phrase: '\s*:shipit:( skip validation| re release)*\s*'
```

The regex being changed, checks if the `skip validation` or `re release` modifiers are included in the comment, but that check will only be reached if the trigger regex matches.

PR #1305, opened 2019-02-04
RE-2354 Install groovy in Jenkins Lint jobs

hughsaunders:
This template is used to created lint jobs for team repos that contain
jjb definitions. Currently it doesn't ensure groovy is installed
so groovy checks are omitted. This commit fixes that by installing
groovy so that groovy scripts can be syntax checked.

Issue: [RE-2354](https://rpc-openstack.atlassian.net/browse/RE-2354)


hughsaunders:
recheck_unit_tests

PR #1307, opened 2019-02-04
Drop AIO jobs in favor of MNAIO snaps

antonym:
Drops a number of AIO/MNAIO jobs to start switching to
using MNAIO snaps and adds some incremental
jobs.

### DIFF
--- a/rpc_jobs/rpc_upgrades.yml
+++ b/rpc_jobs/rpc_upgrades.yml
@@ -135,140 +135,6 @@
     CRON: "{CRON_WEEKLY}"
 
 - project:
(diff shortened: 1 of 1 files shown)


antonym:
recheck_xenial_lint_test

antonym:
It runs some basic validation release jobs but they are more testing to validate things are working than spitting out artifacts.

antonym:
Compressed it a bit further and removed release from the name which isn't the release.

odyssey4me:
FYI the term `snapshot` refers to AIO snapshots, whereas the MNAIO tooling refers to `images`. It may be good to use the appropriate terminology for the job names and scripts to prevent confusion.

PR #1312, opened 2019-02-07
RE-404 Fix periodic cleanup and other Docker related changes

git-harry:
See individual commits for details.

Issue: [RE-404](https://rpc-openstack.atlassian.net/browse/RE-404)


rpc_jobs/periodic_cleanup.yml
```diff
@@ -82,10 +82,28 @@
                   sh "sudo restart vpnc-kronos"
                 }
                 common.docker_cache_workaround()
-                String sha = sh(script: "git rev-parse --verify HEAD", returnStdout: true).trim()
                 // Docker version on master doesn't like '/' in the name, using '-' as the separator
-                name = docker_utils.toInternalRegistryName("periodic_cleanup", sha, "re", env.RE_JOB_REPO_NAME, "-")
-                container = docker_utils.pullOrBuild(name, null)
+                name = docker_utils.toInternalRegistryName("periodic-cleanup", "latest", "re", env.RE_JOB_REPO_NAME, "-")
+                // jenkins docker plugin stuff does not work
```
hughsaunders: `docker.image` and `container.inside` are used below, presumably they are included in "docker plugin stuff"?

git-harry: It turns out this job has never been using a container, it always fails with only a log entry to indicate there was an issue and then runs the code on the enclosing host. I copied that section from the build_summary job and didn't spot the inconsistency with you comment referencing stuff. I'm going to push a patch to update this to remove the use of the Docker plugin in the job.

git-harry: https://rpc.jenkins.cit.rackspace.net/job/Periodic-Cleanup/1638 shows a run using the code from this pull request.

Dockerfile
```diff
@@ -1,9 +1,27 @@
...
+      python-pip \
+      build-essential \
+      python-dev \
+      libssl-dev \
+      curl \
+      libffi-dev \
+      sudo \
+      git-core \
+      libxml2-utils \
+ && rm -rf /var/lib/apt/lists/*
```
odyssey4me: Perhaps using ``apt-get clean`` would also be useful here if we're intending to do some cleanup to maintain a clean image?

scripts/build_summary/Dockerfile
```diff
@@ -1,9 +1,28 @@
...
+      python3-dev \
+      libssl-dev \
+      curl \
+      libffi-dev \
+      sudo \
+      git-core \
+      python3-lxml \
+      python-dev \
+      python-pip \
+ && rm -rf /var/lib/apt/lists/*
```
odyssey4me: Perhaps using ``apt-get clean`` would also be useful here if we're intending to do some cleanup to maintain a clean image?

Dockerfile
git-harry: According to https://docs.docker.com/develop/develop-images/dockerfile_best-practices/ this should already be done.

scripts/build_summary/Dockerfile
git-harry: According to https://docs.docker.com/develop/develop-images/dockerfile_best-practices/ this should already be done.

PR #1319, opened 2019-02-14
RE-2361 Repo Verify Job

hughsaunders:
This job recalculates the checksum for every package in rpc-repo
and finds any Packages(index) file SHAs that don't match the packages on
disk.

Issue: [RE-2361](https://rpc-openstack.atlassian.net/browse/RE-2361)

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -37,6 +37,7 @@ jira==1.0.10
 jmespath==0.9.3
 jsonpatch==1.23
 jsonpointer==2.0
+junit-xml==1.8
 keystoneauth1==3.11.2
 linecache2==1.0.0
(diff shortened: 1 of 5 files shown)


hughsaunders:
Example job, showing packages with inconsistent SHAs as test results: https://rpc.jenkins.cit.rackspace.net/job/rpc-repo-verify/2/

@odyssey4me I Think I might need to rsync from the repo before running the verify, but I could't determine a simple way of doing that by looking at the build job and playbooks, do you have any suggestions?

hughsaunders:
Decided to put off implementation of remediation into a new card [RE-2375](https://rpc-openstack.atlassian.net/browse/RE-2375).

PR #1323, opened 2019-02-15
Allow rpc-gating branch overrides for all jobs

odyssey4me:
Some jobs cannot currently have the rpc-gating branch changed.
This fixes that and ensures that they are all consistently
implemented.

Issue: [RE-2278](https://rpc-openstack.atlassian.net/browse/RE-2278)

### DIFF
--- a/rpc_jobs/bump_snapshots.yml
+++ b/rpc_jobs/bump_snapshots.yml
@@ -17,7 +17,12 @@
         }
       }
 
-      library "rpc-gating@${RPC_GATING_BRANCH}"
+      if (env.RPC_GATING_BRANCH != "master") {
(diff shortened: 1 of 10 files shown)


rpc_jobs/bump_snapshots.yml
```diff
@@ -17,7 +17,12 @@
         }
       }
 
-      library "rpc-gating@${RPC_GATING_BRANCH}"
+      if (env.RPC_GATING_BRANCH != "master") {{
```
git-harry: This is not a template

odyssey4me: Good catch! Fixed.

PR #1326, opened 2019-02-19
ASC-1655 Add Ceph Scenario to Staging Job

ryan-rs:
Add the "ceph" scenario to the system staging job so that we can start
creating Molecule tests for Ceph backends. Also, I dropped a redundant
image specification.

Issue: [ASC-1655](https://rpc-openstack.atlassian.net/browse/ASC-1655)

### DIFF
--- a/rpc_jobs/rpc_openstack.yml
+++ b/rpc_jobs/rpc_openstack.yml
@@ -608,6 +608,7 @@
           SLAVE_TYPE: "nodepool-ubuntu-bionic-om-io2"
     scenario:
       - swift
+      - ceph
     action:
       - system_staging
       - system_maas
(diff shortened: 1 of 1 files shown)


rpc_jobs/rpc_openstack.yml
```diff
@@ -604,10 +604,9 @@
     image:
       - bionic_mnaio_no_artifacts:
           SLAVE_TYPE: "nodepool-ubuntu-bionic-om-io2"
-      - xenial_mnaio_no_artifacts:
```
johnduarte: These two changes are completely unrelated. Can you break them into separate commits?

ryan-rs: @johnduarte Updated.

PR #1329, opened 2019-02-21
RE-2338 Add slack support for notifying of build failures

git-harry:
This change reworks `build_failure_issue` to be used more generally so
that it can also send messages on Slack if required. It is renamed to
`build_failure_notify` to account for this adjustment.

The hardening jobs have been updated to make use of Slack notifications.

Unit test failure expected due to change in function name requiring the job to be rebuilt. Tested manually with: https://rpc.jenkins.cit.rackspace.net/job/RE-unit-test-jira/2030/replay/diff

JIRA: RE-2338

Issue: [RE-2338](https://rpc-openstack.atlassian.net/browse/RE-2338)

### DIFF
--- a/rpc_jobs/standard_job_gate.yml
+++ b/rpc_jobs/standard_job_gate.yml
@@ -102,4 +102,4 @@
       }} else {{
         library "rpc-gating-master"
       }}
-      common.stdJob("gate", "{credentials}", "", "{wrappers}")
+      common.stdJob("gate", "{credentials}", "", "{wrappers}", "", "")
--- a/rpc_jobs/standard_job_premerge.yml
+++ b/rpc_jobs/standard_job_premerge.yml
@@ -93,5 +93,5 @@
         library "rpc-gating-master"
       }}
       if (! common.isSkippable(skip_pattern, "{credentials}")) {{
-        common.stdJob("check", "{credentials}", "", "{wrappers}")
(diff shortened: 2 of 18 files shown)


rpc_jobs/standard_job_premerge.yml
```diff
@@ -93,5 +93,5 @@
         library "rpc-gating-master"
       }}
       if (! common.isSkippable(skip_pattern, "{credentials}")) {{
-        common.stdJob("check", "{credentials}", "", "{wrappers}")
+        common.stdJob("check", "{credentials}", "", "{wrappers}", "")
```
hughsaunders: Should there be two extra parameters here instead of one?

rpc_jobs/standard_job_gate.yml
```diff
@@ -102,4 +102,4 @@
       }} else {{
         library "rpc-gating-master"
       }}
-      common.stdJob("gate", "{credentials}", "", "{wrappers}")
+      common.stdJob("gate", "{credentials}", "", "{wrappers}", "")
```
hughsaunders: Should there be two extra parameters here instead of one?

rpc_jobs/standard_job_premerge.yml
git-harry: Does it matter if it doesn't work?

PR #1330, opened 2019-02-21
Add rpc-ceph mimic and bionic gates

mvollman:
Add mimic gates for xenial and bionic

### DIFF
--- a/rpc_jobs/rpc_ceph.yml
+++ b/rpc_jobs/rpc_ceph.yml
@@ -4,7 +4,7 @@
     repo_name: "rpc-ceph"
     repo_url: "https://github.com/rcbops/rpc-ceph"
 
-    hold_on_error: "12h"
+    hold_on_error: "4h"
 
     branches:
       - "master"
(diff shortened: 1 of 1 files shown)


rpc_jobs/rpc_ceph.yml
```diff
@@ -57,6 +100,7 @@
 
     scenario:
       - "functional"
+      - "functional_mimic"
```
hughsaunders: Any reason bluestore_mimic isn't being added here?

mvollman: Having bluestore_mimic here would provide value.  I left it off to reduce the number of gates we are running to conserve testing resources.  I was worried that we may have too many gates and was being conservative for now.  But I am happy to add it in and we can always come back and modify it later if it becomes a problem.

mvollman:
I can add all the gates to the release stage.  I was being conservative by not adding all the gates at once, but I am happy to add them now and we can pull them back later if there are too many.

mvollman:
recheck_CIT_unit_tests

mvollman:
recheck_unit_tests

mvollman:
recheck_unit_tests

PR #1331, opened 2019-02-21
Checkmarx Scan Failure Should Create a JIRA Ticket

dealako:
* Added logic to check the build result and if unsuccessful throws a
REException which is caught in the calling function. From there the
common code will create a JIRA ticket if the JIRA_PROJECT_KEY is set.

JIRA: RE-2388

Issue: [RE-2388](https://rpc-openstack.atlassian.net/browse/RE-2388)

### DIFF
--- a/pipeline_steps/checkmarx.groovy
+++ b/pipeline_steps/checkmarx.groovy
@@ -1,4 +1,5 @@
 import java.security.SecureRandom
+import com.rackspace.exceptions.REException
 
 // Upload all folders in the current working directory to checkmarx and scan for vulnerabilities.
 // If you wish to scan a subdir of the working dir, call this function within dir("subdir"){}
@@ -81,6 +82,10 @@ def scan(String scan_type, String repo_name, String exclude_folders){
                 throw e
             } //try
         } // retry
+
+        if (currentBuild.getRawBuild().getResult().isWorseThan(Result.fromString("SUCCESS"))) {
+            throw new REException("Checkmarx Scan Threshold Exceeded for repo: " + repo_name)
+        }
     } // withCredentials
 }
 
--- a/pipeline_steps/checkmarx.groovy
+++ b/pipeline_steps/checkmarx.groovy
@@ -83,7 +83,7 @@ def scan(String scan_type, String repo_name, String exclude_folders){
             } //try
         } // retry
 
-        if (currentBuild.getRawBuild().getResult().isWorseThan(Result.fromString("SUCCESS"))) {
+        if (Result.fromString(currentBuild.currentResult).isWorseThan(Result.fromString("SUCCESS"))) {
             throw new REException("Checkmarx Scan Threshold Exceeded for repo: " + repo_name)
         }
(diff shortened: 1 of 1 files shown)


pipeline_steps/checkmarx.groovy
```diff
@@ -80,6 +80,10 @@ def scan(String scan_type, String repo_name, String exclude_folders){
                 // exception must propagate back to the retry call
                 throw e
             } //try
+
+            if (currentBuild.getResult() && currentBuild.getResult().isWorseThan(SUCCESS)) {
```
git-harry: The [documentation](https://rpc.jenkins.cit.rackspace.net/job/PM-Checkmarx_all-rpc-differ/pipeline-syntax/globals#currentBuild) doesn't list `currentBuild` as supporting this method but has others that are similar, has this been tested? This looks like something probably found in `currentBuild.rawBuild`.

pipeline_steps/checkmarx.groovy
```diff
@@ -80,6 +80,10 @@ def scan(String scan_type, String repo_name, String exclude_folders){
                 // exception must propagate back to the retry call
                 throw e
             } //try
+
```
git-harry: This following block is within the retry block so it's going to cause unintentional retries when there is a genuine failure.

pipeline_steps/checkmarx.groovy
```diff
@@ -80,6 +80,10 @@ def scan(String scan_type, String repo_name, String exclude_folders){
...
+            if (currentBuild.getResult() && currentBuild.getResult().isWorseThan(SUCCESS)) {
+                throw new REException("Checkmarx Scan Threshold Exceeded for repo: " + repo_name)
```
git-harry: Does REException need importing?

pipeline_steps/checkmarx.groovy
dealako: Yes, will move below the retry logic ✅

pipeline_steps/checkmarx.groovy
dealako: Yes

pipeline_steps/checkmarx.groovy
dealako: The documentation link is helpful. Pushed update to hopefully address the usage of this object.

pipeline_steps/checkmarx.groovy
```diff
@@ -81,6 +82,10 @@ def scan(String scan_type, String repo_name, String exclude_folders){
                 throw e
             } //try
         } // retry
+
+        if (currentBuild.getRawBuild().getResult().isWorseThan(Result.fromString("SUCCESS"))) {
```
git-harry: If the scan is successful `currentBuild.getRawBuild().getResult() == null` which causes this line to raise an exception. I would suggest:
```suggestion
        if (Result.fromString(currentBuild.currentResult).isWorseThan(Result.fromString("SUCCESS"))) {
```

